PR #14, opened 2026-07-21
fix(audio): recover buffered lead after frontend stalls

### LINKED ISSUE #13: Dropped GUI frames permanently drain the audio safety buffer
### Version

Systemless 0.1.132 (`02f3b75`), desktop GUI.

### Problem

The GUI currently queues one fixed frame of audio on every frontend update.
When a video deadline is missed and that update is dropped, the corresponding
audio duration is also lost. A loading or rendering stall can therefore consume
the audio backend's initial safety lead even when average emulation speed is
adequate.

After the first underrun, the backend resumes as soon as any samples arrive.
Because production and consumption then proceed at approximately the same
average rate, the buffer remains close to empty and ordinary timer jitter
causes a train of short audible gaps.

### Expected behavior

Audio production should follow elapsed wall time independently of whether a
video frame was late or dropped. After an underrun, playback should rebuild its
configured safety lead before resuming, rather than repeatedly balancing on an
empty queue.

### Proposed fix

- Derive each GUI audio budget from elapsed wall time instead of a fixed video
  frame duration.
- Cap one recovery interval at 250 ms so a long host suspension cannot create
  an unbounded callback burst.
- After an underrun, emit silence until the existing 90 ms prefill target has
  been restored.
- Preserve queued samples during rebuffering and resume from the first queued
  frame so short transients are not discarded.
- Add regression tests for a 100 ms frontend stall and post-underrun recovery.

This changes audio recovery only; it does not alter guest CPU or VBL pacing.

rlanday:
Closes #13

### Summary

- derive GUI audio production from elapsed wall time independently of video frame scheduling
- cap a single recovery interval at 250 ms
- after an underrun, hold silence until the existing 90 ms safety lead is rebuilt
- preserve queued audio and resume from its first frame
- add regression coverage for a 100 ms frontend stall and post-underrun recovery

The GUI previously produced exactly one frame's worth of audio per frontend update. When the frame scheduler intentionally dropped a late video update, it also dropped that duration of audio, permanently draining the device queue's safety lead. If the queue then underruns, immediately consuming the first small refill keeps playback balanced at the empty edge and turns ordinary scheduling jitter into repeated audible gaps.

This change keeps video frame dropping intact while accounting for elapsed audio time separately. It does not change CPU or VBL pacing.

### Verification

- `cargo test --lib host_audio_` — 11 passed
- `cargo test --bin systemless step_frame_recovers_audio_elapsed_during_a_dropped_video_frame` — 1 passed
- `cargo test --lib` — 2,633 passed, 3 ignored
- `cargo test --bin systemless` — 26 passed

### DIFF
--- a/src/audio.rs
+++ b/src/audio.rs
@@ -265,11 +265,25 @@ impl SharedAudioState {
             return [0.0, 0.0];
         }
 
-        if self.underrun_samples > 0 && trace_audio_enabled() {
-            eprintln!(
-                "[AUDIO] underrun ended after {} samples",
-                self.underrun_samples
-            );
+        // Once the stream underruns, consuming the first small refill
+        // immediately leaves it permanently balanced on empty: the GUI and
+        // device produce/consume at the same average rate, so ordinary timer
+        // jitter becomes a train of audible gaps. Hold silence until the
+        // original safety lead has been rebuilt, then resume continuously.
+        if self.underrun_samples > 0 && self.buffer.len() < host_audio_prefill_samples() {
+            self.underrun_samples = self.underrun_samples.saturating_add(1);
+            self.source_phase = 0.0;
+            self.last_frame = [0.0, 0.0];
+            return [0.0, 0.0];
+        }
+
+        if self.underrun_samples > 0 {
+            if trace_audio_enabled() {
+                eprintln!(
+                    "[AUDIO] underrun ended after {} samples",
+                    self.underrun_samples
+                );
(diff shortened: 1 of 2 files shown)


benletchford:
Integrated as `3cf0a5f` after resolving the overlap with the merged Metal stack. The audio underrun/recovery tests and the complete combined release gate pass. Closing this PR as superseded by the integrated commit.

PR #18, opened 2026-07-21
fix(gui): detect and cache centered game viewports

### LINKED ISSUE #17: Detect and present a fullscreen game's actual centered viewport
### Version

Systemless 0.1.132 with the Metal presenter work from #10 and #16.

### Problem

Some fullscreen games render a smaller centered playfield inside the configured 800x600 display. Systemless presents the entire display, leaving a permanent border around the game. Window sizing and host-to-guest mouse conversion also continue to use 800x600 rather than the displayed playfield.

Live resizing has two additional problems on macOS: the Metal layer does not track its parent layer immediately, and the newest guest image is not presented until a later emulation frame. This produces stale scaling and transient white borders.

### Expected behavior

- infer the actual centered presentation rectangle from guest behavior, without game-name checks
- size the window to an integer multiple of that rectangle as soon as it is known
- continuously aspect-fit and center content during arbitrary resizing
- map mouse coordinates through the same cropped, letterboxed viewport
- update the drawable immediately during live resize
- avoid cropping ordinary windows, full-screen blits, or temporary offscreen buffers

### Proposed fix

Prefer geometry from the app-managed `CGrafPort` that Systemless has actually selected and centered for presentation. Otherwise require repeated agreement from centered screen `CopyBits` operations or distinct 8-bit framebuffer updates before accepting a crop. Reset detection when the guest screen mode changes.

Pass the accepted rectangle to the Metal shader, use one shared aspect-fit calculation for presentation and input, make the Metal layer follow its parent bounds, and present the latest guest image directly from resize events.

This should remain independent of game identity and retain the full emulated screen when no safe centered rectangle can be established.

rlanday:
Closes #17.

Depends on #16.

## Summary

- Detect centered game content from QuickDraw port geometry, centered CopyBits operations, or stable 8-bit framebuffer borders.
- Crop Metal presentation to the detected content while preserving correct mouse-coordinate mapping.
- Keep live window resizing centered and aspect-correct.
- Upload the complete guest framebuffer even when presenting a crop, preventing the white rectangle caused by missing source rows.
- Cache validated viewport geometry per game so subsequent launches open at the final integer-sized window dimensions without resizing.
- Keep the Metal layer synchronized with native window resizing.
- Temporarily extend the presented guest rectangle when a visible Classic dialog falls outside the cached gameplay crop.
- Expand the native window at the existing guest-pixel scale instead of shrinking the game, preserving the gameplay content's desktop position.
- Restore the original window geometry when the dialog closes without a one-frame resize bounce.
- Preserve a cached crop while startup detection reconfirms the guest screen mode.

## Menu-bar follow-up

The current frontend normally hides the Classic menu bar, and the viewport heuristics therefore treat the gameplay rectangle and transient dialogs as the relevant visible content. When menu-bar support is added, this logic may need revision: the menu bar occupies the guest screen's top edge, affects the coordinate space available to windows and dialogs, and may need to participate in both crop detection and native-window expansion. This PR keeps that policy isolated in the presentation-rectangle calculation so it can be revisited without changing the Metal decoder.

## Validation

Tested on an Intel Mac with Lemmings 1.5.2:

- Detected 640x392 content at `(80,104)` inside the 800x600 guest display.
- No white rectangle or surrounding guest-screen border.
- Window resizing and input mapping work correctly.
- After the first detection, subsequent launches open directly at 640x392 without a resize animation.
- The access-code dialog expands the native window without scaling or moving the gameplay area, then restores the gameplay geometry when closed.
- Relaunch with a cached viewport no longer enters a provisional full-screen transition.
- Sequential gameplay through Fun 4 retains the entrance, exit, and animated water. (The underlying Memory Manager fix is intentionally submitted separately.)
- `cargo test --bin systemless` — 35 passed.
- Focused visible-dialog library tests — 8 passed.
- Focused `RecoverHandle` tests on the combined validation stack — 2 passed.

`cargo fmt --check` remains blocked by pre-existing formatting differences in unrelated files. `cargo clippy -- -D warnings` remains blocked by existing repository-wide warnings unrelated to this change.


rlanday:
I think this cropping is too aggressive because when I click “New Level” in lemmings, the dialog box is cut off on the top. Will revise
Screenshot 2026-07-21 at 5 01 43 PM

benletchford:
Integrated. The viewport commit `74a3391` reached master through #20/#24; the later dialog fixes landed as `b3d4164` and `3ab59ab`, with async/transactional Metal integration in `891562c`. The combined headless build, macOS all-target build, 2,646 library tests, 39 GUI tests, and package verification pass. Closing this PR as superseded by the integrated history.

PR #24, opened 2026-07-22
perf(gui): decouple Metal presentation from emulation

### LINKED ISSUE #23: Metal drawable backpressure stalls emulation and input
## Problem

The macOS Metal guest-frame path calls `CAMetalLayer.nextDrawable()` synchronously from `App::render_frame()`. That function runs on the same winit/AppKit thread that advances the 68k guest and handles mouse and keyboard events.

When every guest frame is submitted, the two-drawable layer fills and `nextDrawable()` applies display backpressure directly to that thread. This is especially visible when the guest cadence (60.15 Hz) is slightly faster than the active display (59.9404 Hz), but it also occurs when Core Animation delays an occluded or background window.

An unchanged-frame optimization reduces how often this happens for mostly static games, but a game that changes every frame still exercises the blocking path.

## Profiling evidence

The workload was run with `SYSTEMLESS_DISABLE_UNCHANGED_FRAME_SKIP=1` so every 60.15 Hz guest frame attempted a Metal submission.

Before decoupling presentation:

- presentation averaged approximately 21.6–21.8 ms per attempted frame;
- 93% of sampled main-thread stacks were blocked in `CAMetalLayer::nextDrawable` or its semaphore wait;
- because guest execution, input, rendering, and audio scheduling share this thread, drawable backpressure delayed all four.

With drawable acquisition moved to a dedicated presenter thread:

- the producer-side framebuffer snapshot and mailbox enqueue averaged 69–70 microseconds per attempted frame;
- `nextDrawable()` wait time, approximately 10.6–12.1 ms per submitted frame in the measured run, occurred only on the presenter thread;
- Metal encoding itself averaged approximately 140–144 microseconds per submitted frame;
- the guest continued to execute at its own cadence while a slower or occluded compositor coalesced pending presentation frames.

The main-thread presentation cost therefore fell by roughly 300x in the adversarial all-submissions workload.

## Proposed behavior

- Snapshot each complete guest framebuffer without waiting for a drawable.
- Store at most one pending frame in a latest-frame mailbox.
- Let a dedicated Metal thread acquire drawables, upload, encode, and submit.
- If the compositor is behind, replace the pending snapshot instead of building an unbounded queue.
- Recycle snapshot storage to avoid allocating a new framebuffer vector every frame.
- Preserve synchronous presentation only for operations that explicitly require a current Core Animation transaction, such as atomic window-geometry changes.
- Report submitted and coalesced counts plus producer, drawable-wait, and render timing in the existing Metal profiler.

This does **not** skip guest execution. A 60.15 Hz game continues to execute every guest frame. Presentation may coalesce the latest completed image when the physical display or compositor cannot consume all unique images; this bounds latency rather than allowing an ever-growing queue.

## Follow-up profiling

After this change lands, profile the remaining CPU and memory costs separately:

- 68k interpreter and memory-resolution hot paths;
- framebuffer snapshot/copy bandwidth and allocation churn;
- idle Toolbox/event behavior;
- audio scheduling and underrun recovery;
- input-to-photon latency;
- a future `CAMetalDisplayLink` path on macOS 14+ for compositor-directed pacing.

rlanday:
## Summary

Move normal native-guest Metal presentation off the AppKit/emulation thread and onto a dedicated presenter thread. The producer now copies each complete guest framebuffer into a bounded, one-slot latest-frame mailbox; the presenter blocks on `CAMetalLayer::nextDrawable`, uploads the newest complete frame, and submits it to Metal.

This prevents display backpressure from stalling 68k execution, event processing, and audio scheduling.

Fixes #23.

## Dependency

This is stacked on #20 because it reuses that PR's native guest-frame path, visible-region metadata, and frame profiler. Until #20 merges, GitHub's aggregate diff against `master` includes its prerequisite commits; the new work in this PR is the top commit, `perf(gui): decouple Metal presentation from emulation`.

## Problem

The prior path called `CAMetalLayer::nextDrawable()` synchronously from `present_guest_frame`, which runs on the same thread as AppKit event handling and guest execution. When both layer drawables were in flight, Core Animation blocked that thread until a drawable became available.

That wait dominated the frame path despite the actual Metal work being very small. In Lemmings it showed up as delayed or inconsistent clicks, sluggish cursor response, and less time available for guest execution and sound work.

This is display pacing, not expensive 68k emulation or expensive GPU rendering. Moving only shader work to another queue would not help because drawable acquisition was the blocking operation.

## Design

The normal guest-frame flow is now:

1. The AppKit/emulation thread builds the frame metadata and performs #20's optional unchanged-frame check.
2. It copies the packed guest framebuffer into a one-slot mailbox and returns immediately.
3. A dedicated `systemless-metal-present` thread waits for a drawable.
4. Once a drawable is available, the worker takes the newest complete mailbox entry, copies it to one of two shared Metal buffers, encodes the existing indexed/monochrome shader pass, and submits it.
5. Mailbox and framebuffer storage are recycled to avoid allocating a new `Vec` every frame.

If another guest frame arrives while one is pending, it atomically replaces the pending frame. An in-progress submission is never mutated. This bounds latency and memory instead of building an ever-growing presentation queue.

The guest still executes every emulated frame. Coalescing affects only physical presentation when the guest producer is ahead of drawable availability—for example, a nominal 60 Hz guest on a measured 59.9404 Hz display. It is impossible to physically show every 60 Hz guest frame on a slower display without either backpressuring emulation or accumulating latency. Keeping the newest complete frame gives input the lowest practical display latency.

The existing generic ARGB/debug presentation path remains synchronous. Before it presents, it pauses and drains the asynchronous worker so an older native guest frame cannot be submitted over the fallback image. The worker sleeps on a condition variable while paused and resumes when the native path next enqueues a frame.

Layer geometry is still mutated only on AppKit's main thread. The worker owns its command queue and Metal staging buffers and only acquires drawables and encodes/submits commands.

## Profiling

Measurements were collected on the Intel Mac used to reproduce the Lemmings latency, with `SYSTEMLESS_PROFILE_METAL_FRAMES=1` and unchanged-frame skipping disabled to stress a continuously changing 60 Hz game.

Before this change, the synchronous presentation path took approximately **21.6–21.8 ms per attempt**. A sampled run attributed about **93% of main-thread stacks** to waiting in drawable/semaphore acquisition.

After this change:

- Producer-side snapshot and enqueue: **~69–70 µs per attempt**
- Presenter-thread `nextDrawable` wait: **~10.6–12.1 ms per submitted frame**
- Presenter-thread upload, encode, and submit: **~140–144 µs per submitted frame**
- Active display refresh measured by macOS: **59.9404 Hz**
- Guest production cadence observed during the run: **~60.15 Hz**

The key result is that the multi-millisecond drawable wait is no longer paid by the thread running input and the guest. The GPU work itself remains a small fraction of a millisecond.

Some longer profiling runs included window occlusion/background intervals and therefore reported substantial mailbox coalescing; those counts should not be interpreted as the normal visible gameplay rate. The forced-changing-frame measurements above establish the important worst case: even when every guest frame changes, producer-side work remains about 70 µs rather than blocking for a display interval.

## Instrumentation

The Metal profiler now reports:

- guest presentation attempts
- submitted frames
- mailbox-coalesced frames
- unchanged frames skipped by #20
- forced presentations
- comparison time per attempt
- producer snapshot/enqueue time per attempt
- drawable wait time per submission
- worker render time per submission

This makes producer cost, display backpressure, actual render cost, and the unchanged-frame optimization independently visible.

## Validation

- `cargo test --bin systemless` — 35 passed
- Lemmings tested interactively on Intel macOS with unchanged-frame skipping both enabled and disabled
- Forced-changing-frame profiling verified that `nextDrawable` no longer blocks guest execution

## Interaction with the viewport/resize PR

#18 has a special Core Animation transaction path for atomic window/content resizing. When these branches are combined, that transaction-only synchronous presentation must first pause and drain this worker, then resume asynchronous presentation afterward. This PR intentionally does not duplicate #18's later resize code; keeping that integration explicit avoids coupling this performance fix to the cropping changes.

## Follow-up work

Once this is reviewed, further profiling can focus on the remaining CPU and memory costs instead of conflating them with display blocking. Likely targets include reducing the producer snapshot copy, measuring per-trap/68k dispatch cost, and checking whether framebuffer dirtiness can safely avoid work without making correctness depend on every drawing path being instrumented.


benletchford:
I reviewed the full GUI stack and the independent fixes together. The combined candidate passes the headless build, a macOS all-target/all-feature build, 2,646 library tests, 39 GUI tests, and package verification. I found two integration issues outside this PR’s standalone diff: `src/bin/metal_present.rs` is auto-discovered by Cargo as an unintended binary, and #18’s transactional resize path needs to pause/drain the async presenter and use a synchronous transactional submission. I have a tested maintainer follow-up for both. I’m merging the compatible stack bottom-up, then landing the #18 follow-ups and integration fixes on master.

PR #27, opened 2026-07-23
perf(runner): fast-forward TickCount BEQ waits

rlanday:
## Summary

- recognize the `BEQ.S` form of the stack-result `TickCount` spin loop
- accelerate it only when the captured tick still equals the comparison register
- keep the final `CMP.L` and branch in the 68k core so CCR flags, stack postincrement, PC, and instruction accounting remain architectural
- retain support and regression coverage for the existing `BCC.S` form

## Why this matters

Classic Mac OS exposes a roughly 60 Hz clock through the `_TickCount` Toolbox trap and the low-memory `Ticks` global. Because there is no mandatory blocking API for waiting until the next tick, applications commonly poll `TickCount` in a short 68k loop.

Lemmings 1.5.2 uses this exact sequence at guest address `$00023086`:

```asm
$23086  CLR.L  -(A7)       ; allocate the Pascal LONGINT result slot
$23088  _TickCount         ; write the current tick to (A7)
$2308A  CMP.L  (A7)+,D7    ; compare and reclaim the result slot
$2308C  BEQ.S  $23086      ; wait while the tick is unchanged
```

Systemless already has a guarded fast-forward path for this stack-result template, but it accepted only condition code `$64` (`BCC`). Its regression test also described that encoding as the exact Lemmings loop. The shipped game actually uses `$67F8` (`BEQ.S *-8`), so the matcher never fired and the emulator repeatedly executed the polling loop until the next host-paced guest tick.

For `BEQ`, this change first verifies that the just-captured stack result equals `Dn`. If they differ, the branch would already fall through and the runner does nothing. If they are equal, the runner advances exactly to `Dn + 1`, subject to the existing GUI tick cap and interrupt handling. It refreshes the stack result after synthetic VBL advancement, then leaves `CMP.L (A7)+,Dn` and `BEQ.S` for the CPU to execute normally.

### Controlled whole-process host sample

Both samples used the same combined Systemless/m68k build, the same Lemmings 1.5.2 disk, level 1, and a 30-second macOS `sample` capture. The only behavioral difference was this patch.

| Top-of-stack symbol | Before | After | Change |
|---|---:|---:|---:|
| `FixtureRunner::run_steps_internal` | 266 | 227 | -14.7% |
| `m68k::execute_mem_op` | 121 | 106 | -12.4% |

The absolute CPU cost was already modest after the preceding JIT work, so this is best read as removal of avoidable idle work rather than a frame-rate claim.

### Guest-site counters

In the retained counter-enabled reference profile, each interpreted wait-loop instruction was the hottest decoded memory site:

| Guest site | Opcode | Before executions | Patched profile |
|---|---:|---:|---:|
| `$00023086` | `$42A7` (`CLR.L -(A7)`) | 6,988,354 | below 36,721-entry top-60 cutoff |
| `$0002308A` | `$BE9F` (`CMP.L (A7)+,D7`) | 6,988,354 | below 36,721-entry top-60 cutoff |

That establishes at least a 99.47% reduction at both polling sites. This counter comparison is supporting mechanism evidence rather than the controlled performance result: the retained baseline predates the latest JIT stack, whereas the host samples above are an exact before/after pair.

## Validation

- `cargo test --lib`: 2,647 passed, 3 ignored
- exact `$42A7/$A975/$BE9F/$67F8` Lemmings loop test
- already-expired `BEQ` test verifies no tick advancement
- existing `BCC` behavior and GUI tick-cap tests remain covered

### DIFF
--- a/src/runner.rs
+++ b/src/runner.rs
@@ -2325,8 +2325,8 @@ impl FixtureRunner {
         let w0 = self.bus.read_word(pc_after_trap);
 
         // Template D consumes TickCount directly from its stack result slot:
-        //   CLR.L -(A7); _TickCount; CMP.L (A7)+,Dn; BCC.S <back-to-CLR>
-        // Lemmings uses this form for its frame delay loop.
+        //   CLR.L -(A7); _TickCount; CMP.L (A7)+,Dn; Bcc.S <back-to-CLR>
+        // Lemmings uses the BEQ form for its frame delay loop.
         if (w0 & 0xF1FF) == 0xB09F {
             let dn = ((w0 >> 9) & 7) as usize;
             return self.try_spin_template_d(pc_after_trap, dn, tick_cap);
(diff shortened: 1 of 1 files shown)


benletchford (approved):
Reviewed the matcher and both branch semantics. BEQ acceleration is gated on the captured stack result still equalling Dn, advances only to Dn+1, respects the GUI tick cap, refreshes the captured result, and leaves CMP/branch execution to the 68k core so flags, A7, PC, and accounting remain architectural. The combined Systemless result passes 2,653 library tests, 39 desktop tests, headless/full-feature builds, package dry-run, and clippy. Worth merging.

PR #36, opened 2026-07-26
feat(macos): mirror guest menus in the menu bar

### LINKED ISSUE #35: Mirror Classic Mac OS menus into the native macOS menu bar
Systemless currently renders or suppresses the guest menu bar inside the emulated framebuffer, but it does not expose a Classic Mac application's menus through the native macOS menu bar. Full-screen games can consequently lose their only route to commands such as sound toggles, level selection, and quitting.

The macOS frontend should mirror the active guest menu list into native menus while preserving Classic menu semantics:

- derive the application menu title from guest application metadata rather than the disk-image name;
- preserve item titles, separators, enabled state, checkmarks, and keyboard equivalents;
- avoid unrelated AppKit-provided Edit-menu entries;
- pause guest execution while native menu tracking is active, matching foreground Classic Mac menu behavior;
- deliver a native selection through the guest's normal `MenuSelect` path, including when the application is busy animating and does not immediately enter menu tracking; and
- keep the implementation isolated to the desktop frontend so other targets remain portable.

Native mirroring cannot necessarily reproduce every application-defined menu. Games that install custom MDEF or MBDF procedures, draw directly into menu structures, or depend on other nonstandard menu behavior may need a compatibility path that renders and tracks the original Classic menu bar inside the guest display instead.

The native window title is currently `Systemless - Macintosh Emulator`; including the active application title there can be handled separately.

rlanday:
## Summary

- mirror the guest Menu Manager model into the native macOS menu bar;
- derive the application menu title from the guest's About item, accepting `...`, the MacRoman ellipsis, and titles without an ellipsis;
- preserve enabled state, separators, checkmarks, and keyboard equivalents while removing unrelated AppKit-provided Edit items;
- latch native selections and redeliver at most once per guest tick until the application consumes them through `MenuSelect`.

The retry behavior matters for animation-heavy applications. A one-shot synthetic event could be discarded before the guest next called `MenuSelect`, making commands such as 3 in Three's **Sound** item appear intermittent. The latched event remains within the guest's normal menu-selection path instead of invoking a menu command out of band.

The AppKit target object is initialized lazily on the main thread, so constructing the emulator application model in Cargo worker-thread tests remains safe.

`docs/menu-bar-ux.md` records the UX, portability boundary, custom MDEF/MBDF limitations, and future work.

Native mirroring should remain the default, but it cannot necessarily reproduce application-defined menu rendering or tracking. Games that install custom MDEF/MBDF procedures or otherwise depend on nonstandard menu behavior may need a compatibility mode that renders and tracks the original Classic menu bar directly inside the guest display.

## Validation

- `cargo test --lib`: 2,698 passed, 3 ignored
- `cargo test --bin systemless`: 43 passed
- manually verified the native menu bar with Lemmings and 3 in Three;
- manually verified that 3 in Three's checked **Sound** item reliably toggles while an animation is running.

## Screenshot notes

The attached screenshot demonstrates the native menu bar and guest File menu. Graphics defects visible in that capture are addressed by separate QuickDraw changes.

The native window is still titled `Systemless - Macintosh Emulator`; a future change may include the active guest application title.

Native macOS menu bar mirroring 3 in Three's guest menus.
Screenshot 2026-07-25 at 8 32 32 PM

Closes #35


benletchford:
Wow, nice work!

PR #100, opened 2026-07-26
fix: initialize QuickDraw cursor vectors

### LINKED ISSUE #99: Dark Castle exits while entering playable gameplay
## Game

Dark Castle demo (68k), SHA-256 `ddbb2489d358ea5d9c89b3df44f913facea71b6f1bc3bfe9dc23ad125cb7dcc4`.

## Current behavior

The verified 68k game executable loads its data and renders the title and Scores of Merit screens on Systemless. Selecting the Play control begins the gameplay-loading path and reads the expected level data, but the application then returns through PC 0 after about 526 million instructions instead of remaining in a controllable game session.

At the equivalent scripted checkpoint in BasiliskII, the application remains alive in the Great Hall. Holding D produces Colin's right-facing movement response and gameplay continues.

Focused follow-up disproved the initial keyboard-bitmap diagnosis: the game does not read `KeyMapLM` or call `GetKeys` on this path, and Systemless delivers the injected keyDown and autoKey events. Changing the internal KeyMap bit order had no effect on the failure and is not part of this issue.

The bounded compatibility gap is therefore the generic application/runtime lifecycle used while entering playable gameplay: the Play path must not fall out through the process return chain while the same 68k executable remains active under the classic Mac oracle.

## Acceptance criteria

- Selecting Play reaches and remains in the Great Hall without an unimplemented trap, premature PC-0 return, halt, or crash.
- Holding D produces the same right-facing movement response as BasiliskII and changes gameplay state rather than only queuing an event.
- At least one additional keyboard action (for example A movement or Space jump) is demonstrated in the live room.
- Mouse aim/click still throws a rock.
- The fix is generic Systemless application/runtime behavior with focused regression coverage and no title-specific runtime special case.
- Existing application exit behavior remains covered: an intentional application quit must still terminate cleanly.

## Evidence boundary

Title artwork, the score screen, an attract-mode Great Hall screenshot, or successful event queuing does not satisfy this issue. Completion requires scripted keyboard-controlled gameplay on Systemless and comparison with the BasiliskII oracle.

benletchford:
Closes #99.

## What changed

- initialize the private QuickDraw `JHideCursor`, `JShowCursor`, and `JShieldCursor` low-memory bottleneck vectors with callable 68k trampolines
- initialize the private `$0574` QuickDraw compatibility hook to the existing no-op callback
- cover the vector opcodes, cursor behavior, return path, and Pascal stack cleanup with a focused runtime test

## Root cause

Dark Castle calls these QuickDraw low-memory vectors directly while entering gameplay. They were left as `NIL`, so the guest eventually jumped to PC `$00000000`. The `JShieldCursor` callback ABI was verified against MPW `Quickdraw.h`: four 16-bit Pascal arguments, requiring eight bytes of caller-frame cleanup. The implementation is generic QuickDraw HLE and contains no title-specific behavior.

## Gameplay validation

A scripted run now proceeds from the title and score screens into the Great Hall, remains running for more than 600 million guest instructions, and responds to left/right keyboard movement and mouse throwing input. A BasiliskII oracle run reaches the same gameplay state and action poses.

## Checks

- `cargo test --lib` — 2,720 passed, 3 ignored
- focused `JShieldCursor`, `HideCursor`, and `ShowCursor` ABI/behavior tests
- `git diff --check`

### DIFF
--- a/src/memory/globals.rs
+++ b/src/memory/globals.rs
@@ -110,6 +110,20 @@ pub mod addr {
     pub const M_TEMP: u32 = 0x0828; // Temporary mouse position (Point) - interrupt level
     pub const MOUSE_LOC: u32 = 0x082C; // Mouse location (Point) - "RawMouse"
     pub const MOUSE_LOC2: u32 = 0x0830; // Secondary mouse location (Point)
+    /// Private QuickDraw compatibility hook at `$0574`.
+    ///
+    /// The clean-room Executor low-memory table names this `JUnknown574` and
+    /// initializes it to a no-argument no-op callback.
+    pub const J_UNKNOWN_574: u32 = 0x0574;
+    /// QuickDraw cursor bottleneck vectors (ProcPtr).
+    pub const J_HIDE_CURSOR: u32 = 0x0800;
+    pub const J_SHOW_CURSOR: u32 = 0x0804;
+    /// JShieldCursor: address of QuickDraw's cursor-shielding bottleneck.
+    ///
+    /// The classic low-memory vector table places `JShieldCursor` at `$0808`.
+    /// MPW Quickdraw.h declares its Pascal callback ABI as four 16-bit
+    /// rectangle coordinates: left, top, right, bottom.
+    pub const J_SHIELD_CURSOR: u32 = 0x0808;
     /// JSwapFont: address of the Font Manager's FMSwapFont routine (ProcPtr).
     ///
     /// This private vector is called directly by QuickDraw text code. Executor's
--- a/src/runner.rs
+++ b/src/runner.rs
@@ -2219,6 +2219,54 @@ impl FixtureRunner {
             );
         }
 
+        // Private QuickDraw callback at $0574. Executor's clean-room table
+        // initializes this as a no-argument no-op; some QuickDraw clients call
+        // it immediately after JShieldCursor, so NIL is not a safe default.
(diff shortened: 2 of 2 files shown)


benletchford:
Merged `master` in to resolve the overlap with #82, #84, and #86, which landed the $0574, $0804, and $0808 vectors in the meantime.

Resolution:
- **$0574** — kept #84's `SwapMMUMode` trampoline rather than this branch's no-op. $0574 is the OS trap-table slot for `_SwapMMUMode` (`$0400 + $5D * 4`), and both seeds present the same zero-argument `RTS` contract to a direct `JSR`, so the Dark Castle call path is unchanged; `mmu_mode` has no behavioural effect beyond the `MMU32Bit` low-memory byte.
- **$0804 JShowCursor** — kept #86's `$A853; RTS` form; behaviourally identical to this branch's pop/JMP form.
- **$0808 JShieldCursor** — kept #82's trampoline, which dispatches through the real `$A855` HLE instead of discarding the arguments, so the vector honours the ShieldCursor intersection visibility added in #94.
- **$0800 JHideCursor** — taken from this branch; the one vector master had not already seeded. Its coverage is now a focused `init_app_seeds_callable_hide_cursor_low_memory_vector` test.

`cargo test --lib` on the merge result: 2,739 passed, 3 ignored.

PR #109, opened 2026-07-27
feat: implement the cooperative Thread Manager and a callable menu definition procedure

### LINKED ISSUE #108: Warcraft II: Tides of Darkness refuses to launch without the Thread Manager, then crashes through a NIL menuProc
## Observed behaviour

Warcraft II: Tides of Darkness (shareware demo 1.2, application `Warcraft II Demo 1.2`) never reaches its title screen under Systemless. During startup it puts up an alert and quits:

> Sorry, this program requires the Thread Manager Extension or System 7.5.

The runner halts at `ExitToShell` after roughly 20k instructions, having drawn nothing but that alert.

## Root causes

Three independent gaps in the generic Toolbox implementation, each of which stops the game at a later point than the last.

### 1. No Thread Manager

`Gestalt('thds')` is an unknown selector, so the game's availability check fails immediately. `_ThreadDispatch` ($ABF2) implements only `ThreadBeginCritical` and `ThreadEndCritical`; the rest of the selector table is missing.

### 2. `MenuInfo.menuProc` is always NIL

Inside Macintosh Volume I, I-352 makes the menu definition procedure part of the menu record's public contract, and I-366 states that the `'MENU'` resource's four `menuProc` bytes are "a placeholder for the handle to the procedure" which `GetMenu` replaces. Systemless writes zero and leaves it there.

Warcraft II calls the MDEF directly while sizing its menu bar, which dereferences address zero and jumps into the boot exception vector table:

```
00864692  MOVEA.L d16(A2),A2   ; A2 = theMenu->menuProc  -> $00000000
00864698  _HGetState
0086469E  _HLock
008646B6  _SetPt               ; hitPt := (0,0)
008646B8  MOVEA.L (A2),A3      ; A3 = *menuProc -> reads address 0
008646CC  JSR (A3)             ; jumps to $40810000
```

The pushed frame — message word, `theMenu`, `VAR menuRect`, `hitPt`, `VAR whichItem` — matches the definition procedure's signature exactly.

### 3. Sound Manager callback trampolines are written once

The Sound Manager's callback trampoline is allocated lazily and has its opcodes written only at allocation time, then reused for every later dispatch. A sound callback runs at interrupt time against whatever the application has since done to the heap. Warcraft II overwrites that block between callbacks, so a later `callBackCmd` jumps into what is by then sample data:

```
[CPU] IllegalInstruction: $1846 at PC=$00BA1834
```

## Expected behaviour

Warcraft II launches, reaches the shareware main menu, enters the Human Hillsbrad campaign, and plays — matching BasiliskII driven with the same input sequence.

## Acceptance criteria

- [ ] `Gestalt('thds')` reports `gestaltThreadMgrPresent`, and `_ThreadDispatch` implements the selector table declared by MPW `Interfaces/CIncludes/Threads.h`. Cooperative style only; `kPreemptiveThread` returns `paramErr`, which is what a 68K machine with no MP services would do.
- [ ] Every `MenuInfo.menuProc` holds a handle to callable code, and a direct call with `mSizeMsg`, `mChooseMsg` or `mDrawMsg` behaves per Inside Macintosh Volume I, I-352.
- [ ] Sound Manager callback trampolines survive a guest that overwrites the block between dispatches.
- [ ] The game runs from launch through real gameplay — select a Footman and order it to move — rather than stopping at the title screen.
- [ ] Rendering matches a BasiliskII reference run of the same input sequence.
- [ ] `cargo test` passes with no new failures, with regression coverage for each behaviour above.
- [ ] No per-game special cases: every fix is a generic Toolbox behaviour cited to Inside Macintosh or the MPW interfaces.

benletchford:
Warcraft II: Tides of Darkness (shareware 1.2) went from refusing to launch to playing the Human Hillsbrad campaign. Three generic Toolbox gaps stood between those two states, each hidden behind the previous one.

Closes #108

## 1. Cooperative Thread Manager

`Gestalt('thds')` was unknown, so the game's availability check failed and it quit with *"Sorry, this program requires the Thread Manager Extension or System 7.5."* before drawing anything.

Inside Macintosh: Operating System Utilities 1994, p. 2448 documents only the gate — `gestaltThreadMgrAttr` bit 0, `gestaltThreadMgrPresent`. The routine selectors come from MPW `Interfaces/CIncludes/Threads.h` (Universal Interfaces 2.0a3, ETO #16, 11 November 1994), which declares each entry point as `THREEWORDINLINE(0x303C, <selector>, 0xABF2)`. The full table is now implemented and cited in the trap comment, with the selector's high byte giving the Pascal frame size in stack words.

Systemless runs the guest on one host thread and switches contexts entirely inside `_ThreadDispatch`, so the whole caller-visible register file can be banked with no host concurrency involved. `kCooperativeThread` is the only supported style; `kPreemptiveThread` returns `paramErr`.

## 2. `MenuInfo.menuProc` was always NIL

Inside Macintosh Volume I, I-352 makes the menu definition procedure part of the menu record's public contract, and I-366 is explicit that the resource's four bytes are "a placeholder for the handle to the procedure" which `GetMenu` **replaces**. Systemless wrote zero and left it there, so an application calling the MDEF directly dereferenced address zero:

```
00864692  MOVEA.L d16(A2),A2   ; A2 = theMenu->menuProc  -> $00000000
008646B8  MOVEA.L (A2),A3      ; A3 = *menuProc          -> reads address 0
008646CC  JSR (A3)             ; jumps to $40810000
```

Every menu record now gets a handle to a ten-byte guest stub:

```
MOVE.W  #$FEFC,D0     ; private Systemless trampoline selector
_Pack8                ; re-enters the HLE
RTD     #18           ; pop the MDEF's Pascal frame
```

`_Pack8` is already the dispatch trap reserved for guest-to-HLE trampoline returns (`$FEFE` finalises an AppleEvent handler call), so `$FEFC` names the synthetic `'MDEF' 0` in the same private selector space. `mSizeMsg` shares one code path with `CalcMenuSize` — I-352 and I-361 define them as the same operation — and `mChooseMsg` hit-tests `hitPt` against `menuRect`, returning 0 for a miss, a disabled menu, a divider or a disabled item.

Telling the resource placeholder apart from a handle an application installed itself uses the same "does the entry look callable" test the Window Manager already applies to a guest `'WDEF'`.

## 3. Sound Manager callback trampolines were written once

The callback trampoline was allocated lazily with its opcodes written only at allocation time. A sound callback runs at interrupt time against whatever the application has since done to the heap, and Warcraft II clobbers that block between callbacks, so a later `callBackCmd` jumped into what had become sample data:

```
[CPU] IllegalInstruction: $1846 at PC=$00BA1834
```

The opcodes are now rewritten on every dispatch, for both the command-callback and file-completion trampolines — which is what the Window Manager's WDEF trampoline already does for the same reason.

## Also included

- **SANE `FOD2B`** (`$0009`): decimal-record to binary conversion, plus `CStr2Dec` (Pack7 selector 4) sharing its scanner with `Str2Dec`. Apple Numerics Manual (1988), Decimal-to-Binary Conversion.
- **`PBHGetVInfo` enumeration**: a positive `ioVolIndex` past the single mounted boot volume now ends the walk with `nsvErr` instead of re-reporting the boot volume forever. Files 1992, 2-145.

## Validation

Driven from launch through real gameplay, the game now runs to completion — 1,296,272,327 instructions with no halt — reaching the shareware main menu, the Human Hillsbrad briefing, the live mission, then selecting a Footman and ordering it to move.

Compared frame by frame against BasiliskII driven with the same input sequence, **inside the game's own 640x480 canvas the two runtimes are pixel identical**.

The whole-frame strict-parity figures — 63.98% at the main menu, 80.26% at the campaign briefing, and 51-53% in the mission — are accounted for entirely by the letterbox around that canvas. Warcraft II runs full screen at 640x480 centred in the 800x600 device and never draws the surround, so those pixels keep the startup blanking fill of colour index 255, which is black in the standard Macintosh colour table but white in the Windows-ordered palette the game installs for its cross-platform art. BasiliskII renders black there, Systemless white. That is a separate, well-scoped gap and I'll raise it on its own.

The startup frame is limited instead by Dialog Manager chrome: Systemless has no Finder desktop, menu bar or tab control to draw behind the game's options dialog.

### Tests

`cargo test` is green — 2712 lib tests, no failures — with seven new regression tests:

- `threaddispatch_newthread_creates_ready_cooperative_thread_and_consumes_mpw_frame`
- `threaddispatch_newthread_rejects_preemptive_style_with_param_err`
- `threaddispatch_get_and_set_thread_state_track_the_ready_queue`
- `threaddispatch_switcher_and_terminator_are_recorded_per_thread`
- `newmenu_installs_a_callable_menu_definition_procedure_handle`
- `menu_def_proc_size_message_fills_menu_width_and_height`
- `menu_def_proc_choose_message_reports_the_item_under_the_point`

Two other games that previously diverged from their BasiliskII references now match as well.

`cargo fmt` clean on every file touched. No per-game special cases: every change is a generic Toolbox behaviour cited to Inside Macintosh or the MPW interfaces.

### DIFF
--- a/src/runner.rs
+++ b/src/runner.rs
@@ -2230,12 +2230,9 @@ impl FixtureRunner {
         // leaves A7 on the four-byte FMOutPtr result slot. Allocate this only
         // after reserving the application zone header so it remains live.
         let swap_font_trampoline = self.bus.alloc(6);
-        self.bus
-            .write_word(swap_font_trampoline, 0x205F); // MOVEA.L (SP)+,A0
-        self.bus
-            .write_word(swap_font_trampoline + 2, 0xA901); // FMSwapFont
-        self.bus
-            .write_word(swap_font_trampoline + 4, 0x4ED0); // JMP (A0)
+        self.bus.write_word(swap_font_trampoline, 0x205F); // MOVEA.L (SP)+,A0
+        self.bus.write_word(swap_font_trampoline + 2, 0xA901); // FMSwapFont
+        self.bus.write_word(swap_font_trampoline + 4, 0x4ED0); // JMP (A0)
         self.bus.write_long(addr::J_SWAP_FONT, swap_font_trampoline);
 
         // Set CPU state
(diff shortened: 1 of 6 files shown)


benletchford:
Superseded — closing without merging.

While I was working, equivalent fixes for all three root causes landed on `master` independently (synthetic `'MDEF' 0` shim with real `'MDEF'` resource loading, `_ThreadDispatch` `NewThread`/yield/critical sections, `Gestalt('thds')`, and `alloc_synthetic` for HLE trampolines so the guest can no longer clobber them).

Verified against current `master`: Warcraft II runs the full sequence end-to-end — launch, shareware main menu, Human Hillsbrad briefing, live mission, select a Footman, order it to move — 1,299,168,793 instructions with no halt.

Current `master` is also **better** than this branch, so merging it would regress parity. Against a BasiliskII reference run of the same input sequence:

| frame | this branch | `master` |
| --- | ---: | ---: |
| menu | 63.98% | **100.00%** |
| campaign | 80.26% | 80.92% |
| gameplay | 51.76% | **87.76%** |
| footman_selected | 51.25% | **87.25%** |
| after_move | 53.27% | **89.27%** |

`master` also fixes the letterbox this branch's description called out as future work.

Two genuine gaps from this branch survive and are worth landing on their own; I'll open a focused PR for them:

1. `_ThreadDispatch` implements six selectors of the table in MPW `Threads.h`. The other sixteen — including `SetThreadSwitcher` ($070A), which Warcraft II calls — return `paramErr`.
2. The synthetic `'MDEF' 0` shim pops its frame and returns without doing anything, so a direct `mSizeMsg` leaves `menuWidth`/`menuHeight` untouched and `mChooseMsg` never reports an item. Inside Macintosh Volume I, I-352 defines both.

Issue #156, opened 2026-07-29
Design desktop save storage and host-document authorization


rlanday:
## Problem

The desktop runner currently persists guest-created files next to the launched archive under `.systemless/saves/<archive-name>/`.

For example:

```text
/Games/Example.sit
/Games/.systemless/saves/Example/
```

This naturally separates two archives with the same filename when they are in different directories, but it is surprising on modern systems and can be impossible or undesirable when the input lives in a read-only, protected, removable, or shared location such as `/Applications`. Even when the source directory happens to be writable, application-generated state generally should not be written beside an application or archive without an explicit user choice.

There is a related but distinct requirement: users may want Classic applications to open and modify host documents they already have. Automatically storing private emulator state and granting access to user documents should not use the same policy.

### Private automatic persistence

Store the desktop runner's private save snapshots in the platform's per-user application-data location:

- macOS: `~/Library/Application Support/Systemless/saves/...`
- Windows: an appropriate `LOCALAPPDATA`/known-folder location
- Linux: `$XDG_DATA_HOME/systemless` or its standard fallback

The key used beneath `saves/` needs deliberate design.

Using only the archive basename can collide. These two inputs would both map to `saves/Example/`:

```text
/Games/Example.sit
/Downloads/Example.sit
```

Using the complete source path, or a hash of it, distinguishes them. However, moving or renaming an archive then changes its identity. Moving `/Games/Example.sit` to `/Applications/Example.sit` could make its existing saves appear to disappear.

Content-derived identity is stable across moves, but an update or patch changes the archive contents and therefore its identity. Application metadata, executable names, creator codes, and per-install identifiers have different collision and continuity tradeoffs. The implementation should decide explicitly which behavior users should experience when an archive is moved, renamed, copied, or upgraded.

### Existing host documents

Use explicit native authorization rather than exposing arbitrary host paths:

- one-shot import through a native Open panel, or
- an explicitly selected host folder mounted as a guest-visible Classic volume.

A sandboxed macOS build would need security-scoped bookmarks to retain persistent access. Other platforms need equivalent user-authorized folder/document handling. The guest should receive Classic volume references, directory IDs, filenames, forks, and Finder metadata rather than host paths or host permission details.

## Current fork-preserving representation

The current desktop save store does not depend on the host filesystem supporting Classic resource forks. Each saved guest file is represented by a host directory containing:

```text
metadata.json
data.fork
resource.fork
```

- `data.fork` contains the guest data fork as ordinary bytes.
[...]

PR #157, opened 2026-07-29
fix: return canonical Pascal Booleans from Standard File

### LINKED ISSUE #179: Standard File replies use non-canonical Pascal true values
## Observed behavior

Successful old and modern Standard File replies encode `good`, `sfGood`, and `sfReplacing` as `$FF`. Applications that compare these one-byte Pascal Boolean fields directly with the ordinal value `1` reject otherwise successful replies. In 3 in Three, saving appears to complete but the application rejects `SFReply.good`, so no file is created.

## Expected behavior

Successful reply fields should contain canonical Pascal `TRUE` (`1`), with the seven non-value bits clear. Cancellation should remain `FALSE` (`0`).

## Reproduction

1. Launch 3 in Three.
2. Save a document through the original `SFPutFile` interface.
3. Attempt to open the saved document.
4. Observe that the application rejects the successful reply before performing File Manager I/O.

Inside Macintosh Volume I, p. I-86 defines `BOOLEAN` as a one-byte type whose value is stored in bit 0.

rlanday:
## Summary

- return canonical Pascal `TRUE` (`1`) in the one-byte Boolean fields of old and modern Standard File replies
- keep all non-value bits clear rather than encoding true as `$FF`
- apply the same representation to `sfReplacing`
- update old and modern Get/Put regression tests to assert the documented representation

## Why

Inside Macintosh, Volume I, p. I-86, “Pascal Data Types,” specifies `BOOLEAN` as a one-byte type whose value is stored in bit 0. Standard File’s `good`, `sfGood`, and `sfReplacing` fields are Boolean record fields.

The previous `$FF` representation made bit 0 true but also set all seven non-value bits. Code compiled with the assumption that a Boolean record field contains the canonical ordinal value of `TRUE` can compare the byte directly with `1` and reject `$FF`.

3 in Three exposes this with the original `SFPutFile` interface. The Save dialog accepted a filename, but the application rejected the returned `SFReply.good` before calling the File Manager. Consequently no file was created, and the next Open dialog reported “No matching files.”

With this change, the same interactive test produced the expected sequence:

- `PBCreate("Boolean Test")`
- `PBOpen` followed by a 498-byte `FSWrite`
- desktop persistence of the created file
- a subsequent Open listed the file
- selecting it produced a successful `PBOpen` and 498-byte `PBRead`

The application then reaches a separate post-open Yes/No dialog, confirming that Standard File selection and File Manager I/O have completed. That downstream dialog issue is outside this PR.

## Scope

This changes only the representation of successful Standard File Boolean fields. Cancellation remains `FALSE` (`0`), and the reply layouts, stack behavior, WDRefNums, FSSpecs, file filtering, and dialog tracking are unchanged.

The desktop save-directory policy is deliberately not part of this PR; that design is tracked separately in #156.

## Tests

- `cargo test --lib standard_get_file` — 7 passed
- `cargo test --lib standard_put_file` — 4 passed
- `cargo test --lib sf_get_file` — 3 passed
- `cargo test --lib sf_put_file` — 3 passed
- `cargo test --lib` — 2,771 passed, 3 ignored
- interactive 3 in Three Save → Open verification described above

Closes #179

### DIFF
--- a/src/trap/toolbox.rs
+++ b/src/trap/toolbox.rs
@@ -184,7 +184,11 @@ fn standard_file_put_reply_old(bus: &mut MacMemoryBus, reply_ptr: u32, vref: i16
     if reply_ptr == 0 {
         return;
     }
-    bus.write_byte(reply_ptr, 0xFF); // good = TRUE
+    // Inside Macintosh, Volume I, p. I-86 ("Pascal Data Types") specifies
+    // BOOLEAN as one byte with its value in bit 0. Write canonical TRUE (1),
+    // leaving its seven non-value bits clear; callers may compare a record
+    // field directly with the ordinal value of TRUE.
+    bus.write_byte(reply_ptr, 1); // good = TRUE
     bus.write_byte(reply_ptr + 1, 0); // copy = FALSE
     bus.write_long(reply_ptr + 2, 0); // fType
     bus.write_word(reply_ptr + 6, vref as u16);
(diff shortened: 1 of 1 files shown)


benletchford (approved):
Validated on current master: the patch applies cleanly and all 33 focused Standard File tests pass.

PR #158, opened 2026-07-29
fix(window): erase newly exposed content when creating windows

### LINKED ISSUE #177: Newly visible windows leave exposed content pixels unerased
## Observed behavior

Creating a visible window draws its frame but leaves the exposed content region unchanged. Transparent black text can therefore disappear over stale black framebuffer pixels. In 3 in Three, the post-open confirmation prompt is missing while its Yes and No controls remain visible because the controls paint their own backgrounds.

## Expected behavior

After final z-order and visibility-region setup, the Window Manager should erase the newly exposed visible content using the port background pattern. Hidden windows must not alter the framebuffer.

## Reproduction

1. Launch 3 in Three and open a saved document.
2. Reach the confirmation `dBoxProc` window.
3. Observe that the controls render but the transparent prompt text does not because the content pixels were never initialized.

rlanday:
## Summary

- erase the exposed content of newly created visible windows with the port background pattern
- perform that erase only after the final z-order has established the window's visibility region
- cover `NewWindow`, `GetNewWindow`, `NewCWindow`, and `GetNewCWindow`
- preserve hidden-window behavior and the existing full-screen kiosk initialization path

## Problem

After opening a saved document in **3 in Three**, the application creates a visible `dBoxProc` window and immediately draws a confirmation prompt with QuickDraw's transparent text mode. Systemless drew the window frame but left its exposed content pixels untouched. Because the previous framebuffer contents were black, the black prompt changed no pixels and appeared missing. The Yes and No controls remained visible only because each control painted its own white background.

This is a Window Manager initialization bug, not a font, clipping, or dialog-layout bug. Focused tracing confirmed that the guest supplied valid text, font state, pen coordinates, `visRgn`, and `clipRgn`, and that nothing subsequently erased the prompt.

PR #157 is what made the saved-file path reliably reachable during testing, but this PR has no code dependency on it and is based directly on `master`.

## Classic Window Manager behavior

Inside Macintosh Volume I describes window creation in terms of the Window Manager's `PaintOne` operation:

- pp. I-278 and I-296: `PaintOne` draws the frame, erases newly exposed content with the window's background pattern, and adds the exposed content to `updateRgn`
- pp. I-282–I-283: `NewWindow` draws a visible window and generates an update event for its content

The initial erase is observable even when an application draws immediately rather than waiting for the update event, because transparent QuickDraw text preserves destination pixels outside the glyph mask.

The implementation performs the erase after applying the `behind` parameter so QuickDraw uses the final visibility region and does not paint through windows in front. Full-screen windows remain on Systemless's existing deliberate black kiosk stage rather than flashing a classic white desktop.

## Testing

- `cargo test --lib trap::window::tests` — 168 passed
- `cargo test --lib` — 2,773 passed, 3 ignored
- debug application rebuilt and verified interactively with the saved-document confirmation window

Regression tests verify that:

- visible `NewWindow` content is erased to the default white background before the call returns
- hidden `NewWindow` creation does not alter the framebuffer
- the existing full-screen kiosk test continues to preserve the black stage

## Visual result

Before/after screenshots show the same confirmation window. Before this fix, only the controls' self-painted white backgrounds are visible.
Screenshot 2026-07-29 at 12 51 09 PM

After the fix, the erased white content area makes the complete transparent-mode prompt visible as intended.
Screenshot 2026-07-29 at 12 46 39 PM

Closes #177

### DIFF
--- a/src/trap/window.rs
+++ b/src/trap/window.rs
@@ -4,6 +4,7 @@ use crate::cpu::{CpuOps, Register};
 use crate::memory::{MacMemoryBus, MemoryBus};
 use crate::trap::dispatch::{DrawOldState, PortDrawState, QueuedEvent};
 use crate::trap::quickdraw::RegionBooleanOp;
+use crate::trap::types::{Rect, ShapeOp};
 use crate::Result;
 use std::sync::OnceLock;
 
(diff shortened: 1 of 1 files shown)


benletchford (approved):
Validated on current master: the patch applies cleanly and all 20 focused new-window tests pass, including visible, hidden, z-order, and kiosk behavior.

PR #164, opened 2026-07-29
fix(window): redraw window pictures during update scans

### LINKED ISSUE #163: CheckUpdate delivers update events for windows with windowPic
## Observed behavior

A visible window with a nonempty update region and a picture installed through `SetWindowPic` still receives an application-visible update event. The application can then redraw unrelated content into a window whose static picture should have been maintained by the Window Manager.

## Expected behavior

`CheckUpdate` should scan visible dirty windows from front to back. When a dirty window has a non-NIL `windowPic`, the Window Manager should redraw that picture with the necessary update-region clipping, clear the update, and continue scanning without returning an update event for that window. It should return an update event only for the next dirty window without a picture.

This follows *Macintosh Toolbox Essentials* (1992), p. 4-116, and *Inside Macintosh Volume I* (1985), p. I-296.

## Reproduction

1. Create a visible window and install a picture with `SetWindowPic`.
2. Invalidate the window or show it so its update region is nonempty.
3. Call `CheckUpdate` or retrieve the next Toolbox event with the update mask enabled.
4. Observe an update event for the pictured window instead of an automatic picture redraw and continued scan.

benletchford:
## Root cause

Systemless queued and delivered application-visible update events for dirty windows with a non-NIL `windowPic`. Classic Window Manager update scans instead redraw those pictures internally and continue scanning front-to-back until they encounter an ordinary dirty window. This caused applications that use `SetWindowPic` for background artwork to repaint the wrong window.

## Fix

- redraw dirty visible `windowPic` windows through the normal `DrawPicture` path
- preserve front-to-back update ordering for `CheckUpdate` and Event Manager calls
- stop at the first ordinary dirty window and deliver its update event
- add regression coverage for picture rendering, region clearing, event suppression, and overlapping-window ordering

## Verification

- `cargo test --quiet --lib --no-default-features window_picture`
- `cargo test --quiet --lib --no-default-features` (2764 passed, 3 ignored)
- `cargo check --quiet --no-default-features`
- runtime reproduction now renders one centered title instead of copying it into both the foreground and background windows

Closes #163

### DIFF
--- a/src/trap/event.rs
+++ b/src/trap/event.rs
@@ -363,6 +363,11 @@ impl super::TrapDispatcher {
     ) -> (u16, u32, i16, i16, u16, bool) {
         self.enqueue_open_application_event_if_needed(event_mask);
         self.enqueue_auto_key_if_due(event_mask);
+        let preferred_update = if Self::event_matches_mask(event_mask, 6) {
+            self.service_window_picture_updates(cpu, bus)
+        } else {
+            None
+        };
         if let Some(event) = self.dequeue_pending_native_menu_event(event_mask) {
             return (
                 event.what,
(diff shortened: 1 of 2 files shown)


benletchford:
Extended integration verification now covers the application beyond its centered title: after the `windowPic` redraw, the front window correctly retains a complex visible-region hole over the pictured background, tutorial setup renders cleanly, live 16-color gameplay remains coherent through 1,200 guest ticks, and map input continues to advance the simulation. This rules out the apparent mid-animation partial frames as a repeated background overdraw regression.

PR #200, opened 2026-07-30
fix: resolve bitmap strikes through FOND associations

### LINKED ISSUE #201: Resolve bitmap strikes through FOND associations
Systemless currently treats bitmap FONT and NFNT resource IDs as though they always encode the font family and point size. Classic NFNT strikes may instead use arbitrary resource IDs mapped by the containing FOND association table. As a result, applications such as System’s Twilight can resolve an unrelated built-in face or scale the wrong strike, producing incorrect text faces, sizes, and metrics.

Parse the FOND association table, register plain bitmap strikes under its family and point-size mapping regardless of resource load order, and retain the legacy family × 128 + point-size fallback only for standalone old-style FONT resources.

rlanday:
## Summary

- parse the mandatory association table that follows a FOND FamRec
- register bitmap NFNT/FONT strikes under the family and point size named by that association
- make registration independent of whether the FOND or bitmap resource loads first
- retain the legacy family * 128 + pointSize decoding only for standalone old-style FONT resources

## Problem

System's Twilight ships bitmap font strikes whose NFNT resource IDs are arbitrary. Systemless treated those IDs as though they encoded the family and point size directly. The requested family therefore failed to resolve and QuickDraw fell back to an unrelated built-in face or scaled the wrong strike.

This was visible in two ways:

- narrative text used a much smaller point size than requested
- speech-bubble text used the wrong face and metrics

## Why this is the Classic Mac OS behavior

Inside Macintosh: Text describes the FOND association table as the mapping from a font family and size/style to its FONT, NFNT, or sfnt resource. NFNT resource IDs do not carry the old family/size encoding.

The implementation parses the 52-byte FamRec header plus the mandatory association table, then maps each plain bitmap strike to the FOND family ID and point size. Styled associations are not installed as plain faces because the current renderer synthesizes QuickDraw styles. Standalone old-style FONT resources keep the historical resource-ID fallback.

References are included in the source comments:

- Inside Macintosh: Text (1993), pp. 4-47–4-48
- Inside Macintosh: Text (1993), pp. 4-95–4-96

## Resource-loading behavior

Resource forks do not guarantee that FOND and NFNT resources are observed in a convenient order. The loader therefore retries the association in both directions:

- when a bitmap strike arrives, look for an already-loaded FOND
- when a FOND arrives, look for already-loaded bitmap strikes

The resource bytes remain guest-provided; this change does not bundle proprietary font data.

## Tests

Added regression coverage for:

- arbitrary NFNT IDs mapped through a FOND association
- registration under the associated family and point size
- FOND-before-NFNT loading
- NFNT-before-FOND loading
- the standalone old-style FONT fallback

## Visual verification

Before:
Screenshot 2026-07-30 at 11 36 24 AM
Screenshot 2026-07-30 at 11 36 40 AM

After:
Screenshot 2026-07-30 at 11 33 24 AM
Screenshot 2026-07-30 at 11 32 54 AM

The crop and viewport behavior visible around the game is separate work and is intentionally not part of this PR.

Closes #201

### DIFF
--- a/src/quickdraw/fonts/mod.rs
+++ b/src/quickdraw/fonts/mod.rs
@@ -217,20 +217,80 @@ fn resource_word(data: &[u8], offset: usize) -> Option<u16> {
     ]))
 }
 
-/// Decode a classic `FONT`/`NFNT` bitmap strike whose resource ID uses the
-/// standard `family * 128 + pointSize` encoding. The resource bytes stay in
-/// the user's application; Systemless only expands their 1-bit glyph bitmap
-/// into the same runtime coverage representation used by its built-in faces.
+/// One entry from a font family (`FOND`) resource's association table.
+///
+/// `font_resource_id` identifies an `NFNT`, `FONT`, or `sfnt` resource;
+/// unlike old-style `FONT` IDs, an `NFNT` ID does not encode its family or
+/// point size. *Inside Macintosh: Text* (1993), pp. 4-47–4-48 and 4-95–4-96.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(crate) struct FondAssociation {
+    pub family_id: i16,
+    pub size: i16,
+    pub style: u16,
+    pub font_resource_id: i16,
+}
+
+/// Decode the mandatory association table that immediately follows a
+/// `FOND` resource's 52-byte `FamRec` header.
+pub(crate) fn parse_fond_associations(
+    fond_resource_id: i16,
+    bytes: &[u8],
+) -> Option<Vec<FondAssociation>> {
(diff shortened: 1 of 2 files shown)


benletchford (approved):
The FOND association parsing and load-order-independent strike registration match the documented Font Manager model, retain the standalone FONT fallback, and are covered by focused tests. CI is green across all required jobs.

rlanday:
The juxtaposition of the game’s title, the name of the emulator, and the text in the image is pretty funny

benletchford:
> The juxtaposition of the game’s title, the name of the emulator, and the text in the image is pretty funny

Thanks for your brilliant contributions :)

PR #265, opened 2026-08-01
fix: render missing Times sizes from host outlines

rlanday:
## Summary

QuickDraw applications request a font family and point size. When Systemless had no exact bitmap strike for that request, it could only scale a compatible bundled bitmap. EV Override requests 24-point Times for its introduction; the old fallback selected an unrelated-looking bitmap face with different metrics, producing visibly wrong text and layout.

This adds an exact-size host-outline fallback while keeping guest and bundled resources authoritative.

The selection order is now:

1. exact application resource or explicit override;
2. exact compatible Systemless bitmap strike;
3. exact-size accepted host outline;
4. the existing integer-scaled compatible bitmap fallback.

This follows the size-selection model described in *Inside Macintosh: Text*, pp. 4-18–4-23: an available outline can produce the requested size directly, while bitmap strikes require a size fallback/scaling search.

## Implementation

- Adds `ab_glyph` for native outline parsing and rasterization. Font files remain host assets and are not redistributed.
- Locates Times with targeted platform-specific lookup rather than a full system-font scan:
  - macOS: known system and supplemental Times paths;
  - Windows: the system and per-user font directories;
  - Linux: cached Fontconfig lookup, followed by a small fixed-path fallback.
- Uses the same family preference on every platform: Times, Times New Roman, Liberation Serif, then Nimbus Roman. Nimbus Roman is only a last-resort metric-compatible Linux fallback.
- Uses Canvas2D in WebAssembly, with the same family preference followed by the browser's generic serif fallback.
- Lowers every selected outline once into Systemless's existing `FontFace` representation and caches it, so measurement and drawing use the same metrics and glyph source.
- Thresholds coverage to monochrome pixels. This does not silently opt a Classic application into later font smoothing.
- Centralizes Mac Roman translation and renders extended characters from the selected face. In particular, byte `$C9` remains one horizontal-ellipsis character rather than three ASCII periods, preserving TextEdit character counts, line breaking, advance, and glyph selection.

Only Times is admitted as a host-outline family in this change. If no accepted host font is present, the existing bitmap fallback remains available.

### Why Times only?

A host-font mapping changes layout, not just appearance: different advances and vertical metrics can change line breaks, clipping, and dialog geometry. A generic name substitution would therefore be unsafe. Helvetica is not normally present on Windows and Arial is not identical; Courier and Courier New can differ in vertical metrics; and classic screen families such as Chicago, Geneva, and Monaco are bitmap-oriented, so an arbitrary modern outline can be less compatible than Systemless's existing bitmap scaling. Linux frequently supplies metric-compatible replacements rather than the original families as well.

Times is the deliberately narrow first mapping because this change has direct evidence for it: EV Override requests Times at a size for which Systemless has no exact strike, the old scaled fallback is visibly wrong, and the result was compared with the application under System 7.5.3. The provider is structured so additional families can be admitted later, but each mapping should specify acceptable platform equivalents and include layout tests before it changes application typography.

### Font selection

Before, the missing Times strike fell through to the scaled bitmap fallback.

Screenshot 2026-07-31 at 5 22 33 PM

After, the requested Times size is rasterized from the host outline.

Screenshot 2026-07-31 at 5 12 28 PM

### Extended Mac Roman punctuation

Before: The two Mac Roman `$C9` ellipses are missing.

Screenshot 2026-07-31 at 7 18 54 PM

After: The two ellipses are now rendered by the selected Times face.

Screenshot 2026-07-31 at 7 12 57 PM

### WebAssembly

Screenshot 2026-07-31 at 7 55 41 PM

The full local website frontend renders the same Times text and both ellipses through the Canvas2D provider.

For comparison, the same introduction was also checked in Basilisk II running System 7.5.3. It uses monochrome Times text without smoothing.

Screenshot 2026-07-31 at 7 10 38 PM

Testing in Basilisk II proved that Systemless is still not rendering Times with ideal rasterization (see e.g. the horizontal bars in the lowercase letter “e” on the macOS frontend; also note the heavier font weight in the WebAssembly frontend). Adding TrueType hinting is therefore listed under “Follow-up work.”

## Testing

- `cargo fmt --check`
- `cargo test --lib`: 2,861 passed, 3 ignored, 0 failed
- `cargo check --target wasm32-unknown-unknown --no-default-features --lib`
- `cargo check --target x86_64-unknown-linux-gnu --no-default-features --lib`
- `cargo check --target x86_64-pc-windows-msvc --no-default-features --lib`
- native macOS EV Override smoke test
- browser-level Canvas probe using the real Systemless WASM font provider:
  - `PASS: browser Times 24 rendered Mac Roman $C9 as 20x4, advance 24, 24 lit pixels; face metrics ascent 18, descent 6, leading 0`
- full `systemless-web` frontend smoke test with EV Override: Times rendered correctly and both ellipses were visible

The Linux and Windows paths are compile-checked here; the macOS and WebAssembly paths additionally received live rendering tests.

## Follow-up work

- execute TrueType hinting, or otherwise improve monochrome outline rasterization, so small text more closely matches classic hinted strikes;
- model configurable Mac OS 8.5-style font smoothing separately;
- admit additional host outline families only with explicit family mappings and layout tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2197,6 +2197,7 @@ dependencies = [
 name = "systemless"
 version = "0.11.1"
 dependencies = [
+ "ab_glyph",
  "clap",
  "cpal",
  "hfs-reader",
@@ -2216,6 +2217,8 @@ dependencies = [
  "tempfile",
  "thiserror 2.0.18",
  "tracing",
+ "wasm-bindgen",
+ "web-sys",
  "winit",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,6 +51,20 @@ gui = [
 # Off by default so it stays out of the published public API and docs.
 test-support = []
 
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+ab_glyph = "0.2.32"
+
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+wasm-bindgen = "0.2.114"
+web-sys = { version = "0.3.91", features = [
+    "CanvasRenderingContext2d",
+    "Document",
+    "HtmlCanvasElement",
+    "ImageData",
+    "TextMetrics",
+    "Window",
+] }
+
 [target.'cfg(target_os = "macos")'.dependencies]
 objc2 = { version = "0.5.2", optional = true }
 objc2-app-kit = { version = "0.2.2", optional = true, features = ["NSApplication", "NSBitmapImageRep", "NSCell", "NSGraphics", "NSImage", "NSImageRep", "NSMenu", "NSMenuItem"] }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -57,6 +57,7 @@ pub mod display;
 mod error;
 pub mod game;
 pub mod loader;
+mod mac_roman;
 pub mod machine_profile;
 pub mod managers;
 pub mod memory;
--- a/src/mac_roman.rs
+++ b/src/mac_roman.rs
@@ -1,0 +1,54 @@
+//! Mac Roman byte-to-Unicode translation shared by guest text consumers.
+//!
+//! Guest strings remain bytes until a consumer needs a Unicode scalar for a
+//! host outline API. In particular, byte `0xC9` is one horizontal-ellipsis
+//! character, not three ASCII periods; preserving that distinction keeps its
+//! advance and bitmap consistent with the selected face.
+
+const HIGH: [char; 128] = [
+    '\u{00C4}', '\u{00C5}', '\u{00C7}', '\u{00C9}', '\u{00D1}', '\u{00D6}', '\u{00DC}', '\u{00E1}',
+    '\u{00E0}', '\u{00E2}', '\u{00E4}', '\u{00E3}', '\u{00E5}', '\u{00E7}', '\u{00E9}', '\u{00E8}',
+    '\u{00EA}', '\u{00EB}', '\u{00ED}', '\u{00EC}', '\u{00EE}', '\u{00EF}', '\u{00F1}', '\u{00F3}',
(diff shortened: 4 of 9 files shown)


benletchford:
Thanks for the detailed investigation and the strong visual evidence here. The underlying problem is real, but after reviewing the implementation our preferred path is much narrower: add the missing Systemless Ironbark 24-point strike.

Systemless already maps Classic Times to our New York-compatible Ironbark face. We currently provide Ironbark/New York strikes at 12, 14, and 18 points, but not 24. A Times 24 request therefore falls back to scaling the 12-point strike by 2×, which explains the chunky result in EV Override.

For this regression, we would prefer to:

1. add an original OFL-licensed Ironbark/New York 24-point strike;
2. register it as the exact `FONT_NEWYORK` 24-point face; and
3. cover EV Override's advances, wrapping, and raster output with focused tests.

That gives us deterministic typography on macOS, Windows, Linux, and WebAssembly, stays within the project's existing clean-room/OFL font model, and keeps the solution proportional to the actual gap.

By contrast, host outlines make layout and raster output depend on the installed font and rasterizer. The browser's generic `serif` fallback is particularly variable, and eagerly lowering proprietary installed fonts into reusable cached bitmap alphabets also creates licensing uncertainty that “the font file is not redistributed” does not completely resolve. It would additionally commit us to completing the guest-visible outline Font Manager behavior and bounding the new caches.

We do not need to expand this fix into new bundled Mac Roman artwork either. If exact single-character ellipsis semantics are needed later, they can be handled separately by composing the active Ironbark period glyph three times while retaining one guest byte.

For those reasons, we do not plan to merge this PR in its current form. A replacement change adding Ironbark 24 would be welcome.

rlanday:
> A replacement change adding Ironbark 24 would be welcome.

I’ll look into trying to do this using the existing sizes as a reference.

PR #272, opened 2026-08-02
fix: support Warcraft II palette and icon probes

### LINKED ISSUE #270: fix: support the compact GetGray PaletteDispatch selector
## Summary

A Warcraft II: Tides of Darkness startup replay issues `MOVE.W #$0C19,D0; _PaletteDispatch` for the Color QuickDraw `GetGray` routine. The HLE currently recognizes the documented `$1219` selector but not the compact C calling-convention form `$0C19`.

## Evidence

- Guest call site: `PC=$0081834C` loads `D0=$00000C19`, then executes `_PaletteDispatch`.
- Current Systemless trace reports `PaletteDispatch unknown selector=$03FF` and advances the stack by only two bytes.
- The replay then halts at `PC=$00885C02` with an illegal instruction, before the startup dialog can be dismissed.
- The observed stack contains three pointer-sized arguments followed by the Boolean result slot, matching the compact 12-byte GetGray call frame.

## Requested behavior

Implement the generic compact `GetGray` PaletteDispatch ABI (while preserving the documented selector path), balance the result stack correctly, and add focused coverage for selector `$0C19`.

### LINKED ISSUE #271: fix: preserve private IconDispatch stack frames
## Problem

Warcraft II: Tides of Darkness Demo 1.2 reaches two private `_IconDispatch` probes that use stack frames larger than the current generic HLE consumes:

- selector `$0613` is emitted with a 12-byte frame;
- selector `$0500` is emitted with a 10-byte frame.

The current fixed 8-byte pop leaves the saved return address misaligned. The guest then halts at `$03FFFDCC` instead of continuing into the live Hillsbrad mission.

## Requested behavior

Preserve the existing public selector-0 contract while consuming the observed private frames with selector-aware sizes, and add focused regression coverage for both probes.

The private selector/frame observations come from the Warcraft II shareware binary's direct execution trace; they are not being presented as public Icon Utilities API selectors.

benletchford:
## Summary

- Implement GetGray against the supplied graphics device, including indexed/direct-color quantization and the documented FALSE/unchanged result when no distinct intermediate color exists.
- Preserve both direct RGB and palette-entry GrafVars state across SaveFore, SaveBack, RestoreFore, RestoreBack, and port switches.
- Normalize Icon Utilities selectors against the published table and derive Pascal frame sizes from their public encoding.
- Add focused regression coverage for device-aware color selection, palette state round trips, and canonical/byte-swapped selector frames.

## Root cause

Warcraft II's startup path exercised Palette Manager and Icon Utilities dispatcher contracts that Systemless only approximated. The incomplete GetGray result handling and fixed IconDispatch frame size could misalign the saved return address before the live Hillsbrad mission.

## Validation

- `cargo test --lib --features test-support -q`: 2,867 passed, 3 ignored.
- The Warcraft II replay reaches the Human Hillsbrad map, selects a Footman, and visibly orders it to move.
- Historical BasiliskII oracle comparison passes the six-frame G3 parity floor; a fresh Docker-backed BasiliskII run was unavailable in this environment.

Closes #270
Closes #271

### DIFF
--- a/src/trap/dispatch.rs
+++ b/src/trap/dispatch.rs
@@ -606,6 +606,8 @@ pub(crate) struct ControlTrackingState {
 pub(crate) struct PortDrawState {
     pub fg_color: (u16, u16, u16),
     pub bg_color: (u16, u16, u16),
+    pub pm_fg_color: Option<(u32, i16)>,
+    pub pm_bg_color: Option<(u32, i16)>,
     pub bk_pat: [u8; 8],
     pub pn_loc: (i16, i16),
     pub pn_size: (i16, i16),
@@ -622,6 +624,8 @@ impl Default for PortDrawState {
         Self {
             fg_color: (0, 0, 0),
             bg_color: (0xFFFF, 0xFFFF, 0xFFFF),
+            pm_fg_color: None,
+            pm_bg_color: None,
             bk_pat: [0x00; 8],
             pn_loc: (0, 0),
             pn_size: (1, 1),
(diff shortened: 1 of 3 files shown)


benletchford (commented):
I found two design/correctness blockers in this draft.

1. The new GetGray path does not implement the documented contract. Imaging With QuickDraw (1994), pp. 4-81–4-82 says the graphics device determines the best available intermediate color, leaves foreGround unchanged when there is no distinguishable third color, and returns FALSE in that case. This implementation ignores device, always writes the arithmetic midpoint, and always returns TRUE. At minimum it needs device-aware availability/quantization and coverage for the FALSE/unchanged path.

2. The IconDispatch change hardcodes Warcraft II-observed values 0x0613 and 0x0500 and calls them private, but the public More Macintosh Toolbox selector table lists IconIDToRgn as 0x1306 and PlotIconID as 0x0005. These observations look byte-swapped, and the frame sizes follow the public signatures. Please normalize the selector encoding and derive handling from the public table/signatures rather than adding application-specific magic values.

The SaveFore/SaveBack/RestoreFore/RestoreBack paths should also preserve the documented ColorSpec value==1 palette-entry state instead of silently ignoring it. I am leaving this draft unmerged until those contracts are modeled generically.

Issue #273, opened 2026-08-02
Warlords II startup stalls after resource initialization


benletchford:
## Observed behaviour

The Warlords II demo now passes its initial named-resource and indexed-resource startup calls in Systemless, but it still does not reach its title/options screen. A bounded run reaches 2,400 guest ticks and leaves the checkerboard framebuffer unchanged while the guest continues executing its world-data unpacker.

## Root cause under investigation

The first failure exposed an incomplete `Get1NamedResource` contract: successful and missing lookups did not return the Handle/nil result in `A0`. The investigation also exposed two related generic contracts in this startup path:

- `GetIndResource` must preserve a resident master pointer when `SetResLoad(FALSE)` is active and the archive loader has already materialized the resource.
- Synthetic tool-trap entries must remain stable when guest code probes or writes through the old trap address.

The focused HLE tests for these contracts pass. With those changes, Warlords II reaches `Get1NamedResource('CODE', 'Main')` and then spends the bounded run in guest `uncompress_world` code without producing UI.

## Reproduction

Run the exact Warlords II Demo archive with the launch probe and allow it to run through the two startup checkpoints. The run completes without an HLE trap error, but captures only the unchanged checkerboard framebuffer:

- 4,131 total traps
- 747,862,478 guest instructions
- no windows or menus
- hot guest range `$00562B04-$00562B46`

A BasiliskII comparison is currently unavailable because the MPW/Docker launcher blocks on the unavailable Docker daemon; no emulator result is inferred.

## Expected behaviour

The demo should finish startup and show its title/options screen so its documented first-turn interaction path can be calibrated.

benletchford:
Qualification update: the exact Warlords II Demo now has an extended bounded probe. The canonical launch probe still reaches the resource initialization path, completes the indexed seg!/mem!/res! lookups and CODE Main lookup, then runs to tick 2400 with 4,131 traps, no window/menu, and an unchanged checkerboard framebuffer. A separate 4,000-tick continuation reaches tick 4600 after 1,662,244,079 guest instructions with the same 4,131 traps and still produces zero windows and menus. Guest tracing identifies the long-running uncompress_world and relocate_world routines around $00562B04-$00562C3E while their source pointers advance through the loaded packed stream. No additional HLE trap or generic missing behavior is evidenced, so this remains blocked before title/options pending a real relocation/unpack diagnosis.

benletchford:
[...]

PR #274, opened 2026-08-02
fix: preserve resident resources and stable trap startup state

benletchford:
## Summary

- return Handle/nil results from `GetNamedResource` and `Get1NamedResource` in `A0`
- distinguish resource-fork backing bytes from guest-resident resource data when `SetResLoad(FALSE)` is active
- reject signed-negative `NewPtr`/`NewHandle` sizes and negative `BlockMove` counts
- keep synthesized tool-trap entries stable and recover native old-trap calls from dispatcher-recorded call state
- return the documented `MaxMem` largest block in `D0` and application-zone growth in `A0`
- add focused regression coverage for unloaded/resident resources, arbitrary native handler entries, memory bounds, and trap contracts

## Root cause

Resource-fork parsing eagerly creates internal backing allocations, but those bytes are not necessarily resident through a guest resource handle. Treating every backing pointer as resident made `SetResLoad(FALSE)` ineffective. The Resource Manager now tracks residency explicitly and preserves a master pointer only when the resource was genuinely loaded.

A handler installed with `SetTrapAddress` may have any routine entry or stack-frame convention. The dispatcher now records the original A-line return PC and argument SP before entering a native handler, then validates the documented MPW/Think C jump-table form if that handler invokes its saved old `_LoadSeg` address. Recovery no longer reads A6 or requires a `JMP.L` prologue.

## Evidence

The Warlords II demo previously reached its `seg!`, `mem!`, and `res!` initialization and `Get1NamedResource('CODE', 'Main')` without an HLE error. Its remaining bounded run is a guest `uncompress_world` phase at `$00562B04-$00562B46`; issue #273 remains open for that later qualification boundary.

## Validation

- `cargo test --lib --features test-support -q`: 2,876 passed, 3 ignored
- focused `GetIndResource`, `LoadSeg`, named-resource, `MaxMem`, and negative-size tests pass

Refs #273

### DIFF
--- a/src/memory/bus.rs
+++ b/src/memory/bus.rs
@@ -450,6 +450,8 @@ pub struct MacMemoryBus {
     heap_ptr: u32,
     /// Systemless-owned executable/data stubs grow downward outside the guest heap.
     synthetic_ptr: u32,
+    /// Guest writes cannot modify synthesized ROM-like instruction stubs.
+    readonly_code_ranges: Vec<(u32, u32)>,
     /// Free list: maps aligned_size → stack of recycled addresses
     free_blocks: HashMap<u32, Vec<u32>>,
     /// Tracks the aligned size of each allocation (address → aligned_size)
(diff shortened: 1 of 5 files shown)


benletchford (commented):
This stack remains blocked by #272 and also has independent Resource Manager/LoadSeg design issues.

The GetIndResource change treats every nonzero pointer in the archive loader's internal table as already resident and force-populates the returned master pointer under SetResLoad(FALSE). Inside Macintosh distinguishes backing resource data from a resource loaded into the guest heap: with automatic loading disabled, an unloaded resource must return an empty handle, while a genuinely resident resource retains its pointer. The HLE needs an explicit residency state; using eager parser materialization as residency makes SetResLoad(FALSE) ineffective for indexed resources.

The empty-name fallback likewise equates an unnamed resource-map entry with a resource named the empty Str255 without a cited Resource Manager rule. Those are distinct encodings in a resource map and should not be conflated without oracle/reference evidence.

Finally, the new LoadSeg recovery reads A6+4 and treats it as a jump-table entry based on a narrow native-shim shape, while validating the installed handler by requiring its first instruction to be JMP.L. SetTrapAddress can install an arbitrary routine entry. Please replace this with validated generic jump-table/caller recovery rather than depending on one CRT frame.

The negative-size checks and MaxMem A0 correction are sound, but this PR should be split or revised before merging.

PR #282, opened 2026-08-02
fix: support Spectre Supreme startup and menu input

### LINKED ISSUE #281: Support Spectre Supreme demo SIZE partition and low-memory QuickDraw vectors
## Summary

The public Spectre Supreme Demo.sit archive reaches an early memory gate under Systemless instead of showing its authentic menu. Once that gate is cleared, the demo also relies on two classic low-memory/window contracts during startup and menu interaction.

Archive under test: spectre-supreme--Spectre-Supreme-Demo.sit
Executable: Spectre Supreme Demo/Spectre Supreme Demo

## Evidence

- The SIZE -1 resource declares a preferred partition of 1,843,200 bytes and a minimum of 750 KiB.
- Startup compares the application A5 world against the visible application zone; the current layout leaves the declared minimum unavailable and reaches the demo's “requires at least 750K of memory” path.
- The full-screen NewWindow path must deliver its initial updateEvt when the application defers its first frame until event delivery.
- Menu clicks call the documented JInitCrsr low-memory vector at $0814; a zero vector halts at address zero.

## Acceptance

The canonical play script should reach the authentic main menu, vehicle-selection dialog, live gameplay HUD, and demonstrate firing input without a runner halt. Focused unit coverage should protect the SIZE relocation, full-screen update event, and JInitCrsr vector behavior.

benletchford:
Closes #281.

## What changed

- Select every default `SIZE` launch from the preferred partition when it fits, otherwise the minimum, with no application-specific threshold.
- Lay out the application heap at the partition's low end, the stack below the fixed A5/image area, and cap `ApplLimit`, guest heap allocation, and Systemless callback storage at their real boundaries.
- Preserve and queue the initial `updateEvt` and save-under state for visible full-screen windows whose first frame is deferred until event delivery.
- Seed `JInitCrsr` ($0814, sourced from MPW `LowMemEqu.a`) with a callable `InitCursor` trampoline.
- Add regressions for preferred/minimum selection, insufficient memory, allocator isolation, full-screen update delivery, and direct `JInitCrsr` invocation.

## Qualification evidence

Spectre Supreme reaches its authentic main menu, vehicle-selection dialog, live gameplay HUD, and firing path. Holding Space changes the visible ammunition counter from 48 to 00 without a runner halt.

## Validation

- `cargo test --lib --features test-support -q` — 2,882 passed, 3 ignored
- `cargo test -q load_app_`
- `cargo test -q negative_`
- `cargo test -q partition_boundaries_keep_heap_and_synthetic_allocations_out_of_the_image`
- `cargo test -q init_app_seeds_callable_init_cursor_low_memory_vector`
- `cargo test -q full_screen_new_window_preserves_initial_update_event`


benletchford (commented):
The full-screen update-event restoration and JInitCrsr trampoline are useful generic changes, but the SIZE relocation policy is not ready to merge.

Inside Macintosh: Memory (1992), pp. 1-7–1-12 describes every application partition as heap at the low end, A5 world at the high end, and stack growing down between them. Processes (1994), pp. 1-5–1-6 says SIZE supplies preferred and minimum partition sizes. There is no system-level 750 KiB relocation threshold. Encoding Spectre Supreme's startup comparison as SIZE_RELOCATION_MINIMUM makes application-specific behavior part of the loader and still leaves smaller SIZE partitions with a non-classic layout.

Please model partition selection and placement generically for all SIZE resources (prefer preferred size when available, fall back to minimum, place the A5 world at the selected partition's high end, and keep stack/ApplLimit consistent). The 750 KiB case can remain a regression fixture, not a policy constant. Please also add the source for JInitCrsr=$0814 beside that low-memory vector.

rlanday:
Testing the new application-partition enforcement with **EV Override** exposed two pre-existing Resource/Memory Manager gaps that had been hidden by Systemless’s previous, effectively larger heap.

EV Override declares a preferred partition of 13,924,000 bytes and a minimum of 12,900,000 bytes. PR #282 correctly selects the preferred partition and constrains the application heap to it. I do not think that behavior should be reverted or that EV should receive an application-specific memory override.

The first issue is that Systemless eagerly materializes every resource from each additional resource file when it is opened. EV opens several large resource forks during startup, including EVO MAGMA, whose map contains 807 resources. This consumes a substantial part of the application partition before the resources are requested.

That differs from the documented Resource Manager behavior:

- *Inside Macintosh*, Volume I, p. I-104: “Resource data is normally read into memory when needed.”
- `OpenResFile`, p. I-115, reads the resource map and additionally loads resources whose `resPreload` attribute is set.
- `SetResLoad(FALSE)`, p. I-119, suppresses resource-data loading, including preload resources.

I have a local prototype that registers the resource-map entries immediately, materializes only preload resources when `ResLoad` is enabled, and loads ordinary resources on demand while preserving their original handles. That lets EV proceed much farther through startup.

It then exposes a second issue: Systemless does not reclaim unlocked, purgeable resource data when an allocation cannot fit. EV loads a large series of pictures and eventually fails a 391,680-byte allocation with its heap near the partition limit. It then displays:

> EV encountered an error while starting up, probably due to lack of memory. Please increase EV’s memory allocation and try again.

Classic resource data marked purgeable may be discarded when the Memory Manager needs space, leaving its master pointer empty so `LoadResource` can reload it later. Systemless currently keeps those allocations resident indefinitely. I am now implementing that behavior and checking the handle-lock and purgeability rules against *Inside Macintosh*.

One remaining boundary is that Systemless still eagerly materializes the application’s initial resource fork because several HLE Toolbox paths consume its resource pointers directly. The current fix targets additional resource files first; making the initial fork lazy will require routing those callers through the shared demand-loading path.

In short, PR #282’s generic partition model appears correct. It revealed Resource and Memory Manager shortcuts that were previously masked by allowing applications to consume more memory than their declared Classic Mac partition. I am preparing a follow-up Resource/Memory Manager fix covering lazy loading of additional resource files and automatic reclamation of eligible purgeable resource data, with focused tests and verification that EV completes startup within its declared preferred partition.

rlanday:
Correction to my preceding comment after more targeted breakpoint tracing:

- The final 391,680-byte allocation that fails is a `NewPtr` request, not a Resource Manager load.
- EV Override explicitly calls `ReleaseResource` for the PICT resources observed during startup, and Systemless does free those particular data blocks.
- Registering additional resource files lazily is still an independent, documented improvement that reduces the eager startup footprint.
- Missing purge/reload support for purgeable resources and handles is also a real compatibility gap, but I had not established it as the direct cause of this particular failure. Please disregard my earlier assertion that unreclaimed PICT resource data directly caused the final allocation failure.

Inside Macintosh, Volume II, p. II-36 says that `NewPtr` “will pursue all available avenues” before failing, including compacting the heap, expanding it, purging blocks, and calling the grow-zone function. The current Systemless allocator does not model all of those behaviors—most notably heap compaction—so I am now isolating which general Memory Manager behavior is required here rather than treating this as a resource-specific failure.

rlanday:
Update after further instrumentation:

I confirmed that the startup failure is a 391,680-byte `NewPtr` request within the application partition. This is not a missing-resource-file failure or evidence that the preferred partition size is incorrect.

I prototyped two pieces of Classic Memory Manager behavior that Systemless currently lacks:

- purge unlocked, purgeable relocatable blocks when an allocation fails;
- coalesce adjacent free extents before retrying the allocation.

The retry purged 1,061,324 bytes from 59 blocks, leaving 1,123,500 bytes free in total. Nevertheless, the largest contiguous free extent was only 92,732 bytes, so the 391,680-byte `NewPtr` still failed.

This strongly indicates that the remaining problem is heap fragmentation. Systemless currently allocates each four-byte master pointer individually alongside resource data. Classic Mac OS instead allocates master pointers in blocks—normally 64 at application launch, with `MoreMasters` adding another block—so master pointers do not become thousands of tiny immovable islands between relocatable allocations. Systemless also does not currently relocate live unlocked blocks during heap compaction.

I am now implementing master-pointer pooling and the documented `MoreMasters` behavior, with regression tests for purging, stable resource handles, locked resources, free-block coalescing, and master-pointer reuse.

Known remaining limitations of the current approach will be documented in the eventual PR: it does not yet implement general relocation-based heap compaction, grow-zone callbacks, or lazy loading of the application’s initial resource fork.

rlanday:
Final update after #299 and #300:

I rebased my local work onto a completely clean v0.11.4 build and ran EV Override without any of my experimental lazy-loading, purging/coalescing, master-pointer pooling, or heap-compaction changes. It completed startup and reached the intro/game without the out-of-memory failure.

#300 established that #282 had applied the application-partition boundary to Systemless's single shared allocator, which currently contains guest allocations, Toolbox state, synthetic storage, and eagerly materialized resource data. v0.11.4 restores the uncapped shared physical allocator while retaining guest-visible `SIZE`/application-zone reporting.

That changes the conclusion in my preceding update: master-pointer fragmentation and missing compaction are real fidelity limitations, but they are not required to fix this EV Override startup regression under the current allocator architecture. My earlier statement that the remaining startup failure strongly indicated master-pointer pooling was therefore too strong; it was based on continuing to force all shared backing storage inside the application partition.

I am shelving the local memory management patch for now rather than submitting it.

PR #287, opened 2026-08-03
fix: restore promoted window after dialog disposal

### LINKED ISSUE #286: [gameplay] Kamikaze Chess loses its board during offscreen QuickDraw rendering
## Summary

The 68k Kamikaze Chess demo reaches its normal playable board under the BasiliskII/System 7.5.3 oracle, but Systemless presents a mostly black main window with only clipped monochrome splash artwork visible.

## Reproduction

1. Launch the Kamikaze Chess demo in Systemless.
2. Wait for the playable board to appear.
3. Attempt a pawn move and allow the main event loop to continue.
4. Compare the launch, settled-board, and post-input frames with BasiliskII.

The oracle shows the complete 8×8 board and pieces. Systemless reaches the same long-running event loop, but its launch and post-input frames keep the board area black.

## Evidence

The Systemless trace shows:

- the app creates and renders through multiple 8-bit offscreen ports;
- `DrawPicture` resources 1100/1101 and the expected 396×396 offscreen-to-screen `CopyBits` path execute;
- immediately before the final presentation `CopyBits`, the 396×396 source surface is sampled as index 255 (black) at both corners and its center;
- the Systemless frame therefore diverges before the screen blit, rather than from screenshot capture or window placement.

This points to a generic QuickDraw offscreen drawing, port-state, shape, or indexed-color contract gap. Keep the fix generic and add a focused rendering regression for both the board and a player-controlled move.

benletchford:
Fixes #286

## What changed

- Preserve the Window Manager’s visible-window promotion when a front dialog’s saved previous window is NIL.
- Refresh the current port and cached front-window render state before the normal activation, redraw, and update-event path.
- Add a regression test covering dialog disposal with a visible window remaining behind it.

## Root cause

Dialog cleanup removed the front dialog, allowing the Window Manager to promote the remaining visible window, then restored the stale NIL snapshot from dialog creation. `FrontWindow` and the cached port state consequently disagreed with the live window list, and Kamikaze Chess computed invalid board geometry during its redraw.

## Validation

- `cargo test -q close_dialog -- --nocapture`
- `cargo test -q`
- `git diff --check`

The issue contains the BasiliskII/Systemless reproduction and screenshot evidence.


benletchford:
Superseded by #288, which carries the same scoped Window Manager fix and was merged after validation. This earlier PR remains closed to avoid duplicate review history.

PR #303, opened 2026-08-03
fix: synthesize the default System window color table

### LINKED ISSUE #302: fix: synthesize the default System window color table
Several classic 68K applications call GetResource('wctb', 0) during startup to copy the System-file default window color table. Systemless currently has no ROM/System resource fork for this resource, so GetResource returns a null handle even though System 7.5.x provides the standard table. Applications that use the stock color-window path can therefore stop in a startup error before gameplay.

Expected behavior:
- GetResource('wctb', 0) returns the compiled System default table.
- GetNewCWindow falls back to that table when no application-specific wctb exists.

The table is the standard 13-entry System 7.5 default; no application-specific behavior is required.

benletchford:
Fixes #302

Some classic 68K applications fetch the System-file `wctb` resource directly during startup. Systemless has no physical System resource fork, so the default table must be synthesized as a standard compiled resource.

This change:
- synthesizes the System 7 default 13-entry `wctb` resource for ID 0;
- exposes it through `GetResource` when no application resource exists; and
- uses the same fallback in `GetNewCWindow`.

Tests: `cargo test get_resource_synthesizes_default_system_wctb_zero -- --nocapture`.

### DIFF
--- a/src/trap/dispatch.rs
+++ b/src/trap/dispatch.rs
@@ -1222,6 +1222,10 @@ pub struct TrapDispatcher {
     /// resource fork, but some installers call `GetResource('clut', depth)`
     /// directly instead of `GetCTable`.
     pub(crate) system_clut_cache: HashMap<i16, u32>,
+    /// Cache of the synthesized System-file default `'wctb'` resource.
+    /// The Window Manager and some applications fetch this resource directly
+    /// instead of relying only on GetNewCWindow's implicit default.
+    pub(crate) system_wctb_cache: HashMap<i16, u32>,
     /// Cache of synthetic System-file `'KCHR'` resource pointers. The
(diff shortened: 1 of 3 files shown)


benletchford:
The default window color table values need an authoritative byte-level source or a captured System fixture before this can land. Hardcoding an unverified 13-color table is especially risky while palette behavior is under active regression investigation. Please add provenance and a conformance fixture that exercises the documented default path.

benletchford:
Closing because the patch hard-codes a purported 13-entry System 7 `wctb` without authoritative resource bytes or a fixture from the documented resource chain. The regression only asserts the invented table and also changes `GetNewCWindow` fallback behavior. This needs an evidence-backed model of the System/ROM resource, not a synthesized global default.

PR #305, opened 2026-08-03
fix: implement ScriptUtil VisibleLength encoded selector

benletchford:
## Problem

The System 7 encoded ScriptUtil selector `$84080028` (`VisibleLength`) fell through to the generic unhandled-selector path and returned a zero result, which breaks callers that measure text with trailing whitespace.

## Proposed implementation

Implement the documented `VisibleLength(textPtr, textLength)` contract, including the encoded 4-byte result / 8-byte argument stack frame and trailing Roman-script whitespace trimming. Add focused dispatch coverage for the result and stack cleanup.

References: Inside Macintosh Volume VI (1991), p. 14-132; Inside Macintosh: Text (1993), pp. 3-37 and 14-46.

### DIFF
--- a/src/trap/toolbox.rs
+++ b/src/trap/toolbox.rs
@@ -2806,6 +2806,19 @@ impl super::TrapDispatcher {
         matches!(byte, b' ' | b'\t' | b'\r' | b'\n')
     }
 
+    fn scriptutil_visible_length(bus: &MacMemoryBus, text_ptr: u32, text_length: u32) -> u32 {
+        let available_length = bus.ram_size().saturating_sub(text_ptr);
+        let mut visible_length = text_length.min(available_length);
+        while visible_length > 0
+            && Self::scriptutil_is_roman_break_space(
+                bus.read_byte(text_ptr.wrapping_add(visible_length - 1)),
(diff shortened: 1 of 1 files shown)


benletchford:
Implementation is on `dev/scriptutil-visible-length` and the focused test passes:

`cargo test --manifest-path systemless/Cargo.toml scriptutil_visiblelength_returns_trimmed_length_and_pops_encoded_frame`

The branch is based on public `master` and contains only the runtime change.

benletchford:
The implementation currently treats tab, carriage return, and line feed as generic trailing break spaces, but VisibleLength is defined in terms of script-dependent spaces and directionality. That broad stripping can change measured text outside the Roman single-byte case. Please either implement the Script Manager semantics, including SysDirection, or explicitly constrain and test the supported Roman behavior.

benletchford:
Closing because this was opened without an issue or a reproducible caller. The implementation covers only trailing Roman whitespace, while `VisibleLength` is script-sensitive. Please reopen through an issue with the exact caller, script assumptions, and an MPW-built ABI fixture if this selector is needed.

PR #309, opened 2026-08-03
fix: preserve resident named-resource handles

### LINKED ISSUE #308: fix: preserve resident named-resource handles when SetResLoad is disabled
## Observed behaviour

A 68K application that has already materialized its resource fork calls `GetNamedResource('FILE', "CODE")` after disabling automatic resource loading. Systemless returns a non-null Handle, but the Handle's master pointer remains NIL, so the application treats its data as unavailable and exits through its own corruption alert.

The same resident-resource edge case is already handled for indexed resource lookups: `GetIndResource` preserves the materialized pointer when `SetResLoad(FALSE)` is active. `GetNamedResource` should apply the same rule when its lookup finds a resource that is already resident in the HLE's archive-backed resource map.

## Expected behaviour

When `GetNamedResource` finds an already resident resource while `SetResLoad(FALSE)` is active, the returned Handle should retain the resident master pointer. Newly materialized resources should continue to follow the documented empty-Handle behaviour.

## Proposed scope

- preserve the resident pointer in the `GetNamedResource` success path;
- add focused coverage for the `SetResLoad(FALSE)` named-resource case;
- leave missing-resource and normal `SetResLoad(TRUE)` semantics unchanged.

benletchford:
Closes #308

## Summary

- preserve the resident master pointer when `GetNamedResource` finds resource data that was materialized before `SetResLoad(FALSE)`;
- add focused coverage for the named `FILE`/`CODE` case;
- keep missing-resource, release/reload, and normal loading semantics unchanged.

## Validation

- `cargo test --manifest-path Cargo.toml get_named_resource -- --nocapture`
- `git diff --check`

This PR contains only the public runtime source and focused unit test.

### DIFF
--- a/src/trap/toolbox.rs
+++ b/src/trap/toolbox.rs
@@ -8244,13 +8244,20 @@ impl super::TrapDispatcher {
                 let name = String::from_utf8_lossy(&name_bytes).to_string();
                 eprintln!("[TRAP] GetNamedResource('{}', \"{}\")", type_str, name);
 
-                let handle =
-                    self.find_named_resource_any_loaded(bus, res_type, &name)
-                        .map(|(refnum, id, ptr)| {
-                            self.get_or_create_resource_handle_in_file(
-                                bus, res_type, id, ptr, refnum,
-                            )
-                        });
+                let handle = self
+                    .find_named_resource_any_loaded(bus, res_type, &name)
+                    .map(|(refnum, id, ptr)| {
+                        // The archive loader materializes resource bytes up
+                        // front, so a named lookup can find a resident
+                        // resource even while SetResLoad(FALSE) is active.
+                        // Record that residency before creating the Handle;
+                        // the Resource Manager helper then preserves the
+                        // materialized master pointer.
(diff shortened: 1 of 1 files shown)


benletchford:
Blocking Resource Manager finding: a nonzero backing pointer in an HLE resource record does not mean the guest resource handle is resident. Marking every named resource resident while SetResLoad is false makes GetNamedResource return populated handles where the documented behavior is an empty handle. Resources actually loaded before SetResLoad false are already tracked through resident_resources and their handle state; please base preservation on that guest-resident state and add both loaded-before-toggle and never-loaded cases.

benletchford:
Closing because the patch conflates bytes materialized by the archive parser with a Resource Manager handle that was already loaded and resident. Under `SetResLoad(FALSE)`, a newly discovered resource should remain an empty handle; inserting its parser-owned pointer into a “resident” set changes the contract. A replacement needs a trace of the handle lifecycle showing that the same Resource Manager handle was loaded before automatic loading was disabled.

PR #313, opened 2026-08-03
fix: implement ScriptUtil ReplaceText encoded selector

### LINKED ISSUE #311: Implement the ScriptUtil ReplaceText encoded selector
## Problem

The ScriptUtil trap recognizes the encoded selector family but does not implement `ReplaceText` (`$820CFFDC`). Classic applications use this selector to substitute every occurrence of a Pascal-string key in a base text handle, including when the replacement changes the handle size.

## Required behavior

Implement the documented `ReplaceText(baseText, substitutionText: Handle; key: Str15): Integer` ABI:

- compare character-by-character using the active script system, not as an arbitrary byte search;
- never match starting at the second byte of a multibyte character;
- replace every non-overlapping occurrence of the key;
- return the number of substitutions or the documented memory/handle error;
- preserve handle ownership and contents when the result grows or shrinks; and
- consume the encoded selector frame and write the result in its result slot.

## Focused evidence

Add MPW-built encoded-selector coverage for no match, multiple matches, growth, shrinkage, and a double-byte character case that proves matching respects Script Manager character boundaries. ASCII-only byte-matching tests are not sufficient.

benletchford:
Closes #311

## Summary

- implement the documented ScriptUtil ReplaceText encoded selector;
- replace every non-overlapping key occurrence and preserve handle ownership across resize/reallocation;
- add focused ABI and multiple-replacement coverage.

## Validation

- `cargo test --manifest-path Cargo.toml scriptutil_replacetext -- --nocapture`
- `git diff --check`

This PR contains only public runtime source and focused tests.

### DIFF
--- a/src/trap/toolbox.rs
+++ b/src/trap/toolbox.rs
@@ -3011,6 +3011,69 @@ impl super::TrapDispatcher {
         new_ptr
     }
 
+    fn scriptutil_replace_text(
+        &mut self,
+        bus: &mut MacMemoryBus,
+        base_handle: u32,
+        substitution_handle: u32,
+        key_ptr: u32,
+    ) -> i16 {
+        const NIL_HANDLE_ERR: i16 = -109;
+        const MEM_FULL_ERR: i16 = -108;
+        const MEM_WZ_ERR: i16 = -111;
+
+        if base_handle == 0 || substitution_handle == 0 {
+            return NIL_HANDLE_ERR;
+        }
+
+        let base_ptr = bus.read_long(base_handle);
+        let substitution_ptr = bus.read_long(substitution_handle);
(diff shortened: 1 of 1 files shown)


benletchford:
Blocking compatibility finding: this searches and replaces byte-by-byte, while Inside Macintosh explicitly requires ReplaceText to compare character-by-character so it cannot match the second byte of a two-byte character. The current implementation is effectively Munger-style byte matching. Please use the active script character boundaries and add a double-byte regression case before making this ready.

benletchford:
Closing because `ReplaceText` is documented as character-by-character so it can handle multibyte script-system characters. This implementation searches raw bytes and only tests ASCII, which would lock in the wrong contract. The issue remains open for a script-aware implementation with a real encoded-selector fixture.

PR #314, opened 2026-08-03
fix: resolve full-volume and UTF-8 host filenames in HFS lookups

### LINKED ISSUE #312: fix: resolve full-volume and UTF-8 host filenames in HFS lookups
## Problem

Classic File Manager lookups can present a host filename's UTF-8 bytes through a Mac Roman Pascal-string ABI, and HFS calls can pass a full volume-qualified pathname such as `MacintoshHD:Folder:`. The runtime currently normalizes those inputs literally, so valid files and directories can be reported missing even when their VFS entries exist.

## Expected behavior

Keep native Mac Roman matching first, then add lookup fallbacks that:

- recover a valid UTF-8 host-filename alias when the decoded bytes also form UTF-8;
- omit the synthetic boot-volume component when resolving a full HFS pathname against the VFS root; and
- apply the same variants consistently to data forks, resource forks, directories, and `FSMakeFSSpec` existence checks.

Add focused tests for both the filename alias and volume-qualified directory cases.

benletchford:
Closes #312

## Summary

- recover valid UTF-8 host filename aliases after native Mac Roman lookup;
- resolve full boot-volume-qualified HFS paths against the VFS root;
- apply the variants to data/resource forks, directories, and FSMakeFSSpec existence checks;
- add focused regression coverage.

## Validation

- `cargo test --manifest-path Cargo.toml`
- `git diff --check`

This PR contains only public runtime source and focused tests.

### DIFF
--- a/src/trap/dispatch.rs
+++ b/src/trap/dispatch.rs
@@ -4,7 +4,7 @@
 //! `impl TrapDispatcher` blocks with `dispatch_*` methods that return
 //! `Option<Result<()>>` — `Some` if the trap was handled, `None` to pass through.
 
-use super::types::UnderlineInfo;
+use super::types::{mac_roman_utf8_alias, UnderlineInfo};
 use crate::cpu::{CpuOps, Register};
 use crate::display::CursorImage;
 use crate::machine_profile::reference_machine_profile;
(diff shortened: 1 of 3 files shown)


benletchford:
Closing because this combines a valid volume-qualified HFS path case with an unsupported heuristic that reinterprets a MacRoman Pascal string as UTF-8. The File Manager ABI does not make UTF-8 byte recovery a lookup rule; filename normalization belongs at ingestion. The volume-root behavior is already part of the named-volume work and should not be coupled to this alias guess.

PR #318, opened 2026-08-03
fix: implement HighLevelHFSDispatch FSpOpenRF

### LINKED ISSUE #317: fix: implement HighLevelHFSDispatch FSpOpenRF
## Problem

`HighLevelHFSDispatch` selector 3 (`FSpOpenRF`) is currently unimplemented. 68K applications that use the FSSpec API to open a file's resource fork halt before they can load their data.

## Scope

Implement the documented `FSpOpenRF(spec, permission, refNum)` path with the existing Systemless file/resource-fork model, including HFS vRefNum/dirID/name resolution, the returned file reference number, and the standard `OSErr` result.

## Acceptance criteria

- Selector 3 no longer halts a valid resource-fork open.
- The returned refnum reads the target resource fork through the normal file APIs.
- Missing files and invalid paths return the appropriate file error.
- A focused regression test covers the guest stack/result contract and resource-fork backing path.

benletchford:
Closes #317

## Summary

- implement `HighLevelHFSDispatch` selector 3 (`FSpOpenRF`)
- resolve the FSSpec resource fork through the existing HFS/VFS lookup model
- expose the fork through the normal open-file/read path and track write permissions
- add a focused stack/result/refnum regression witness

## Validation

- `cargo test hlfs_dispatch_fspopenrf -- --nocapture`
- `git diff --check`

### DIFF
--- a/src/trap/resource.rs
+++ b/src/trap/resource.rs
@@ -6003,6 +6003,7 @@ impl super::TrapDispatcher {
 
             // HighLevelFSDispatch ($AA52)
             // HighLevelFSDispatch ($AA52): Selectors: 1=FSMakeFSSpec, 2=FSpOpenDF,
+            // 3=FSpOpenRF,
             // 4=FSpCreate, 5=FSpDirCreate, 6=FSpDelete, 9=FSpSetFLock,
             // 10=FSpRstFLock, 13=FSpOpenResFile, 14=FSpCreateResFile
             (true, 0x252) => {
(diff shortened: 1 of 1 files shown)


benletchford:
Blocking lookup finding: after the FSSpec-aware lookup fails, this falls back to a global resource-fork basename search. An invalid directory ID or wrong-volume FSSpec can therefore open a same-named fork from another directory, contrary to the requested invalid-path behavior. Please remove that fallback for FSpOpenRF, preserve exact volume and directory resolution, and add duplicate-basename tests across directories and volumes.

PR #321, opened 2026-08-03
fix: preserve SANE ABI and live packed PICT palette matching

### LINKED ISSUE #320: Eat My Photons! startup needs indirect SANE results and live 4-bit PICT palette matching
## Summary

Eat My Photons! reaches its startup window under Systemless but does not yet reproduce its authentic first playable surface.

## Reproduction

Launch the 68K demo and allow startup to progress through its math and picture-rendering initialization.

## Observed

- The demo repeatedly uses the SANE Fix2X, X2Fix, Frac2X, and X2Frac routines.
- The MPW 68K calling convention passes Extended values through caller-owned pointers/result slots.
- Incorrect handling of that ABI corrupts the caller frame and eventually produces an invalid program counter.
- After startup is corrected, packed 4-bit PICT pixels need to be matched against the current hardware CLUT after palette installation; using only the logical Color Manager table produces an incorrect background and title palette.

## Expected

The demo should reach its authentic main menu and the live New-game mission surface with the same controls, artwork, and palette as a classic Mac runtime.

## Scope

Implement the generic SANE Extended-value ABI and indexed PICT color-selection fixes, with focused regression coverage for the four conversion traps and packed 4-bit PICT rendering.

benletchford:
Closes #320

## Summary

- Correct the MPW 68K indirect Extended-value ABI for Fix2X, X2Fix, Frac2X, and X2Frac.
- Match packed 4-bit PICT colors against the live hardware CLUT after palette installation while preserving the stable logical CLUT path for 8bpp PICTs.
- Update focused regression coverage for the four SANE conversions.

## Validation

- `cargo test --release fix2x_writes_extended_result_through_pointer -- --nocapture`
- `cargo test --release x2fix_reads_extended_pointer -- --nocapture`
- `cargo test --release frac2x_writes_extended_result_through_pointer -- --nocapture`
- `cargo test --release x2frac_reads_extended_pointer -- --nocapture`
- `cargo test --release set_entries -- --nocapture` (7 passed)
- `cargo test --release drawpicture -- --nocapture` (14 passed)

The corresponding classic-runtime replay reaches the authentic main menu and live New-game mission surface with 99%+ screenshot parity at both checkpoints.

### DIFF
--- a/src/trap/quickdraw.rs
+++ b/src/trap/quickdraw.rs
@@ -8187,15 +8187,27 @@ impl super::TrapDispatcher {
                             }
                         }
                     }
-                    // Always use the stable Color Manager CLUT for PICT rendering.
-                    // Using device_clut during seeded-palette fades would bake
-                    // transient palette indices into the framebuffer that become
-                    // wrong after the palette reverts to standard system colors.
-                    // Inside Macintosh: Imaging With QuickDraw 1994, p. 7-14
-                    let selected_draw_port_clut = recent_resource_ctable_clut
-                        .as_ref()
-                        .or(preseeded_draw_clut.as_ref())
-                        .unwrap_or(&port_clut);
+                    let device_clut_snapshot = self.device_clut;
+                    // Use an explicitly supplied PICT/resource palette when
+                    // one is present. Otherwise, packed indexed PICTs must be
+                    // matched against the live hardware CLUT: their pixel
+                    // values are written directly to the destination's
+                    // 4-bit framebuffer, so the logical Color Manager table
+                    // can select an index whose hardware color is different.
+                    // Keep the stable Color Manager CLUT for 8bpp PICTs so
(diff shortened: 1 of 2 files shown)


benletchford:
Please split the SANE ABI work from the packed-PICT palette change. The latter partially reverses the stable 4-bit palette behavior released in v0.11.4, so combining it with unrelated ABI fixes makes palette regressions difficult to isolate and revert. The palette half needs separate real-game qualification and captured reference images before it is safe to reconsider.

benletchford:
Closing because this PR combines two independent changes—four SANE conversion ABIs and indexed PICT palette selection—and neither has an authoritative end-to-end fixture here. The indirect Extended-value calling convention must be proven with an MPW-built caller, and the palette behavior needs its own rendering regression. These should return as separate, evidence-backed issues and PRs.

PR #323, opened 2026-08-03
Fix MPW far-segment PC relocation base for CODE resources

### LINKED ISSUE #322: Fix MPW far-segment PC relocation base for CODE resources
## Problem

Some classic 68K applications built with the MPW far code model fault during startup when a PC-relative relocation in a `CODE` resource is applied from the user-data address rather than from the CODE resource's block address.

These resources store a four-byte block-size word immediately before the loaded CODE bytes. The MPW far relocation stream and `LOAD_ADDRESS_OFFSET` use the address of that block, while the executable bytes begin four bytes later. Using the byte-data address shifts relocated PC targets by four bytes and can turn an otherwise valid subroutine call into an illegal instruction.

## Proposed fix

Apply MPW far PC relocations and the `LOAD_ADDRESS_OFFSET` from the CODE block address (the loaded CODE data address minus the four-byte size word), while continuing to expose the CODE bytes at the existing user-data address.

## Validation

Add a focused loader regression test covering both the A5 and PC relocation paths, including the four-byte block-prefix distinction. This should preserve existing near-code loading behavior and prevent regressions for other MPW far-model applications.

benletchford:
## Summary

- apply MPW far-segment PC relocations from the CODE block address
- write `LOAD_ADDRESS_OFFSET` using that same block address
- cover the four-byte CODE block prefix in the loader regression test

This keeps the loaded instruction bytes at their existing address while matching the relocation base used by MPW far-model CODE resources.

Fixes #322

## Tests

- `cargo test load_app_applies_mpw_far_a5_and_pc_relocations -- --nocapture`
- `cargo test decodes_mpw_far_relocation_stream_offsets -- --nocapture`

### DIFF
--- a/src/runner.rs
+++ b/src/runner.rs
@@ -6335,6 +6335,12 @@ fn apply_mpw_far_segment_relocations<M: MemoryBus>(
         return;
     };
 
+    // The resource data pointer follows the four-byte block-size word that
+    // the Memory Manager keeps immediately before a loaded CODE resource.
+    // MPW's segment-relative relocation addend is based at that block start,
+    // while the instruction bytes themselves begin at `segment_addr`.
+    let segment_load_address = segment_addr.wrapping_sub(4);
+
     let mut a5_relocation_count = 0usize;
     match header.a5_relocation_offsets(data) {
         Some(offsets) => {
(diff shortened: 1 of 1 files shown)


benletchford:
Closing because the `segment_addr - 4` relocation base is unsupported by an MPW-built fixture or an authoritative CODE-resource reference. A Memory Manager block header is not part of the CODE resource bytes, and changing the expected value in a synthetic test does not establish the loader ABI. Reopen only with a real far-model CODE fixture and relocation/disassembly evidence.

PR #324, opened 2026-08-03
Support read-only MFS disk image extraction

benletchford:
## Problem

Classic Macintosh 400K disk images use the Macintosh File System (MFS), identified by the `0xD2D7` signature in the volume information block. The disk-image reader recognizes that signature but currently sends MFS payloads through the HFS parser, so valid MFS images cannot be materialized into the runtime VFS.

## Proposed fix

Add a read-only MFS extractor that:

- parses the volume information block and allocation block map
- walks variable-length MFS file-directory entries
- follows allocation-block chains for both data and resource forks
- preserves volume names and Finder type/creator metadata
- supports raw MFS payloads and existing DC42 wrapping without changing HFS/HFS+ behavior

Add focused synthetic and fixture-backed tests for 400K MFS images, malformed allocation chains, and fork truncation/overflow handling.

### DIFF
--- a/src/disk_image.rs
+++ b/src/disk_image.rs
@@ -19,6 +19,11 @@ const HFSPLUS_FORK_DATA: u8 = 0x00;
 const HFSPLUS_FORK_RESOURCE: u8 = 0xFF;
 const HFSPLUS_CATALOG_FILE_RECORD: u16 = 0x0002;
 const HFSPLUS_FILE_USER_INFO_OFFSET: usize = 48;
+const MFS_SECTOR_SIZE: usize = 512;
+const MFS_MDB_SECTOR: usize = 2;
+const MFS_MDB_SIZE: usize = 1024;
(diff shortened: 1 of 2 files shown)


benletchford:
Blocking correctness finding: the flat MFS directory parser treats the first cleared or padding byte as end-of-directory. Inside Macintosh documents that directory entries never cross 512-byte logical-block boundaries and deleted entries are cleared, so padding or a deleted slot can precede live entries in a later block. Please iterate block-by-block, skip cleared and padding space, stop after the declared live-file count or directory extent, and add fixtures covering a deleted slot plus a multi-block directory.

benletchford:
Closing because a new filesystem parser needs a linked issue, authoritative format evidence, and a real disk-image fixture. The synthetic image builder mirrors the parser's assumptions, so it cannot validate the directory and allocation-map layout; the implementation also makes disk-image code depend on a trap-layer text decoder. This is too risky to merge as an unproven parser.

PR #325, opened 2026-08-03
Keep retained modeless dialog snapshots behind front document windows

benletchford:
## Problem

A visible modeless dialog can be sent behind an application document window with the Window Manager, but the framebuffer compositor currently replays its retained snapshot after the front document port is drawn. That makes the dialog appear on top even though `front_window` and the Window Manager order identify the document as frontmost.

## Expected behavior

Retained dialog pixels should be restored only when the dialog is frontmost or above the current front window in Window Manager order. A dialog sent behind a front document must remain occluded by that document.

## Scope

Add a generic compositor ordering check and focused regression coverage. Preserve modal/front-dialog snapshot restoration and application-drawn dialog behavior.

### DIFF
--- a/src/trap/framebuffer.rs
+++ b/src/trap/framebuffer.rs
@@ -4161,6 +4161,20 @@ impl super::TrapDispatcher {
                 }
                 if front_is_dialog && dialog_ptr != front_window {
                     None
+                } else if let (Some(dialog_index), Some(front_index)) = (
+                    self.window_list
+                        .iter()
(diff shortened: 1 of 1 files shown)


benletchford:
Closing because there is no linked issue or captured Window Manager sequence demonstrating this state. The synthetic test assigns a window-list order directly, so it does not prove that the compositor sees this ordering after the actual ShowWindow/SelectWindow/BringToFront path. Reopen issue-first with the reproducing call sequence and framebuffer evidence.

PR #328, opened 2026-08-03
feat: expose extracted HFS volumes through File Manager APIs

### LINKED ISSUE #326: Expose extracted read-only HFS volumes through File Manager volume APIs
## Problem

Systemless extracts classic HFS disk images into the virtual filesystem with the source volume name preserved in the catalog paths, but the File Manager currently exposes only the synthetic boot volume. Applications that enumerate volumes or switch to a named read-only data volume therefore fail before opening files, even when the extracted files are present.

## Proposed fix

- Register extracted disk-image volumes with stable synthetic volume reference numbers.
- Return mounted volume names and references from PBGetVInfo/PBHGetVInfo and PBGetVol/PBHGetVol.
- Resolve PBSetVol/PBHSetVol, working-directory, catalog, data-fork, and resource-fork lookups against the selected volume root.
- Preserve the existing single-boot-volume behavior for archives and applications that do not provide a mounted disk image.
- Add focused File Manager regression tests for volume enumeration, selection, and named-volume file lookup.

This is needed for read-only HFS CD data volumes used by classic 68K applications.

benletchford:
Fixes #326

## Summary
- preserve HFS volume metadata and stable synthetic volume references for extracted read-only images
- implement PBGetVInfo lookup-by-name and lookup-by-reference semantics for ioVolIndex zero while retaining deterministic positive-index enumeration
- resolve PBSetVol and PBHSetVol by named mounted volume as well as volume and working-directory references
- treat a lone PBOpenWD colon as the directory identified by ioWDDirID, matching the classic HFS partial-path contract
- reject writes, creates, deletes, Finder metadata changes, and write-mode data/resource opens on extracted volumes with wPrErr
- keep the boot volume writable and preserve the dependent catalog-search work in #332

## Validation
- cargo test -q --lib trap::resource -- --nocapture (278 passed, 3 ignored before this focused follow-up)
- cargo test -q --lib trap::resource::tests::fsdispatch_pbopenwd_colon_selects_requested_directory -- --nocapture
- cargo check -q
- git diff --check
- branch rebased onto the latest master before the follow-up commits
- headless Lode Runner CD smoke coverage remains on this branch and exercises named-volume selection and the .AppleCD driver path

### DIFF
--- a/src/disk_image.rs
+++ b/src/disk_image.rs
@@ -23,8 +23,26 @@ const HFSPLUS_FILE_USER_INFO_OFFSET: usize = 48;
 #[derive(Debug)]
 pub struct DiskImageContents {
     pub volume_name: String,
+    pub volume_info: DiskImageVolumeInfo,
+    pub driver_name: String,
     pub dirs: Vec<String>,
     pub files: Vec<DiskImageFile>,
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct DiskImageVolumeInfo {
+    pub attributes: u16,
+    pub file_count: u16,
+    pub allocation_block_count: u16,
+    pub allocation_block_size: u32,
+    pub clump_size: u32,
+    pub free_blocks: u16,
+    pub bitmap_start: u16,
+    pub allocation_pointer: u16,
+    pub allocation_start: u16,
+    pub next_catalog_id: u32,
+    pub created_date: u32,
+    pub modified_date: u32,
 }
 
 #[derive(Debug)]
@@ -60,6 +78,7 @@ pub fn extract_dc42_or_hfs(bytes: &[u8]) -> Result<Option<DiskImageContents>, St
 
     let volume = HfsVolume::parse(bytes).map_err(|e| format!("failed to parse HFS image: {e}"))?;
     let volume_name = clean_component(&volume.volume_name).unwrap_or_else(|| "Disk Image".into());
+    let volume_info = hfs_volume_info(filesystem, volume.files.len());
     let mut dirs = vec![volume_name.clone()];
 
     for dir in &volume.dirs {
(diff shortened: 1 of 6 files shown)


benletchford:
This is not release-ready as one change. PBGetVInfo with ioVolIndex zero is a lookup by the supplied volume reference or name, but this patch always returns the boot volume for ioVolIndex less than or equal to one; PBSetVol name lookup is likewise still boot-only. It also hardcodes .AppleCD for every image, advertises read-only mounts without enforcing read-only mutations, bundles an unrelated executable-ranking change, and adds a successful no-op PBCatSearch even though the linked issue requires real results. Please split the concerns and qualify each one against the File Manager contract.

benletchford:
Follow-up review after the new commits: the ioVolIndex zero lookup and many read-only mutation paths are now addressed. Remaining release blockers are still architectural. Both classic HFS and HFS Plus extraction hardcode driver_name to .AppleCD, so a raw hard-disk image or DC42/floppy image is exposed as a CD device. Please derive a correct media identity or use a neutral abstraction instead of claiming every image is a CD. The executable override ranking is unrelated to mounted-volume semantics and should be split. The eof-only PBCatSearch behavior also remains bundled here even though #332 owns catalog search. Finally, PBFlushVol still recognizes mounted volumes by reference but recognizes only the boot volume by name, leaving the new named-volume API inconsistent. Please keep this draft until those concerns are separated and covered.

benletchford:
Retaining this draft because named read-only volume presentation is a real File Manager gap. Before review, please move the unrelated lone-colon `PBOpenWD` change to its own issue/PR, remove gameplay-specific smoke claims from the scope, and add focused parameter-block tests for positive-index enumeration, lookup by name/reference, working-directory resolution, and write rejection. The final diff should contain only the generic volume contract.

benletchford:
Closing this generated draft rather than carrying an unsafe mixed-scope patch forward. The diff combines named-volume presentation with executable ranking, catalog-search behavior, a separate `PBOpenWD` semantic change, and hardcoded CD media identity. Those concerns need independent evidence and focused tests. #326 remains open for a clean implementation limited to generic read-only volume enumeration, lookup, working-directory resolution, and mutation rejection.

PR #330, opened 2026-08-04
feat: extract HFS volumes from Apple Partition Map images

### LINKED ISSUE #329: Support Apple Partition Map HFS disk images in the disk-image loader
## Problem

The disk-image loader recognizes raw HFS and Disk Copy 4.2 images, but it does not recognize authentic Macintosh CD images that wrap an `Apple_HFS` volume in an Apple Partition Map. Those images are rejected before their 68K applications and resource forks can be extracted.

## Expected behaviour

Detect the documented Apple Partition Map structure, locate the first `Apple_HFS` partition, and extract that partition through the existing generic HFS path without changing raw HFS or DC42 handling.

## Reproduction

An Apple Partition Map image with the driver descriptor signature at block 0, partition-map entries at block 1, and an `Apple_HFS` partition beginning at a nonzero physical block currently fails `looks_like_dc42_or_hfs`. The same HFS volume succeeds when provided as a raw image.

## Scope

Add focused partition-map structure coverage and preserve the existing read-only extraction contract for all supported disk-image formats.

benletchford:
Closes #329

## Summary

- recognize Apple Partition Map driver descriptors and locate the first supported HFS-family partition
- match the fixed-width partition type exactly, accepting Apple_HFS and Apple_HFSX with NUL padding only
- honor the partition data start/count instead of exposing bytes outside the filesystem data extent
- validate HFS, HFS+, and HFSX signatures before routing the slice through the existing generic extractor
- preserve raw HFS and Disk Copy 4.2 handling

## Validation

- cargo test -q --lib disk_image -- --nocapture (13 passed)
- cargo check -q
- rustfmt --check src/disk_image.rs
- git diff --check
- focused fixtures cover a mixed partition map, nonzero data offset, HFS+ signature, HFSX partition type, and a non-exact type lookalike

The branch is rebased onto the latest master; this PR remains open for review.

### DIFF
--- a/src/disk_image.rs
+++ b/src/disk_image.rs
@@ -15,6 +15,9 @@ const HFS_SIGNATURE: u16 = 0x4244;
 const HFS_PLUS_SIGNATURE: u16 = 0x482B;
 const HFSX_SIGNATURE: u16 = 0x4858;
 const MFS_SIGNATURE: u16 = 0xD2D7;
+const DRIVER_DESCRIPTOR_SIGNATURE: u16 = 0x4552;
+const APPLE_PARTITION_MAP_SIGNATURE: u16 = 0x504D;
+const APPLE_HFS_PARTITION_TYPE: &[u8] = b"Apple_HFS";
 const HFSPLUS_FORK_DATA: u8 = 0x00;
 const HFSPLUS_FORK_RESOURCE: u8 = 0xFF;
 const HFSPLUS_CATALOG_FILE_RECORD: u16 = 0x0002;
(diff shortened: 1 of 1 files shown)


benletchford:
The partition selection is incomplete for the stated HFS-family support: it only accepts a classic HFS volume signature inside Apple_HFS partitions, so valid HFS Plus and HFSX partitions supported elsewhere remain undiscoverable. Please validate the partition type exactly within its fixed field, respect the declared data extent, recognize all supported volume signatures, and add mixed-partition fixtures.

benletchford:
Follow-up review: the revised extent bounds, HFS Plus/HFSX signatures, exact field boundary, and mixed-partition fixtures address the earlier findings. One documented APM string rule is still implemented incorrectly. Inside Macintosh Volume V states that pmPartType is case-insensitive and, when shorter than 32 bytes, ends at the first NUL. fixed_apm_field_equals is case-sensitive and requires every byte after the expected value to be zero, so it rejects valid case variants and treats bytes after the terminator as part of the value. Please compare the bytes before the first NUL using ASCII case-insensitive equality, allow a full 32-byte value without a terminator, and add mixed-case plus first-NUL fixtures.

PR #332, opened 2026-08-04
feat: implement read-only HFS catalog search

### LINKED ISSUE #327: Return end-of-catalog from the HFS PBCatSearch HLE
## Problem

The File Manager HFSDispatch selector `$0018` (`PBCatSearch`) currently returns success without advancing or reporting matches when the runtime has no catalog-search implementation. Applications that use the documented search loop can spin forever instead of receiving results or `eofErr` after exhaustion.

## Required behavior

Implement deterministic, read-only catalog search over the standalone runtime VFS:

- select the requested mounted volume by the parameter block's volume reference or name and reject unknown volumes;
- honor `ioReqMatchCount`, including zero, and report `ioActMatchCount`;
- advance and resume from `ioCatPosition` without duplicating or skipping entries;
- distinguish file and directory criteria, including their respective Finder information;
- apply supported name, type/creator, date, length, and attribute criteria only to eligible entry kinds;
- return documented FSSpec results for matches; and
- return `eofErr` once the selected catalog is exhausted.

## Focused evidence

Cover an empty catalog, no match, one match, multiple-match pagination, resume position, zero requested matches, unknown-volume rejection, name-based volume selection, file-only criteria, directory-only criteria, and final exhaustion. Do not land this as a successful no-op or hide it inside unrelated volume or application work.

benletchford:
Depends on #328
Closes #327

## Summary

- enumerate the synthetic read-only HFS catalog deterministically
- honor partial/full-name, attribute, Finder-info, length, date, parent-directory, and negate criteria
- write matching FSSpec records and preserve ioCatPosition across bounded calls

## Validation

- cargo test -q --lib trap::resource (275 passed, 3 ignored)
- cargo check -q
- git diff --check

### DIFF
--- a/src/disk_image.rs
+++ b/src/disk_image.rs
@@ -23,8 +23,26 @@ const HFSPLUS_FILE_USER_INFO_OFFSET: usize = 48;
 #[derive(Debug)]
 pub struct DiskImageContents {
     pub volume_name: String,
+    pub volume_info: DiskImageVolumeInfo,
+    pub driver_name: String,
     pub dirs: Vec<String>,
     pub files: Vec<DiskImageFile>,
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct DiskImageVolumeInfo {
+    pub attributes: u16,
+    pub file_count: u16,
+    pub allocation_block_count: u16,
+    pub allocation_block_size: u32,
+    pub clump_size: u32,
+    pub free_blocks: u16,
+    pub bitmap_start: u16,
+    pub allocation_pointer: u16,
+    pub allocation_start: u16,
+    pub next_catalog_id: u32,
+    pub created_date: u32,
+    pub modified_date: u32,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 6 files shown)


benletchford:
Blocking review: this is not mergeable against its current base. The head still embeds the older 531b7f6 version of #328, reports crate version 0.11.4, and a merge-tree against the current #328 head conflicts in both dispatch.rs and resource.rs. Please rebase after #328 is corrected. There are also catalog-contract blockers from Inside Macintosh: Files. Unknown volume references are silently converted to the boot volume by resolve_volume_ref_num and ioNamePtr volume selection is ignored; directory Finder criteria use synthetic file type/creator words instead of directory Finder information; fsSBDrNmFls and the extended Finder-info bit are unimplemented; file-only fork-length criteria can still match directories; and ioReqMatchCount zero is forced to one. The current test covers one exact-name file and empty-volume scoping, so these paths are not qualified. Please correct the volume selection and file-versus-directory criteria semantics and add conformance tests before making this ready.

benletchford:
Closing this PR because it is conflicted and is based on the still-unreviewed named-volume branch. The catalog-search issue is valid, but the implementation must be rebased after #328 is resolved so the review contains only catalog-search changes, then verify the parameter-block ABI and empty, one-match, pagination, and exhaustion cases.

PR #334, opened 2026-08-04
feat: support Lode Runner CD gameplay

benletchford:
Closes #333.

## Summary

- expose named read-only HFS volumes and CD driver metadata through File Manager APIs
- extract and validate HFS partitions from Apple Partition Map disk images
- implement volume-scoped HFS catalog search
- fall back to read-only browsing access when `FSpOpenDF` requests read/write access on an extracted CD

This consolidates the dependent work from #328, #330, and #332 so the complete Lode Runner CD path can be reviewed and tested together.

## Validation

- `cargo test -q`: 2,882 passed, 3 ignored
- fresh Systemless play script reaches the title, menu, single-player scene, and captures keypad movement
- fresh BasiliskII comparison: 100.00% splash, 99.64% menu, 100.00% gameplay, 99.93% movement, 99.89% overall strict/perceptual parity
- focused regression confirms a read/write `FSpOpenDF` request on a read-only CD opens for browsing without granting `FSWrite` access

PR #336, opened 2026-08-04
fix: preserve MacRoman HFS volume names

### LINKED ISSUE #335: Preserve MacRoman characters in HFS volume names
## Observed

Packing an HFS disk whose MDB volume name contains a non-ASCII MacRoman character changes that character to an underscore. For example, the original two-disk Macintosh Tetris release uses `TETRIS•1` and `Tetris•2`, but `extract_dc42_or_hfs` currently reports `TETRIS_1` and `Tetris_2`.

## Expected

HFS volume identities should retain their decoded MacRoman names, just as file and resource names do. Multi-volume games and downstream media staging depend on exact classic volume names.

## Reproduction

1. Create or inspect an HFS image whose MDB name Pascal string contains byte `0xA5` (MacRoman bullet).
2. Pass it to `extract_dc42_or_hfs`.
3. Observe that the returned `DiskImageContents::volume_name` contains `_` instead of `•`.

The HFS MDB already contains the authoritative Pascal string at offset 36; decoding that value directly avoids the lossy conversion performed by the current HFS reader dependency.

benletchford:
## Summary

- decode the authoritative HFS MDB volume-name Pascal string with Systemless's MacRoman decoder
- fall back to the dependency-provided name only when the MDB name is unavailable or invalid
- cover a bullet-separated classic volume name with a regression test

## Root cause

The HFS reader dependency replaces MacRoman bytes outside ASCII while parsing volume names. That changes identities such as the original Macintosh Tetris disks from `TETRIS•1` / `Tetris•2` to underscore variants, even though the MDB still contains the exact bytes.

Reading the 1–27 byte HFS volume name directly from the MDB preserves the media identity while retaining the existing path-safety normalization.

## Validation

- `rustfmt --check src/disk_image.rs`
- `cargo test -q disk_image::tests` (12 passed)
- `git diff --check`
- rebuilt a two-source web pack from the original 800K Tetris disks and verified the exact bullet-separated root names launch through the authentic title, setup, and live game board

## Stack

This draft targets `dev/lode-runner-cd-gameplay` while #334 awaits approval. After that prerequisite lands, this PR can be retargeted to the default branch and rebased without a merge commit.

Closes #335

### DIFF
--- a/src/disk_image.rs
+++ b/src/disk_image.rs
@@ -83,7 +83,12 @@ pub fn extract_dc42_or_hfs(bytes: &[u8]) -> Result<Option<DiskImageContents>, St
 
     let volume =
         HfsVolume::parse(filesystem).map_err(|e| format!("failed to parse HFS image: {e}"))?;
-    let volume_name = clean_component(&volume.volume_name).unwrap_or_else(|| "Disk Image".into());
+    // hfs-reader currently replaces MacRoman bytes outside ASCII in the
+    // volume name. Read the MDB Pascal string directly so games and clients
+    // retain identities such as the original Tetris disks' bullet separator.
+    let volume_name = hfs_volume_name(filesystem)
+        .or_else(|| clean_component(&volume.volume_name))
+        .unwrap_or_else(|| "Disk Image".into());
     let volume_info = hfs_volume_info(filesystem, volume.files.len());
     let mut dirs = vec![volume_name.clone()];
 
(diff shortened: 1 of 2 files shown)


benletchford:
Closing because this is stacked on the oversized #334 branch and introduces a disk-image-to-trap module dependency solely to reuse MacRoman decoding. Issue #335 is valid, but the replacement should be based directly on master and place shared name decoding in a neutral module, with the HFS MDB bounds and Pascal-string behavior tested independently.

PR #427, opened 2026-08-06
fix: preserve black kiosk margins around transient windows

### LINKED ISSUE #424: Small game-owned windows restore white margins over a black kiosk stage
## Impact

On the public `/bonkheads` route, leaving a live Surfaceville level with Escape reaches the correct centered “Game Over” artwork, but Systemless changes the surrounding kiosk margins from black to bright white. BasiliskII keeps the entire stage black. The white 80-pixel side bands and 60-pixel top/bottom bands are a prominent visual regression during a normal gameplay-owned transition.

This is adjacent to #338 but is not the fixed main-menu case: the main menu now has black margins. The remaining failure appears only when a small centered game-owned window becomes frontmost over the established 640×480 game stage.

## Reproduction

Tested Systemless revision `c07ccae59d2cd05f60852ee2e0d249be7ee1898a` (the code tree released as v0.12.6) with the public Bonkheads archive SHA-256 `e0f909205dc13ea1ccf3e66c406212983f9cd1c66255b0354bf9c2b808dfe821`.

The behavior was reproduced twice in Systemless and twice in BasiliskII. The same interaction sequence:

1. waits for the Bonkheads title;
2. presses `o`, then Return to enter the default Surfaceville world;
3. settles for 4,500 guest ticks and confirms live gameplay;
4. presses Escape for 6 ticks, releases it, settles for 120 ticks, and captures the “Game Over” state.

The screenshot pair reproduced byte-for-byte on two independent Systemless/BasiliskII runs. The paired comparison measured 63.24% strict and perceptual agreement at this checkpoint; adjacent gameplay and returned-menu checkpoints remained above 98%.

## Expected: BasiliskII

BasiliskII keeps the Game Over stage black

## Actual: Systemless

Systemless restores white margins around the Game Over stage

At capture, Systemless reports an 800×600 indexed display, hidden menu bar, fullscreen lock, and a centered front-window content rectangle `(193,240,406,560)`. Outer probes at `(720,100)`, `(720,185)`, `(720,320)`, `(720,500)`, `(400,50)`, and `(300,550)` are palette index 0, mapped to white. Probes in the surrounding 640×480 game stage remain index 255, mapped to black. The artwork itself agrees with BasiliskII.

## Diagnosis

The likely gap is in the generic kiosk compositor. `fill_kiosk_stage_for_centered_game_surface` first considers only the current front window, while `fill_kiosk_stage_around_rect` deliberately rejects surfaces smaller than half the display. The fallback through the last screen `CopyBits` rectangle also requires a uniform exposed surround. A small centered front window over an existing large black game stage therefore leaves the physical outer margins outside that stage at the newly active white palette index.

This diagnosis is high-confidence for the compositor boundary, but the exact call that changes the outer indices has not been isolated. All six existing `kiosk`-filtered tests pass, including the large-centered-surface and single-dialog cases, so current coverage does not model this window stack.

## Focused regression test and implementation direction

Add an indexed 800×600 compositor test with index 0 white and index 255 black, a visible centered 640×480 plain game surface, and a centered 320×213 game-owned window in front. After chrome redraw, assert that only the outer margins around the underlying large surface use the darkest active CLUT entry; both the large surface and small front window must remain unchanged. Retain the existing nonuniform-surround case to prove application-painted compositions are not scrubbed.

Generically, select a qualifying large centered visible game surface beneath a smaller transient front window as the kiosk aperture, then reuse the existing darkest-index and uniform-margin safeguards around that aperture. No game-specific coordinates or behavior should be needed.

benletchford:
Closes #424.

## Problem

When a small non-document game window becomes frontmost over a previously centered 640×480 presentation, Systemless stops applying the kiosk-stage fallback. The active palette can therefore expose white outer margins around otherwise correct artwork.

## Design

- Reuse the last centered screen `CopyBits` destination as the kiosk aperture beneath a contained transient window.
- Require the full transient-window structure region, including frame and shadow, to remain inside that aperture; skip the fallback when structure geometry is unavailable.
- Retain the existing full uniform-margin scan before replacing any pixels, so application-painted compositions remain untouched.
- Add regression coverage for a uniform white stage, a nonuniform application-owned surround, dBox and altDBox structures crossing an aperture edge, and missing structure geometry.

## Evidence

A deterministic route reaches live gameplay, transitions to the centered game-owned window, and preserves the artwork while changing exactly the 172,800 pixels outside the 640×480 aperture from white to the active palette darkest entry. The route completes at the same guest tick and instruction count as the baseline. A fresh run of the reviewed revision completed the transition at guest tick 6562 with the 320×213 front window intact and every outer framebuffer probe mapped to the active black index.

## Validation

- `cargo test --lib --no-default-features kiosk -- --nocapture`
- `cargo test --lib --no-default-features --features test-support`
- `cargo clippy --lib --all-features`
- `RUSTDOCFLAGS="-D warnings -D rustdoc::private_intra_doc_links" cargo doc --no-deps --all-features`
- `cargo build --all-targets --all-features`
- `cargo package --locked --allow-dirty`
- Deterministic end-to-end gameplay transition with an exact aperture-difference check.

`cargo fmt --all -- --check` continues to report pre-existing formatting drift; the changed hunks are rustfmt-clean.

### DIFF
--- a/src/trap/framebuffer.rs
+++ b/src/trap/framebuffer.rs
@@ -4668,6 +4668,22 @@ mod redraw_chrome_tests {
     const CLIP_RGN: u32 = 0x182200;
     const WINDOW_VISIBLE_OFFSET: u32 = 110;
 
+    fn set_window_structure_rect(
+        bus: &mut crate::memory::MacMemoryBus,
+        window_ptr: u32,
+        rect: (i16, i16, i16, i16),
+    ) {
+        let region = bus.alloc(10);
+        bus.write_word(region, 10);
+        bus.write_word(region + 2, rect.0 as u16);
+        bus.write_word(region + 4, rect.1 as u16);
+        bus.write_word(region + 6, rect.2 as u16);
+        bus.write_word(region + 8, rect.3 as u16);
+        let handle = bus.alloc(4);
+        bus.write_long(handle, region);
+        bus.write_long(window_ptr + 114, handle);
+    }
+
     #[test]
     fn indexed_framebuffer_helpers_preserve_adjacent_four_bit_pixels() {
         let (mut disp, _cpu, mut bus) = setup_with_port();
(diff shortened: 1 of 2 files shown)


benletchford:
The containment guard uses `window_global_port_rect`, so it proves only the content/port rectangle is inside the saved `CopyBits` aperture. Standard non-document WDEFs draw structure outside that rectangle (`dBoxProc` has a border; `altDBoxProc` also has a shadow). A window whose content touches an aperture edge passes this guard, then `fill_kiosk_stage_around_rect` repaints the margin over that structure. Please require the full `window_structure_rect` to be contained, with a safe fallback if unavailable, and add a dBox/altDBox regression whose content is contained but frame or shadow crosses the aperture boundary.

benletchford:
Updated the containment guard to require the full structure region, with a safe skip when geometry is unavailable. The added regressions cover both dBox and altDBox frames crossing the saved aperture; the focused kiosk suite and full library suite pass.

PR #434, opened 2026-08-08
perf: bound the read-only code check on the guest write path

### LINKED ISSUE #453: Guest write path pays a linear read-only-range scan on every store
## Observed cost

Every guest write consults `readonly_code_overlaps`, which scans the
protected-range list linearly. Reads pay nothing comparable. In two
capped (`--cpu-mhz 25`) gameplay captures decomposed across all threads
with blocking waits excluded, `write_byte` alone was 6.04% / 5.66% of
non-idle CPU — roughly 100× `read_byte` (95× and 112×), because the
write path runs the range scan, a write probe, and three env-gated
hooks while reads go straight to memory. (Those captures are from a
build generation with less aggressive cross-crate inlining; on current
builds much of the cost is inlined into the interpreter frame and
attributed there, so per-symbol shares are lower bounds.)

## Why the scan is avoidable

The protected ranges are synthetic-region trap trampolines, allocated
together and far from the heap, stack, and framebuffer that real guest
writes target. A bounding box over the ranges rejects the overwhelmingly
common write in two comparisons; only writes landing inside the box need
the exact scan.

## Expected improvement

A deterministic headless A/B isolating the scan shows removing it
removes ~70% of `write_byte`'s self time. A same-generation capped
gameplay A/B measures `write_byte` self time −76% (1.51% → 0.36%) and
all bus writes 2.51% → 0.64%, a −1.87pp lower bound on non-idle CPU
(inlined copies receive the same fix but cannot be separately counted).

Fix in preparation: widen-only bounding box ahead of the exact scan.

rlanday:
Closes #453.

## Summary

Every guest write consults `readonly_code_overlaps`, which scanned the
protected-range list linearly. This adds a bounding box over those ranges
so the overwhelmingly common write rejects in two comparisons.

## Why it matters

Reads pay nothing comparable, and the asymmetry is stark. In a capped
(`--cpu-mhz 25`) gameplay capture, decomposed across all threads with
blocking waits excluded:

| Method | capture A | capture B |
|---|---:|---:|
| `write_byte` | **6.04%** | **5.66%** |
| `write_word` | 2.40% | 2.44% |
| `write_long` | 1.24% | 1.32% |
| `read_byte` | **0.06%** | **0.05%** |
| all writes | 9.68% | 9.42% |

`write_byte` costs ~100× `read_byte` (95× and 112× in the two captures)
because it runs the range scan, a write probe, and three env-gated hooks,
while `read_byte` goes straight to memory.

The protected ranges are synthetic-region trap trampolines, allocated
together and well away from the heap, stack and framebuffer that real guest
writes target — so a bounding box rejects almost every write without
touching the list.

## Measured effect

Deterministic headless A/B, two profiles per arm:

| | base | candidate |
|---|---:|---:|
| `write_byte` share of samples | 0.53% | **0.16%** |

**This is the load-bearing result for the claim above.** `readonly_code_overlaps`
is inlined into `write_byte`, so a profile alone cannot separate the scan's
cost from the rest of the function — the causal story had to be tested, not
asserted. Removing the scan removes ~70% of `write_byte`'s self time, which
confirms the scan was its dominant cost.

The absolute figure here is small because this startup-weighted workload does
few guest writes; the role of this A/B is the causal test, not the magnitude.
The gameplay magnitude is measured directly in the next section.

Base-to-base spread on this metric is 0.01pp.

## Measured in gameplay (same-generation A/B)

Two capped (`--cpu-mhz 25`) GUI sessions on identical binaries except this
fix, host-sampled for 90 s of play each, decomposed across all threads
with blocking waits excluded:

| metric | baseline | with fix |
|---|---:|---:|
| `write_byte` self time | 1.51% | **0.36%** (−76%) |
| all bus writes | 2.51% | **0.64%** |

The −76% matches the earlier headless prediction (~70%), confirming the
range scan was the dominant cost of the write path.

Why these absolute shares are smaller than the ~6% quoted above: the two
capture generations attribute differently. This build generation's
cross-crate LTO inlines most bus accesses into the interpreter frame,
where they are invisible to symbol attribution — only the out-of-line
copies are counted here. The measured −1.87pp of non-idle CPU is
therefore a **lower bound** on the fix's total effect; the inlined copies
receive the same fix but cannot be separately counted. The sessions are
human-played and not length-matched, so no throughput claim is made.

The measured binaries were built on the 0.12.2-generation master; this
branch is based on current master (0.12.8). `src/memory/bus.rs` is
unchanged between the two, so the measurements apply to the submitted
code exactly.

## How this was measured

Primary evidence is a capped (`--cpu-mhz 25`) GUI capture of steady-state
gameplay — the configuration the application actually runs in. The cap matters
because uncapped execution is elastic: the guest does as much work as the host
allows, so CPU share conflates efficiency with throughput.

The capture is decomposed across **all threads with blocking waits excluded**.
Scoping to the emulator thread alone hides real work on the audio and render
threads, and including waits swamps the sample with idle: in these captures
over 90% of raw samples are blocking syscalls. Two independent captures agree
within a point on the figures above.

Before/after is measured on the deterministic `--headless` harness using
share-of-samples per symbol. Wall clock is not usable there: the process is
killed once sampling finishes, so every run reports the same duration by
construction, and wall-clock A/B on this machine drifted up to 42% between
runs of an *identical* binary in any case.

## Correctness

The box is only ever **widened**, never narrowed, so it cannot reject a
write the scan would have allowed; it only ever short-circuits to the same
answer. When an address does land inside the box, the existing scan runs
unchanged.

The added test covers what a wrong implementation would get wrong: two
disjoint protected ranges with a writable gap between them, so the box
deliberately spans memory it must **not** protect, plus protected writes at
all three widths and ordinary writes on both sides of the box.

Also hoists a duplicated `fb_write_trace_range()` lookup in `write_byte`
so the untraced path reads it once rather than twice.

## Validation

- New test: `readonly_code_protection_survives_the_bounding_box_fast_path`

### DIFF
--- a/src/memory/bus.rs
+++ b/src/memory/bus.rs
@@ -1111,9 +1111,9 @@ impl MacMemoryBus {
         if end <= span_start as u64 || address as u64 >= span_end as u64 {
             return false;
         }
-        self.readonly_code_ranges.iter().any(|&(start, stop)| {
-            (start as u64) < end && (stop as u64) > address as u64
-        })
+        self.readonly_code_ranges
+            .iter()
+            .any(|&(start, stop)| (start as u64) < end && (stop as u64) > address as u64)
     }
 
     /// Allocate memory from the heap with a stronger start-address alignment.
--- a/src/memory/bus.rs
+++ b/src/memory/bus.rs
@@ -452,6 +452,13 @@ pub struct MacMemoryBus {
     synthetic_ptr: u32,
     /// Guest writes cannot modify synthesized ROM-like instruction stubs.
     readonly_code_ranges: Vec<(u32, u32)>,
+    /// Half-open bounding box over `readonly_code_ranges`, maintained on
+    /// insert. Every guest write consults the protection list, but the
+    /// ranges are synthetic-region trampolines clustered well away from the
+    /// heap, stack and framebuffer that real writes target -- so this lets
+    /// the overwhelmingly common case reject in two comparisons instead of
+    /// scanning the list.
+    readonly_code_span: Option<(u32, u32)>,
     /// Free list: maps aligned_size → stack of recycled addresses
     free_blocks: HashMap<u32, Vec<u32>>,
     /// Tracks the aligned size of each allocation (address → aligned_size)
@@ -836,6 +843,7 @@ impl MacMemoryBus {
             heap_ptr: 0x200000, // Start heap at 2MB
             synthetic_ptr: screen_buffer_start,
             readonly_code_ranges: Vec::new(),
+            readonly_code_span: None,
             free_blocks: HashMap::new(),
             alloc_sizes: HashMap::new(),
             alloc_bucket_sizes: HashMap::new(),
@@ -915,6 +923,7 @@ impl MacMemoryBus {
             heap_ptr: 0x200000,
             synthetic_ptr: screen_buffer_start,
             readonly_code_ranges: Vec::new(),
+            readonly_code_span: None,
             free_blocks: HashMap::new(),
             alloc_sizes: HashMap::new(),
             alloc_bucket_sizes: HashMap::new(),
@@ -1076,8 +1085,12 @@ impl MacMemoryBus {
 
     pub(crate) fn protect_readonly_code(&mut self, address: u32, len: u32) {
         if len != 0 {
-            self.readonly_code_ranges
-                .push((address, address.saturating_add(len)));
+            let end = address.saturating_add(len);
+            self.readonly_code_ranges.push((address, end));
+            self.readonly_code_span = Some(match self.readonly_code_span {
+                Some((lo, hi)) => (lo.min(address), hi.max(end)),
+                None => (address, end),
+            });
         }
     }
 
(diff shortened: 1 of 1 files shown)


benletchford (changes requested):
Please link a public Systemless issue that records the observed guest-write cost and expected improvement, then add `Closes #…` here. Also format the new `readonly_code_ranges` iterator chain in `src/memory/bus.rs`; rustfmt reports that new block. The protection logic and regression coverage otherwise look sound: the focused test and the full Systemless library suite pass.

rlanday:
Both points addressed:

- **Issue**: #453 now records the observed guest-write cost (the capped-capture shares, the ~100× write/read asymmetry, and the attribution caveat) and the expected improvement, and the PR body opens with `Closes #453`.
- **rustfmt**: the `readonly_code_ranges` iterator chain in `src/memory/bus.rs` is formatted (`rustfmt` on that file; the remaining `--check` diff in `src/bin/systemless.rs` predates this branch and is untouched here).

No functional changes in the new commit.

benletchford (approved):
The current Systemless change looks sound. The focused read-only span regression passes, the complete Systemless library suite passes (2,931 passed, 3 ignored), and `src/memory/bus.rs` passes rustfmt. Issue #453 now records the measured write-path cost and the pull request closes it. The bounding-box fast path preserves the existing per-range scan for writes inside the span, including writable gaps between disjoint protected ranges.

Issue #456, opened 2026-08-09
Launch-path body inlined into the per-batch hot loop


rlanday:
## The problem

`switch_to_launched_application` runs at most once per application
launch, but the optimiser inlines its 93-line body into
`service_pending_launch_application`, which is called from three sites
in the per-batch emulator loop. The inlining leaves the caller with a
34 KB stack frame — visible as `addq $0x84e8, %rsp` in its epilogue —
so the hot path pays large-frame setup and code-size costs on every
batch to support a body that almost never executes.

## How this was measured

**Harness:** `systemless --headless --max-instructions 900000000` —
boots the guest and runs it with no window, no input, and no frame
pacing, so a run is deterministic and reproducible without a human
session. Note the scope this implies: headless is startup- and
dialog-weighted, and its event/batch loop runs unpaced (flat-out),
unlike the paced GUI.

**Instrument:** `/usr/bin/sample <pid> 90 1` — a 90-second host profile
at 1 ms sampling, taken after warm-up. Two independent profiles per arm.

**Metric:** each symbol's share of non-idle self-time samples, across
all threads, with blocking-wait samples (semaphores, mach_msg, kevent…)
excluded. Share of samples rather than wall clock because this harness
kills the process when sampling ends — every run reports the same
duration by construction — and because wall-clock A/B on the same
machine drifted up to 42% between runs of an identical binary.

**The two arms ("combined-branch A/B"):** base is `master`; the
candidate is a branch carrying all three hygiene changes together —
the window-list clone removal (#454), the native-trap table scan
(#455), and this outlining change. The candidate was not built with
this change alone. Per-change attribution therefore rests on symbol
specificity: `service_pending_launch_application` is touched by none
of the other two changes, so movement in *its* self time is
attributable to this one.

## Result

| `service_pending_launch_application` self time | profile 1 | profile 2 |
|---|---:|---:|
| base | 2.66% | 2.66% |
| with the attributes | 0.42% | 0.50% |

Base-to-base spread on this metric is 0.00pp, so the drop is far above
noise. Reading it plainly: with the launch body inlined, this function
consumed ~2.7% of all non-idle headless CPU purely as per-batch
overhead; outlining the once-per-launch body leaves a ~0.5% residue
(the `Option` check and call bookkeeping).

## Scope

A headless-harness result: the per-batch loop runs unpaced there, which
is what makes a per-batch fixed cost this visible. No gameplay-CPU claim
[...]

Issue #455, opened 2026-08-09
Native-trap table costs more than the lookup it protects


rlanday:
`native_trap_table` is consulted on every A-line trap dispatch but holds only the handful of traps a program installs native handlers for. At that size the HashMap costs more than the lookup: SipHash dominated hashing samples, and with a cheap hasher the SwissTable group probe remained the self-time hotspot of `TrapDispatcher::dispatch` (`pcmpeqb`/`pmovmskb`/`tzcnt` in disassembly). Measured expectation from the combined-branch A/B: SipHash share of samples 4.6% → 2.9% (noise 0.10pp); the probe removal itself is below harness resolution and is claimed as simplification, not a win. Fix in preparation: a linear scan over the few `u16` keys behind the same HashMap-shaped API.

rlanday:
Clarifying exactly what was measured here, since "SipHash share" is easy to misread:

**Instrument and workload.** `/usr/bin/sample` at 1 ms over a 90 s window, on the deterministic headless harness (`--headless --max-instructions 900000000`), two profiles per arm, A/B = master vs the combined hygiene branch. The metric is **share of non-idle self-time samples per symbol** (all threads, blocking waits excluded). Share-of-samples rather than wall clock because this harness kills the process when sampling ends — every run reports the same duration by construction — and wall-clock A/B on the same machine drifted up to 42% between runs of an identical binary.

**What "SipHash share" means.** The fraction of those samples attributed to the standard library's SipHash frames — i.e. hashing time for the *whole process*, not this table alone. The old `native_trap_table` was a std `HashMap<u16, u32>` with the default hasher, so every A-line dispatch hashed the trap word before probing. Measured: 4.62% / 4.52% (base, two profiles) → 3.19% / 2.62% (with fix), base-to-base noise 0.10pp. It does not go to zero because other HashMaps in the process still hash; the delta (~1.5–1.9pp) is the part this table accounted for.

**What could not be measured.** The motivating disassembly showed the SwissTable group probe (`pcmpeqb`/`pmovmskb`/`tzcnt`) as the self-time hotspot *within* `dispatch`, but the A/B could not resolve a `dispatch` self-time improvement: it is flat within noise (+0.9% when normalised against `run_steps_internal` — normalised because raw shares inflate when other symbols shrink). An earlier round that appeared to show the probe removal was sleep-contaminated and was withdrawn. So the claim structure is: **measured win = the SipHash elimination; the probe removal = code simplification only.**

[...]

PR #471, opened 2026-08-09
perf: stop cloning the window list on every event poll

### LINKED ISSUE #454: Event poll clones the window list on every call
`pending_update_event` clones the whole window list purely to iterate it; the clone exists only so the front window can be pushed as a fallback when the list is empty. The poll runs constantly (runner poll, dialog filter real-event check, window trap handler), and in deterministic headless profiles this one clone is essentially all allocator traffic: allocator share of samples 19.04%, with 6,010/7,201 sampled allocations under `trap::window::*pending*` across two profiles. Scope: this is an idle-event-loop/headless cost — in two capped gameplay captures the path is 0.00% of non-idle CPU, so no frame-rate effect is expected; the wins are headless throughput and idle CPU/battery in the GUI. Expected improvement: allocator share → ~0 and the named call path → exactly 0 sampled allocations (a falsifiable positive control). Fix in preparation: borrowing iterator with a chained fallback.

rlanday:
Closes #454.

## Summary

`pending_update_event` no longer clones the window list to iterate it.
The clone existed only so the front window could be pushed as a fallback
when the list is empty; a borrowing iterator with a chained fallback
expresses the same thing without allocating. Reached by three call paths
per poll (runner poll, dialog filter real-event check, window trap
handler).

## How this was measured

**Harness:** `systemless --headless --max-instructions 900000000` —
deterministic (no window, no input, no frame pacing), which also means
startup/dialog-weighted and unpaced. **Instrument:** `/usr/bin/sample
<pid> 90 1` (90 s at 1 ms) after warm-up, two independent profiles per
arm. **Metric:** each symbol's share of non-idle self-time samples,
all threads, blocking-wait samples excluded — share rather than wall
clock because the harness kills the process when sampling ends and
wall-clock A/B drifted up to 42% between identical binaries. **Arms:**
base is `master`; the candidate carried all three hygiene changes
(#454, #455, #456) together, so per-change attribution rests on symbol
specificity — the rows below are symbols only this change touches.

## Result

| metric | base (2 profiles) | with fix | base-to-base noise |
|---|---:|---:|---:|
| allocator share of samples | 19.15% / 19.37% | **0.05% / 0.08%** | 0.22pp |
| samples allocating under `trap::window::*pending*` | 6,205 / 6,694 | **0 / 0** | — |

The second row is the positive control: the specific call path this
removes goes to exactly zero — a falsifiable named-symbol prediction.

**Scope, stated plainly:** this is a headless-throughput and idle-CPU
improvement, not a frame-rate one. In two capped gameplay captures this
path is 0.00% of non-idle CPU. Anyone running `--headless` (tests, CI,
batch) executes the poll unpaced where this cost dominates; a paced GUI
still pays the clone at frame cadence while idle in dialogs and menus.

## Coverage

New test `pending_update_event_selects_front_to_back_and_falls_back_when_list_is_empty`
pins the behavior the change must preserve: front-to-back selection
(first dirty window wins, clean front skipped), visibility gating
(hidden dirty window skipped), the empty-list front-window fallback,
no event when the fallback is unset, and the event-mask gate.

## Validation

- No behavioural change: same iteration order, same fallback condition,
  same returned event

### DIFF
--- a/src/trap/window.rs
+++ b/src/trap/window.rs
@@ -4810,7 +4810,11 @@ mod tests {
         );
         let region_ptr = bus.alloc(10);
         bus.write_word(region_ptr, 10);
-        let rect = if dirty { (0i16, 0i16, 10i16, 10i16) } else { (0, 0, 0, 0) };
+        let rect = if dirty {
+            (0i16, 0i16, 10i16, 10i16)
+        } else {
+            (0, 0, 0, 0)
+        };
         bus.write_word(region_ptr + 2, rect.0 as u16);
         bus.write_word(region_ptr + 4, rect.1 as u16);
         bus.write_word(region_ptr + 6, rect.2 as u16);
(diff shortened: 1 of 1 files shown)


benletchford (changes requested):
`cargo fmt --all -- --check` reports pre-existing drift elsewhere, but it also reports new drift in this change's `make_window` helper and regression test: the `rect` conditional, event extraction, and final assertion all need rustfmt's multiline form. Please format only the added code so this pull request does not increase formatting drift. The focused regression and full `cargo test --lib` suite pass on the current Systemless base.

rlanday:
Formatted exactly the added code — the `make_window` helper's `rect` conditional, the event extractions, and the final assertion now take rustfmt's multiline form — while the two pre-existing drift regions elsewhere in `window.rs` are deliberately untouched, so the change adds no drift and removes what it introduced. Focused regression and the full library suite pass on the branch.

benletchford (approved):
Verified against the current Systemless base: the focused update-event regression, full library suite, and headless check pass. The iterator preserves front-to-back selection, visibility gating, update-mask handling, and the empty-list fallback while removing the per-poll allocation.

PR #472, opened 2026-08-09
perf: direct-index the native-trap table

### LINKED ISSUE #455: Native-trap table costs more than the lookup it protects

rlanday:
Closes #455.

## Summary

`native_trap_table` becomes a **direct-indexed table** over the two
normalized trap bands. Every key producer — dispatch's
`0xA800 | (trap & 0x03FF)` / `0xA000 | (trap & 0x00FF)` and
`trap_address_table_key` behind `SetTrapAddress`/`NSetTrapAddress` in
all its variants — normalizes into the Operating System band
(`0xA000..=0xA0FF`, 256 slots) or the Toolbox band (`0xA800..=0xABFF`,
1,024 slots), so a 1,280-slot array covers the entire reachable key
space: **O(1) at any occupancy**, no hashing, no probe, ~10 KB
allocated once. Off-band keys (which no production caller generates)
go to a spill list so the `HashMap`-shaped contract stays total; the
dispatch path never touches it because its keys are in-band by
construction. All 22 call sites are unchanged.

This revision replaces the linear-scan version reviewed earlier:
`SetTrapAddress` can populate arbitrarily many slots, so a scan made
the hottest dispatch path O(n) in patched-trap count. Direct indexing
is bounded without a threshold heuristic.

The motivation stands: the previous `HashMap` paid SipHash plus a
SwissTable probe per lookup — SipHash dominated hashing samples, and
with a cheap hasher the group probe remained the self-time hotspot of
`dispatch` (`pcmpeqb`/`pmovmskb` in disassembly).

## How this was measured

**Harness:** `systemless --headless --max-instructions 900000000` —
deterministic (no window, no input, no frame pacing), which also means
startup/dialog-weighted and unpaced. **Instrument:** `/usr/bin/sample
<pid> 90 1` (90 s at 1 ms) after warm-up, two independent profiles per
arm. **Metric:** each symbol's share of non-idle self-time samples,
all threads, blocking-wait samples excluded — share rather than wall
clock because the harness kills the process when sampling ends and
wall-clock A/B drifted up to 42% between identical binaries. **Arms:**
base is `master`; the candidate carried all three hygiene changes
(#454, #455, #456) together, so per-change attribution rests on symbol
specificity — the rows below are symbols only this change touches.

Caveat specific to this row: "SipHash share" is whole-process hashing
time — every std `HashMap` with the default hasher contributes — so it
cannot reach zero; the delta is the part this table accounted for.

## Result (measured on exactly this revision, base = current master)

"Hashing frames" below = the standard library's SipHash `write` plus
`BuildHasher::hash_one` dispatcher frames — the out-of-line hashing
cost the default-hasher table paid per lookup.

| metric | base (2 profiles) | with this table |
|---|---:|---:|
| hashing frames, share of non-idle samples | 5.10% / 5.83% | **3.07% / 3.00%** |
| `TrapDispatcher::dispatch` self time | 5.51% / 5.70% | 5.22% / 5.50% |

The candidate arm's spread is 0.07pp and the base arm's 0.73pp; the
conservative delta (worst candidate vs best base, −2.0pp) is roughly
3× the worst spread. Hashing does not reach zero because other
`HashMap`s in the process still hash; the delta is this table's part.

One claim stays downgraded from the earlier round: disassembly showed
the SwissTable probe as `dispatch`'s self-time hotspot, but `dispatch`
self time is flat within noise here too — the measured benefit is the
hashing elimination; probe removal and the bounded worst case are
design properties, the latter guaranteed by the full-occupancy test
rather than by a profile. Gameplay component is direction-only
(hashing ~1% of non-idle gameplay CPU, of which this table is part).

How the design evolved: the first revision used a linear scan on a
"the table holds a handful of entries" assumption, and the review
correctly rejected it — `SetTrapAddress`/`NSetTrapAddress` put the
occupancy in the guest's hands, making every A-line lookup O(n) in
patched-trap count. The scan-era cache argument against direct
indexing (a sparsely touched multi-KB table) only held under that same
guaranteed-tiny-occupancy assumption; with repetition-driven locality
the hot slots stay resident regardless of table size, so direct
indexing wins on every axis that matters here: O(1) at any occupancy,
no hashing, no probe, no threshold heuristic.

## Coverage

Two tests pin the contract:

- The `HashMap` contract the call sites rely on: `insert` returning the
  previous value on replacement and `None` on first insert, `remove`
  returning the removed value exactly once, present/absent lookups, and
  a replace-then-remove sequence.
- **The many-entry boundary**: both bands filled completely (256 OS +
  1,024 Toolbox words), every lookup exact at full occupancy,
  replacement/removal exact with every slot occupied and neighbors
  surviving, plus off-band words (between and beyond the bands)
  storable/removable through the spill and absent otherwise.

## Validation

### DIFF
--- a/src/trap/dispatch.rs
+++ b/src/trap/dispatch.rs
@@ -983,8 +983,6 @@ pub(crate) struct InverseTableCacheEntry {
     pub bytes: Vec<u8>,
 }
 
-/// Trap dispatcher with resource fork access and emulator state.
-
 /// Native trap dispatch table.
 ///
 /// Consulted on every A-line trap dispatch. The default `HashMap` paid
@@ -1063,12 +1061,9 @@ impl TrapWordMap {
             }
         }
     }
-
-    pub(crate) fn contains_key(&self, trap: &u16) -> bool {
-        self.get(trap).is_some()
-    }
 }
 
+/// Trap dispatcher with resource fork access and emulator state.
 pub struct TrapDispatcher {
     /// Synthetic keyboard and mouse entries exposed by the ADB Manager.
     pub(crate) adb: crate::adb::AdbManager,
--- a/src/trap/dispatch.rs
+++ b/src/trap/dispatch.rs
@@ -984,6 +984,51 @@ pub(crate) struct InverseTableCacheEntry {
 }
 
 /// Trap dispatcher with resource fork access and emulator state.
+
+/// Native trap dispatch table.
+///
+/// Consulted on every A-line trap dispatch, but it holds only the handful of
+/// traps a program has installed native handlers for. At that size a hash map
+/// costs more than the lookup it protects: the default hasher dominated, and
+/// even with a cheap hasher the SwissTable group probe remained visible in
+/// profiles of `dispatch`. A linear scan over a couple of `u16` keys needs
+/// neither.
+///
+/// The API mirrors the `HashMap` methods previously used so call sites are
+/// unchanged.
+#[derive(Debug, Default, Clone)]
+pub(crate) struct TrapWordMap {
+    entries: Vec<(u16, u32)>,
+}
+
+impl TrapWordMap {
+    pub(crate) fn get(&self, trap: &u16) -> Option<&u32> {
+        self.entries
+            .iter()
+            .find(|(word, _)| word == trap)
+            .map(|(_, handler)| handler)
+    }
+
+    pub(crate) fn insert(&mut self, trap: u16, handler: u32) -> Option<u32> {
+        match self.entries.iter_mut().find(|(word, _)| *word == trap) {
+            Some(slot) => Some(std::mem::replace(&mut slot.1, handler)),
+            None => {
+                self.entries.push((trap, handler));
+                None
+            }
+        }
+    }
+
+    pub(crate) fn remove(&mut self, trap: &u16) -> Option<u32> {
+        let index = self.entries.iter().position(|(word, _)| word == trap)?;
+        Some(self.entries.swap_remove(index).1)
+    }
+
+    pub(crate) fn contains_key(&self, trap: &u16) -> bool {
+        self.entries.iter().any(|(word, _)| word == trap)
+    }
+}
+
 pub struct TrapDispatcher {
     /// Synthetic keyboard and mouse entries exposed by the ADB Manager.
     pub(crate) adb: crate::adb::AdbManager,
@@ -1772,7 +1817,7 @@ pub struct TrapDispatcher {
     /// and a native handler exists, the dispatcher simulates a JSR to the handler
     /// instead of running HLE code. This allows CRT-installed handlers (LoadSeg,
     /// UnloadSeg, ExitToShell) to run natively with proper code relocation.
-    pub(crate) native_trap_table: HashMap<u16, u32>,
+    pub(crate) native_trap_table: TrapWordMap,
     /// Most recent original call for each active native trap handler. Entries
     /// are replaced by a newer call and consumed when a handler invokes its
     /// saved old trap address.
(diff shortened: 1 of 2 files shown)


benletchford (changes requested):
`SetTrapAddress` and `NSetTrapAddress` can populate arbitrary OS and Toolbox trap slots, so this table is not bounded to the handful of handlers in the profiled workload. Replacing the hash table changes every A-line lookup from amortized O(1) to O(n); an application that patches many traps can therefore regress the hottest dispatch path. Please preserve bounded lookup—for example with direct indexing or a hash-backed fallback beyond a measured threshold—and add coverage for the many-entry boundary. The current two-key map-contract test establishes behavior but does not cover this scaling risk.

rlanday:
Scaling objection addressed by removing the O(n) design rather than thresholding it: the table now direct-indexes the two normalized trap bands. Every key producer — dispatch's own masks and `trap_address_table_key` behind all `SetTrapAddress`/`NSetTrapAddress` variants — normalizes into `0xA000..=0xA0FF` (256 OS slots) or `0xA800..=0xABFF` (1,024 Toolbox slots), so a 1,280-slot array covers the entire reachable key space: O(1) at any occupancy including full, no hashing, no probe, no threshold heuristic to tune, ~10 KB once. Off-band keys (never produced in production) go to a spill list so the HashMap-shaped contract stays total; dispatch cannot reach the spill because its keys are in-band by construction.

Coverage for the many-entry boundary is in as requested: both bands filled completely (all 1,280 words), every lookup exact at full occupancy, replacement/removal exact with all slots occupied and neighbors surviving, plus off-band words between and beyond the bands exercised through the spill.

The body's measurement section is refreshed with an A/B of exactly this revision against current master (same harness/instrument/metric, two profiles per arm): hashing frames 5.10%/5.83% → 3.07%/3.00% of non-idle samples, `dispatch` self time flat within noise. Full suite green (2,989).

benletchford (changes requested):
`cargo clippy --lib --tests` reports two new warnings introduced by this revision. The existing `/// Trap dispatcher with resource fork access and emulator state.` comment is now separated from `TrapDispatcher` by `TrapWordMap`, so Rust attaches it to the map and reports `empty_line_after_doc_comments`; please move that comment back immediately above `TrapDispatcher`. `TrapWordMap::contains_key` has no callers and reports `dead_code`, so please remove it unless the change needs to use it. Also update the pull-request title to describe direct indexing instead of a linear scan, and remove or rewrite the body paragraph that says direct indexing is worse for cache, since that now contradicts the implementation. The two focused map tests and the full `cargo test --lib` suite pass on the current Systemless base.

rlanday:
All four addressed:

- The `/// Trap dispatcher with resource fork access and emulator state.` doc comment is back immediately above `TrapDispatcher`; `TrapWordMap` keeps its own doc block, and `empty_line_after_doc_comments` is gone.
- `TrapWordMap::contains_key` is removed. Its one caller was the bare-trap-number assertion in the `NSetTrapAddress` test, which the lib pass can't see; that assertion now uses `get(&0x01F4).is_none()` directly, so the behavioral check is preserved without the dead method.
- Title now says direct indexing.
- The alternatives paragraph is rewritten as the design's actual evolution: the scan-era cache argument against direct indexing only held under the same guaranteed-tiny-occupancy assumption the review rejected, and with repetition-driven locality the hot slots stay resident regardless of table size — so it no longer argues against the shipped implementation.

`cargo clippy --lib --tests` clean of both new warnings; the two focused map tests and the full library suite pass.

benletchford (approved):
Verified against the current Systemless base: both trap-table contract regressions, the full library suite, and the headless check pass. Direct indexing covers every normalized OS and Toolbox trap slot with bounded lookup, while the off-band spill preserves the map-shaped API.

PR #502, opened 2026-08-11
fix(control): truncate fixed-width popup titles

### LINKED ISSUE #501: Fixed-width pop-up control titles are not truncated to fit
## Summary

A fixed-width standard pop-up control draws an overlong selected-item title outside its content box. In Damage Incorporated, `Standard/Marathon` runs through the pop-up arrow and into the Cancel-button area; BasiliskII truncates it with an ellipsis inside the control.

Inside Macintosh Volume VI documents that `popupFixedWidth` must truncate content to fit and append ellipses.

## Reproduction

2. Open **PREFS**, select **Environment**, and choose **Use Keyboard & Mouse**.
3. Open **Configure Keyboard**.
4. Inspect the **Layout** pop-up near the lower-right corner.

## Expected

The selected layout is truncated with an ellipsis inside the pop-up content area, leaving the arrow and neighboring Cancel button unobscured.

BasiliskII reference

## Actual

The full `Standard/Marathon` string escapes the control bounds and crosses the Cancel-button area.

Systemless

Observed twice with Systemless 0.12.8 at `d1b2758a02974d80109ef240bfb1eef8d2004355`; the affected frame measured 97.72-97.74% strict whole-frame parity.

## Focused regression test

Create a narrow `popupFixedWidth` control whose selected item is wider than the box. Draw it in normal and inactive states, then assert that the rendered title ends in an ellipsis, no title pixels occur outside the content rectangle, and the arrow remains intact. Also verify that any temporary clipping state is restored.

Before drawing selected-item text, compute the available content width after the inset and arrow area, shorten the Mac Roman string to the longest prefix that fits with `...`, and clip drawing to that content rectangle. Apply the same bounded layout regardless of enabled state, without changing pop-up sizing, selection, or tracking behavior.

benletchford:
## Summary

- shorten overlong pop-up selected-item titles to the longest measured prefix followed by `...`
- apply the same bounded title layout to enabled and inactive controls while retaining inactive palette ink
- clip enabled title glyphs to the content rectangle so they cannot overwrite the arrow or neighboring controls
- add focused framebuffer coverage for both states of the `Standard/Marathon` regression

## Root cause

The standard pop-up renderer sent the complete selected-item title directly to the framebuffer. It did not account for the fixed content width or reserve the arrow area, so overlong text escaped the control bounds. The inactive path also skipped the selected title instead of sharing the same layout.

The new layout follows the documented `popupFixedWidth` behavior while also bounding auto-sized controls if their resolved box is clamped at the screen edge.

## Validation

- `cargo test --lib draw_popup_control_keeps_overlong_title_inside_content_area`

Closes #501

### DIFF
--- a/src/trap/control.rs
+++ b/src/trap/control.rs
@@ -1565,7 +1565,7 @@ impl super::TrapDispatcher {
         );
     }
 
-    fn draw_control_label_text(
+    pub(crate) fn draw_control_label_text(
         &self,
         bus: &mut MacMemoryBus,
         label_top: i16,
--- a/src/trap/dialog.rs
+++ b/src/trap/dialog.rs
@@ -7136,6 +7136,15 @@ impl super::TrapDispatcher {
 
         let font_id = 0i16;
         let font_size = 12i16;
+        let text_x = left + 15;
+        // The standard arrow occupies the right side of the box. Inside
+        // Macintosh Volume VI (1991), p. 3-17 requires popupFixedWidth item
+        // text that does not fit this content area to be truncated with
+        // ellipses. Keep the same bound for auto-sized controls when their
+        // resolved box is clamped to the screen edge.
+        let text_right = (right - 19).max(text_x);
+        let display_title =
+            Self::popup_control_display_title(title, text_right - text_x, font_id, font_size);
 
         if !self.draw_theme_control_chrome(
             bus,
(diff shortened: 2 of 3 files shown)


benletchford:
Blocking: the new truncation path still runs only when `enabled` is true, while #501 explicitly requires the same bounded layout for inactive pop-ups. #506 adds inactive selected-item rendering to this function, so the two changes currently compose into an unclipped inactive title. Please make enabled and inactive titles share the measured ellipsis and content clipping, and cover both states without changing inactive palette ink.

PR #504, opened 2026-08-11
fix(device): hide unavailable AppleTalk drivers

### LINKED ISSUE #484: Synthetic .MPP driver falsely advertises unavailable AppleTalk networking
## Summary

Systemless currently lets `PBOpen(".MPP")` succeed against a synthetic driver even though AppleTalk multiplayer is not implemented. Applications therefore detect networking as available and expose paths that cannot work.

Damage Incorporated demonstrates the mismatch through both of its multiplayer buttons. On a clean reference System 7.5.3 installation with networking unavailable, **GATHER** and **JOIN** are ignored and the game remains on its menu (before its normal attract demo begins). Under Systemless, **GATHER** opens **SETUP NET COMBAT** and **JOIN** advances into network pregame fire-team selection; neither can proceed to a real network session.

This is a capability-advertising problem rather than a dialog-rendering difference: Systemless should not report a classic networking driver as present until the advertised API is usable.

## Reproduction

Archive SHA-256:

```
9c02dead11d05691dfc0c285327ee21cca6b9f1646ab7150b0a84dfbb9986d52
```

Tested with Systemless `d1b2758a02974d80109ef240bfb1eef8d2004355` (`0.12.8`).

1. Launch `Damage Incorporated Demo` from a clean archive.
2. Dismiss the opening title painting to reach the button menu.
3. Click **GATHER** or **JOIN**.
4. Wait five seconds.

## GATHER reference result

The unavailable network action is ignored and the main menu remains visible.

BasiliskII reference: GATHER remains on the menu

## GATHER Systemless result

The game believes networking is available, opens **SETUP NET COMBAT**, advertises Ethernet, and cannot proceed to a real network session.

Systemless: GATHER opens unusable network setup

## JOIN reference result

The unavailable client action is likewise ignored and the menu remains visible.

BasiliskII reference: JOIN remains on the menu

## JOIN Systemless result

The game advances into network pregame fire-team selection despite having no usable network service.

Systemless: JOIN opens network pregame fire-team selection

## Focused regression tests

- Add a Driver Manager test proving that `PBOpen(".MPP")` returns the reference error when no AppleTalk backend is configured.
- Add a capability-contract test proving that every registered synthetic driver supports the control/status operations used to detect its advertised service.
- Add integration tests that launch this archive, select **GATHER** and **JOIN**, and confirm that unsupported network paths remain unavailable rather than opening pregame UI.

1. Stop registering the synthetic `.MPP` driver by default when no networking backend exists.
2. If AppleTalk support is added later, register `.MPP` only when its required open, control, status, and data-path semantics are implemented.
3. Centralize synthetic-driver capability registration so driver presence consistently means that the advertised service is usable.

Fonts are intentionally excluded from this comparison.

benletchford:
## Summary

- stop advertising `.MPP`, `.ATP`, and `.XPP` as available synthetic drivers
- return Device Manager `dInstErr` for those unavailable driver names while ordinary missing files retain File Manager `fnfErr`
- keep the modeled serial input/output drivers available
- add regression coverage proving unavailable AppleTalk opens clear `ioRefNum` and install no synthetic driver

## Root cause

The synthetic-driver allowlist treated AppleTalk driver names as equivalent to modeled serial drivers. `PBOpen(".MPP")` therefore returned `noErr` and a positive refnum even though Systemless has no AppleTalk control, status, or protocol implementation. Applications interpreted that successful open as a usable networking capability.

The allowlist now represents only services backed by the HLE, and missing Device Manager driver names follow the documented driver-installation error path.

## Validation

- `cargo test --lib pbopen_does_not_advertise_unavailable_appletalk_drivers`
- `cargo test --lib pb_open_not_found`
- `cargo test --lib pbopen_synthetic_driver_refnum_supports_read_write_close`
- `cargo test --lib` (2,941 passed; 3 ignored)

Closes #484

### DIFF
--- a/src/trap/resource.rs
+++ b/src/trap/resource.rs
@@ -216,11 +216,11 @@ impl super::TrapDispatcher {
     pub(crate) const RES_MAP_CHANGED_ATTR: u16 = 0x0020;
     pub(crate) const RES_MAP_READ_ONLY_ATTR: u16 = 0x0080;
 
-    fn is_synthetic_driver_name(filename: &str) -> bool {
-        matches!(
-            filename,
-            ".AIn" | ".AOut" | ".BIn" | ".BOut" | ".MPP" | ".ATP" | ".XPP"
-        )
+    fn is_available_synthetic_driver_name(filename: &str) -> bool {
+        // Driver presence is a capability probe. Keep this list limited to
+        // services backed by the HLE: AppleTalk's .MPP/.ATP/.XPP drivers must
+        // remain absent until their control, status, and protocol paths exist.
+        matches!(filename, ".AIn" | ".AOut" | ".BIn" | ".BOut")
     }
 
     fn address_is_loaded_code(&self, address: u32) -> bool {
@@ -3774,7 +3774,7 @@ impl super::TrapDispatcher {
                     bus.write_word(pb + 16, 0); // noErr
                     cpu.write_reg(Register::D0, 0);
                     eprintln!("[TRAP] PBOpen -> refnum={} vfs=\"{}\"", refnum, vfs_name);
-                } else if Self::is_synthetic_driver_name(&filename) {
+                } else if Self::is_available_synthetic_driver_name(&filename) {
                     let refnum = self.next_refnum;
                     self.next_refnum += 1;
                     self.synthetic_drivers.insert(refnum, filename.clone());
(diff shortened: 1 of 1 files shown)


benletchford:
Blocking: this fallback returns `fnfErr` (`-43`), a File Manager result, for unavailable driver names. `PBOpen` is servicing a Device Manager open here; Inside Macintosh Volume II, II-183 lists `dInstErr` (`-26`) when the requested driver cannot be found in the resource file.

Please keep clearing `ioRefNum`, but return `dInstErr` in D0 and `ioResult` for `.MPP`, `.ATP`, and `.XPP`, then update the regression expectations. Focused coverage should also distinguish an unavailable driver open from an ordinary missing-file open, which must continue returning `fnfErr`.

benletchford:
Blocking: the two new driver-name helpers use exact `matches!` comparisons, but the Device Manager ignores case when comparing driver names (Inside Macintosh Volume II, II-175). As written, `PBOpen(".mpp")` falls through to File Manager `fnfErr` instead of the intended Device Manager `dInstErr`, and `PBOpen(".ain")` no longer resolves the modeled serial input driver. Please make both checks case-insensitive and add mixed-case regression coverage for one available serial driver and all three unavailable AppleTalk drivers.

benletchford:
Fixed in `601605fa`. Device Manager driver-name matching is now case-insensitive for both modeled serial drivers and unavailable AppleTalk drivers. Mixed-case regressions cover `.ain`, `.mpp`, `.AtP`, and `.xPp`; unavailable drivers return `dInstErr`, while an ordinary missing file still returns `fnfErr`.

Validation: `cargo fmt --all -- --check`, focused available/unavailable/missing-file PBOpen regressions, `cargo check --no-default-features`, and `git diff --check`.

PR #505, opened 2026-08-11
fix(sound): dispatch legacy Sound Driver writes

### LINKED ISSUE #485: FSWrite rejects legacy Sound Driver writes to reference number -4
## Observed behavior

Launching Shufflepuck Cafe and selecting Nerual reaches the match, but the game sends a 5,024-byte `FSWrite` request to Sound Driver reference number `-4`. Systemless reads the 16-bit field as unsigned `65532`, does not find it in the open-file table, and returns `rfNumErr` (`-51`):

```text
[TRAP] FSWrite ref=65532 buf=$002154FC count=5024 posMode=0 posOff=0
[TRAP] FSWrite -> rfNumErr (refnum 65532 not open)
```

This discards a valid legacy sound payload. The visual match continues, so an image-only gameplay check does not expose the failed device write.

## Expected behavior

Classic Device Manager reference numbers are signed 16-bit values. Inside Macintosh: Devices assigns `.Sound` unit 3/reference number `-4` and documents `FSWrite`/`PBWrite` as the interface for sending data to an opened device driver. Systemless should dispatch this request to its Sound Driver compatibility path instead of treating it as an unknown file reference.

## Reproduction

1. Launch Shufflepuck Cafe in Systemless.
2. Enter the cafe and select Nerual (the bartender portrait at the upper left).
3. Let the match initialize.
4. Observe the `FSWrite` request above and the returned `rfNumErr`.

The same request occurs at match startup independent of the opponent; Nerual is simply the route used to capture the evidence.

- Exercise `FSWrite` with `ioRefNum = -4` and a valid legacy Sound Driver buffer; assert that it does not return file-manager `rfNumErr`.
- Assert that `ioRefNum` is interpreted as signed before choosing file versus Device Manager dispatch.
- Assert that a valid `.Sound` write reports the transferred byte count and reaches the sound backend.
- Retain `rfNumErr` for genuinely unknown positive file reference numbers.

Decode `ioRefNum` as `i16` at the shared read/write dispatch boundary. Route negative reference numbers through the Device Manager/unit-table compatibility layer before consulting the VFS open-file table, then implement `.Sound` write parsing/playback against the existing sound backend. Keep this generic to driver reference numbers rather than special-casing Shufflepuck Cafe.

benletchford:
## Summary

- treat negative `FSWrite` reference numbers as Device Manager drivers instead of File Manager refs
- decode legacy `.Sound` free-form, square-wave, and four-tone synth buffers into the existing PCM mixer
- interpret each `FTSoundRec` phase as its documented waveform byte offset before applying 16.16 rates
- bound decoded legacy PCM expansion with checked arithmetic and fallible allocation
- keep the internal legacy channel outside the guest heap and preserve existing positive-refnum file behavior

## Root cause

`FSWrite` read `ioRefNum` as an unsigned value, so the ROM Sound Driver reference number `-4` became `65532` and failed the File Manager lookup with `rfNumErr`. The four-tone decoder also treated a raw byte-offset phase as if it were already a 16.16 accumulator, causing nonzero starting phases to select the wrong waveform byte. Finally, compact square-wave records could expand guest-controlled durations into unbounded host allocations.

## Validation

- `cargo test --lib four_tone_phase_is_a_waveform_byte_offset`
- `cargo test --lib legacy_sound_driver_decodes_supported_synths`
- `cargo test --lib square_wave_synth_rejects_oversized_decoded_output`
- `cargo test --lib` (2,942 passed; 3 ignored)

Closes #485

### DIFF
--- a/src/trap/dispatch.rs
+++ b/src/trap/dispatch.rs
@@ -1351,6 +1351,9 @@ pub struct TrapDispatcher {
     pub(crate) open_files: HashMap<u16, String>,
     /// Synthetic Device Manager drivers opened by name via PBOpen/OpenDriver.
     pub(crate) synthetic_drivers: HashMap<u16, String>,
+    /// Guest SndChannel storage used by writes to the ROM Sound Driver
+    /// reference number (-4). Allocated lazily on the first StartSound write.
+    pub(crate) legacy_sound_driver_channel: Option<u32>,
     /// Refnums opened with write permission (fsRdWrPerm=3 or fsWrPerm=2).
     /// Used to enforce opWrErr (-49) per IM:Files 9578.
     pub(crate) write_refnums: std::collections::HashSet<u16>,
@@ -2958,6 +2961,7 @@ impl TrapDispatcher {
             working_directories: HashMap::new(),
             open_files: HashMap::new(),
             synthetic_drivers: HashMap::new(),
+            legacy_sound_driver_channel: None,
             write_refnums: HashSet::new(),
             file_positions: HashMap::new(),
             recent_file_read: None,
--- a/src/trap/resource.rs
+++ b/src/trap/resource.rs
@@ -3918,8 +3918,22 @@ impl super::TrapDispatcher {
                 let pos_offset = bus.read_long(pb + 46) as i32;
                 eprintln!(
                     "[TRAP] FSWrite ref={} buf=${:08X} count={} posMode={} posOff={}",
-                    ref_num, buffer, request_count, pos_mode, pos_offset
+                    ref_num as i16, buffer, request_count, pos_mode, pos_offset
                 );
+
+                let signed_ref_num = ref_num as i16;
+                if signed_ref_num < 0 {
+                    let result =
+                        self.write_device_driver(bus, signed_ref_num, buffer, request_count);
+                    let (err, actual_count) = match result {
+                        Ok(actual_count) => (0i16, actual_count),
+                        Err(err) => (err, 0),
+                    };
+                    bus.write_long(pb + 40, actual_count as u32);
+                    bus.write_word(pb + 16, err as u16);
+                    cpu.write_reg(Register::D0, err as i32 as u32);
+                    return Some(Ok(()));
+                }
 
                 let Some(filename) = self.open_files.get(&ref_num).cloned() else {
                     if self.synthetic_drivers.contains_key(&ref_num) {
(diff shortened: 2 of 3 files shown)


benletchford:
Blocking: `FTSoundRec` stores each `soundNPhase` as a byte offset into the 256-byte waveform, but `decode_four_tone_synth` treats the raw value as 16.16 fixed-point by shifting it right 16 bits. A nonzero phase such as 1 therefore starts at byte 0 instead of byte 1. Keep the accumulator in fixed point by initializing it from `(phase as u64) << 16`, and add a focused nonzero-phase regression so the documented starting offset is preserved.

benletchford:
Blocking: `decode_square_wave_synth` grows `samples` from the guest-controlled triplet count and 16-bit duration without any total output bound. A valid-sized `StartSound` buffer can contain thousands of `duration = 65535` triplets; the loop will attempt to allocate and render hundreds of gigabytes before `FSWrite` returns.

Please use checked arithmetic and enforce a supported decoded-output budget before allocating, returning `memFullErr` or `paramErr` when the request exceeds it. Add a focused regression with multiple maximum-duration triplets that proves the decoder rejects the request without a large allocation, while preserving ordinary square-wave, four-tone, and free-form synthesis.

benletchford:
Blocking: the current Systemless candidate is not rustfmt-clean. `cargo fmt --all -- --check` reports changes in `src/trap/sound.rs` around legacy channel insertion, waveform validation, and the test imports. Please run rustfmt on the touched file and push the formatting result; the formatting gate must be clean before merge.

benletchford:
Fixed in `1fe0747a`. The branch is rebased onto the current default branch and the sound implementation plus regressions are rustfmt-clean.

Validation: `cargo fmt --all -- --check`, focused legacy-driver, four-tone phase, and oversized-output regressions, `cargo check --no-default-features`, and `git diff --check`.

benletchford:
Additional real-world validation: Where in the World Is Carmen Sandiego? Macintosh v1.3 from https://archive.org/details/where-in-the-world-is-carmen-sandiego-mac-en (`Carmen World.img` SHA-256 `edcbb1ff8123a9bbbc631315c7f87f998ee6138a9f859df6bfc57f363fd2f556`).

On current `master`, each legacy `FSWrite` to refnum `-4` returns `rfNumErr`, and the game stalls after the first typed detective-name character while waiting for sound completion. On this branch, those writes complete successfully; the same deterministic play sequence creates a detective, enters the first Buenos Aires case, reaches the live investigation interface, and opens the travel map from its first control. The focused gameplay gate passes.

PR #506, opened 2026-08-11
fix(control): preserve disabled popup titles

### LINKED ISSUE #500: Disabled pop-up controls lose their titles under custom indexed palettes
## Summary

Disabled standard pop-up controls can lose both their labels and selected-item titles when an application installs a custom 8-bit palette. This is a semantic control-rendering defect, not a font-shape difference: BasiliskII keeps the five disabled Game Files rows readable, while Systemless reduces the same strings to background-colored or fragmented pixels.

Related: #379 fixed palette-aware inactive text for checkboxes and radio buttons, but the pop-up CDEF still follows a separate inactive rendering path.

## Reproduction

2. Open **PREFS**.
3. Open the category pop-up and select **Environment**.
4. Inspect the disabled **Game Files** rows: Map, Shapes, Physics, Sounds, and Voices.

## Expected

The inactive labels and selected values remain visibly dimmed: `Map: Demo Map`, `Shapes: Demo Shapes`, `Physics: Default`, `Sounds: Demo Sounds`, and `Voices: Demo Voices`.

BasiliskII reference

## Actual

The row labels become dotted/fragmented and the selected values disappear.

Systemless

Observed twice with Systemless 0.12.8 at `d1b2758a02974d80109ef240bfb1eef8d2004355`; both runs produced the same affected frame (95.16% strict whole-frame parity).

## Focused regression test

Create a standard pop-up control on an 8-bit device whose live CLUT has no intermediate neutral gray, populate it with a selected item, set `contrlHilite` to 255, and draw it. Assert that both the external label and selected-item text retain non-background pixels in a palette-appropriate inactive ink. Keep an enabled companion control in the fixture to prove active rendering is unchanged.

Route the pop-up label and selected-item title through the same live-device-CLUT-aware inactive text policy used by the standard checkbox/radio path. Preserve the pop-up frame, arrow, title layout, and Control Manager state; avoid checkerboard-clearing text that has already resolved to an unusable palette index.

benletchford:
## Summary

- preserve each standard pop-up control's creation-time title width after publishing `contrlMin = 1` and `contrlMax = menu item count`
- draw inactive external labels and selected-item titles with the existing live-device-CLUT-aware inactive ink
- bound enabled and inactive selected-item titles to the pop-up content area
- leave pop-up chrome, tracking, and theme state unchanged

## Root cause

The pop-up renderer reused the current `contrlMax` as the title width even though the standard CDEF contract repurposes that field for the menu item count after creation. The creation path cached the width but initially failed to publish the documented post-creation range. Separately, inactive labels were checker-cleared after drawing with active ink, while inactive selected-item text was skipped entirely. Under a custom indexed palette, both paths made the text unreadable.

## Validation

- `cargo test --lib inactive_popup_label_and_selected_title_use_live_device_palette`
- focused assertions verify `contrlMin` and `contrlMax` immediately after control creation

Closes #500

### DIFF
--- a/src/trap/control.rs
+++ b/src/trap/control.rs
@@ -988,7 +988,16 @@ impl super::TrapDispatcher {
             // in contrlMax, so retain the pixel width as private CDEF state.
             // Macintosh Toolbox Essentials 1992, pp. 5-25 to 5-27.
             self.popup_control_title_widths.insert(ctrl_ptr, max);
-            self.create_popup_control_private_data(bus, min)
+            let data = self.create_popup_control_private_data(bus, min);
+            let item_count = self
+                .menus
+                .iter()
+                .rev()
+                .find(|menu| menu.id == min)
+                .map_or(0, |menu| menu.items.len().min(i16::MAX as usize) as i16);
+            bus.write_word(ctrl_ptr + 20, 1);
+            bus.write_word(ctrl_ptr + 22, item_count as u16);
+            data
         } else {
             0
         };
@@ -5263,10 +5272,9 @@ mod tests {
                 1009,
                 0,
             );
+            assert_eq!(bus.read_word(ctrl_ptr + 20), 1);
+            assert_eq!(bus.read_word(ctrl_ptr + 22), 1);
             bus.write_byte(ctrl_ptr + 17, hilite);
-            // The standard pop-up CDEF replaces the creation-time title width
-            // in contrlMax with the number of menu items.
-            bus.write_word(ctrl_ptr + 22, 1);
             assert_eq!(disp.popup_control_title_width(ctrl_ptr, 1), 60);
             disp.draw_control(&mut cpu, &mut bus, ctrl_ptr);
         }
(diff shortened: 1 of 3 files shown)


benletchford:
Blocking: `initialize_control_record` stores the creation-time title width in `popup_control_title_widths`, but it still leaves that width in `ControlRecord.contrlMax`. Macintosh Toolbox Essentials (1992), pp. 5-25–5-27 says the pop-up definition replaces `contrlMin` with 1 and `contrlMax` with the menu item count after creation. The new regression writes `contrlMax = 1` manually, so it cannot catch the missing creation postcondition.

Please make pop-up creation publish the documented minimum and maximum values while retaining the creation-time title width for rendering, and extend the regression to assert both values immediately after `create_control_record` without manually rewriting the record.

benletchford:
Fixed in `b1d72d00`. Pop-up creation now preserves the creation-time title width separately while publishing `contrlMin = 1` and `contrlMax = menu item count`; the regression asserts those postconditions immediately after creation. The branch is also rebased onto the current default branch.

Validation: `cargo fmt --all -- --check`, focused pop-up control tests, `cargo check --no-default-features`, and `git diff --check`.

benletchford:
Correction: the pushed head commit is `979d2a32`.

PR #508, opened 2026-08-11
fix: fence disposed modal dialog presses

### LINKED ISSUE #499: EV Override shipyard Leave falls through and launches from the planet
## Observed behavior

In Escape Velocity Override, leaving the Shipyard after viewing a ship information pane skips the landed services screen in Systemless and launches the player directly into space. BasiliskII closes the Shipyard and returns to the Earth services screen, where the player must independently click the planet-level **Leave** button to launch.

The two **Leave** controls overlap: the Shipyard button spans the lower-right of its custom dialog, directly above the planet-level **Leave** control. The divergence is stable across repeated runs. A strict probe for the planet-level button never appears in Systemless even after 500,000,000 instructions; the game is already in live space. Moving the later physical mouse-up outside the underlying button does not change the outcome, which indicates that the held press/application-owned tracking transition has already fallen through while the child dialog is being disposed.

This is a gameplay-affecting input/state transition, not a font difference.

Tested Systemless revision: `d1b2758a02974d80109ef240bfb1eef8d2004355` (`0.12.8`).

Public archive SHA-256: `3be7447ee8c80f99558eb29db58bc51253b63048f9b21888010bb3dd6c2ecb5e`.

## Expected behavior

One press on the Shipyard **Leave** control should close only the Shipyard and return to the landed Earth services screen. Launching should require a separate press on the planet-level **Leave** control.

## Reproduction

1. Launch Escape Velocity Override and create a new pilot.
2. Land on Earth.
3. Open **Visit Shipyard**.
4. Select the Freighter, click **Info**, then click **Done**.
5. Press the Shipyard **Leave** button once and release normally.
6. Observe the state immediately after the Shipyard closes.

BasiliskII returns to the landed services screen. Systemless passes through that state and enters live space without a second click.

## Paired evidence

| Systemless after Shipyard Leave | BasiliskII reference |
| --- | --- |
| Systemless has already launched into space | BasiliskII returns to the landed Earth services screen |

Evidence commit: https://github.com/benletchford/systemless/commit/ffdc1bf35347e3892458bb5e8f82e15c41752199

## Focused regression test

Build an application-owned custom child dialog whose dismiss control overlaps an enabled custom control in its parent surface.

1. Deliver one mouse-down to the child dismiss control and allow the application to dispose the child before the physical mouse-up arrives.
2. Assert that the same held press cannot begin tracking or activate the now-exposed parent control.
3. Deliver the matching mouse-up and assert that the parent remains inactive.
4. Deliver a later independent click to the parent control and assert that it activates exactly once.
5. Record and verify the low-memory button state plus the results of `Button`, `StillDown`, `WaitMouseUp`, `GetNextEvent`, and `WaitNextEvent` across child disposal.

Cover both overlapping and non-overlapping controls, and keep Dialog Manager-owned controls covered separately by the regression from #344.

Trace the mouse lifecycle and front-window transition while an application-owned custom control disposes its dialog during a held press. Compare the physical button state, queued event sequence, hit-testing window, and polling calls against BasiliskII. Preserve the distinction introduced by #344 for Dialog Manager-owned controls, but ensure one physical press cannot be interpreted as a fresh press by a newly exposed application-owned control merely because the window hierarchy changed before release.

The input lifecycle may also relate to the custom-control toggle behavior in #488, but the nested-dialog fall-through has a distinct reproduction and should not be assumed to share a root cause until a focused trace proves it.

benletchford:
Closes #499.

When a ModalDialog-owned button handler disposes the front dialog before the physical mouse release arrives, end the guest-visible held state and retain ownership of the matching release. The pending state records the owned mouse-down event so disposal removes only that event if an application leaves a queued copy; older unrelated input remains ordered and deliverable.

Regression coverage verifies Button, StillDown, and WaitMouseUp observe the ended child press, an earlier unrelated mouse-down remains deliverable, the matching child release is consumed, and a later independent click is delivered normally.

Validation:
- focused modal press-ownership regressions

### DIFF
--- a/src/trap/dialog.rs
+++ b/src/trap/dialog.rs
@@ -9082,8 +9082,16 @@ impl super::TrapDispatcher {
             // 5-100). If the application disposes that dialog before release,
             // discard the already-handled down event so the newly exposed
             // dialog cannot receive the same press.
-            if let Some(idx) = self.event_queue.iter().position(|event| event.what == 1) {
-                self.event_queue.remove(idx);
+            if let Some(owned) = self.pending_modal_dialog_mouse_down.as_ref() {
+                if let Some(idx) = self.event_queue.iter().rposition(|event| {
+                    event.what == owned.what
+                        && event.message == owned.message
+                        && event.where_v == owned.where_v
+                        && event.where_h == owned.where_h
+                        && event.modifiers == owned.modifiers
+                }) {
+                    self.event_queue.remove(idx);
+                }
             }
             self.mouse_button = false;
             self.adb.note_mouse_state(self.mouse_pos, false);
@@ -9569,8 +9577,11 @@ impl super::TrapDispatcher {
         }
     }
 
-    fn consume_or_retain_dialog_mouse_up(&mut self) {
+    fn consume_or_retain_dialog_mouse_up(&mut self, mouse_down: &QueuedEvent) {
         self.pending_modal_dialog_mouse_up = !self.consume_dialog_mouse_up();
+        self.pending_modal_dialog_mouse_down = self
+            .pending_modal_dialog_mouse_up
+            .then(|| mouse_down.clone());
     }
 
     fn dialog_item_screen_rect(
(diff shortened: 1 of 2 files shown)


benletchford:
Blocking: disposal removes the first queued `mouseDown` regardless of which press produced it. If an earlier unrelated `mouseDown` is already queued, that event is discarded while the child-dialog press remains available to the newly exposed parent, reversing event order and leaving the reported fall-through possible.

Please identify the `ModalDialog`-owned down event rather than removing the first event of that type, and add a regression with an earlier unrelated queued down event. The earlier event must remain ordered and deliverable, while only the child press and its matching release are consumed across disposal.

benletchford:
Fixed in `d512f277`. Dialog disposal now removes the specifically retained ModalDialog-owned `mouseDown`, preserving any earlier unrelated queued press in its original order; the owned release is still consumed and later independent clicks remain deliverable.

Validation: `cargo fmt --all -- --check`, the focused ownership/order regression, the existing queued-release regression, `cargo check --no-default-features`, and `git diff --check`.

PR #509, opened 2026-08-11
fix: preserve explicit colors across client-ID palette installs

### LINKED ISSUE #495: Uniform-value SetEntries remaps Prince of Destruction gameplay UI colors
## Observed behavior

Prince of Destruction's live gameplay UI contains stable solid-color substitutions under Systemless that are not font differences. Numeric sidebar fields that are black or white on a reference Macintosh acquire pink and olive rectangular backgrounds. Opening the documented Talk input with `T` adds two olive rectangles along the otherwise white speech bubble's right edge; the same blocks remain on the NPC response bubble after submitting `hello`.

The conversation itself works: text entry, Return submission, the Keeper's response, bubble lifetime, and subsequent movement all agree. Whole-frame parity remains 99.54-99.62%, which hides the small but conspicuous solid-color regions.

Tested with Systemless 0.12.8 at `d1b2758a02974d80109ef240bfb1eef8d2004355` and the public Prince of Destruction archive (SHA-256 `d2080a5462832e2737ec3f706fb8f8c48c4cfe4782048fa15cf79e5730546e79`). Two fresh paired replays reproduced the same blocks at identical coordinates.

## Expected behavior

The gameplay sidebar should retain the reference black/white field backgrounds, and the speech bubbles should have uninterrupted white interiors bounded by their black outlines. Colored text may remain colored, but unused solid bubble areas must not resolve through unrelated olive palette entries.

## Reproduction

1. Launch `MARS™ Master (sw)` from the public Prince of Destruction archive and continue as an evaluation user.
2. Start the default single-player scenario.
3. Observe the Power, Accuracy, Defence, Toughness, Level, and equipment fields in the right sidebar.
4. Press `T`, type `hello`, and inspect the right edge of the editable speech bubble.
5. Press Return and inspect the Keeper's response bubble.

## Trace evidence and likely generic gap

Systemless receives repeated full-range `SetEntries(0, 255, ...)` calls whose `ColorSpec.value` fields are uniformly zero. At the gameplay transition, the trace shows the logical table being installed as the physical device CLUT; representative logical entries include olive and pink colors that then appear in regions the reference runtime presents as white or black.

This is related to the Color Manager lifecycle investigated in #347, but it is a distinct later-game sequence: #347 fixed the registration fade by preserving the stable physical palette for its final uniform-client-ID table. The same semantic distinction is not being retained across the later gameplay palette installation.

Before changing behavior, extend the reference trace around the gameplay `SetEntries` call to record the logical GDevice table, physical display table, `ctSeed`, and inverse-table state. The implementation should then model uniform `ColorSpec.value` fields as client IDs where the classic Color Manager does so, preserve the independently valid physical mapping, and invalidate/rebuild inverse-search state coherently. Do not special-case Prince of Destruction, a palette entry, or speech-bubble coordinates.

- Create an 8-bit GDevice with a known physical table, then issue a full-range `SetEntries` table whose entries use uniform client IDs but carry different logical RGB values. Assert the logical table and physical display mapping separately against a reference fixture.
- Cover the same operation after a dimmed fade, without a fade, and after repeated full-table installs so the #347 fix is not limited to one transition shape.
- Draw solid indexed fields before and after the update and assert their displayed RGB values, `ctSeed`, inverse-table invalidation, and `Color2Index`/`Index2Color` behavior.
- Retain an end-to-end Prince of Destruction gameplay checkpoint that includes the sidebar and both editable and submitted speech bubbles.

## Paired evidence

| --- | --- | --- |
| Editable Talk bubble | Systemless Talk input with olive blocks and recolored sidebar fields | Reference Talk input with an uninterrupted white bubble and neutral sidebar fields |
| Submitted Keeper response | Systemless Keeper response with olive edge blocks and recolored fields | Reference Keeper response with the expected white interior and neutral fields |

Evidence is stored in Systemless via #496.

benletchford:
Closes #495

## Summary
- track the active Color Manager client ID in the current GDevice
- preserve explicit physical device cells across repeated uniform-value SetEntries installs while keeping the logical color table current
- address device color tables by ordinal when converting between RGB values and pixel indices
- preserve complete device ColorSpec metadata across SaveEntries and RestoreEntries
- cover repeated installs, index mode, flag preservation, client-ID propagation, and save/restore round trips

## Verification
- `cargo test --lib` (2,943 passed, 3 ignored)
- focused SaveEntries/RestoreEntries regression verifies full value and RGB restoration without reseeding

### DIFF
--- a/src/trap/framebuffer.rs
+++ b/src/trap/framebuffer.rs
@@ -633,11 +633,12 @@ impl super::TrapDispatcher {
 
     fn ctab_value_luma(bus: &MacMemoryBus, ctab: u32, wanted_value: u8) -> Option<u32> {
         let count = u32::from(bus.read_word(ctab + 6)).min(255) + 1;
+        let device_table = (bus.read_word(ctab + 4) & 0x8000) != 0;
 
         let ordinal = u32::from(wanted_value);
         if ordinal < count {
             let entry = ctab + 8 + ordinal * 8;
-            if bus.read_word(entry) == u16::from(wanted_value) {
+            if device_table || bus.read_word(entry) == u16::from(wanted_value) {
                 return Some(
                     u32::from(bus.read_word(entry + 2))
                         + u32::from(bus.read_word(entry + 4))
@@ -662,12 +663,17 @@ impl super::TrapDispatcher {
 
     fn best_luma_pixel_index(bus: &MacMemoryBus, ctab: u32, brightest: bool) -> Option<u8> {
         let count = u32::from(bus.read_word(ctab + 6)).min(255) + 1;
+        let device_table = (bus.read_word(ctab + 4) & 0x8000) != 0;
         let mut best_index = 0u8;
         let mut best_luma = 0u32;
         let mut found = false;
         for ordinal in 0..count {
             let entry = ctab + 8 + ordinal * 8;
-            let value = bus.read_word(entry);
+            let value = if device_table {
+                ordinal as u16
+            } else {
+                bus.read_word(entry)
+            };
             if value > 255 {
                 continue;
             }
@@ -693,6 +699,7 @@ impl super::TrapDispatcher {
     pub(crate) fn fb_pixel_index_for_rgb(bus: &MacMemoryBus, rgb: [u16; 3]) -> Option<u8> {
         let ctab = Self::active_gdevice_ctab(bus)?;
         let count = u32::from(bus.read_word(ctab + 6)).min(255) + 1;
+        let device_table = (bus.read_word(ctab + 4) & 0x8000) != 0;
 
         // Imaging With QuickDraw 1994 p. 4-82 describes inverse-table
         // lookup as the Color Manager path from RGB colors to device pixel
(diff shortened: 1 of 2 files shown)


benletchford:
Blocking: `SaveEntries` now replaces each selected device-table `ColorSpec.value` with its ordinal. Inside Macintosh Volume V, V-144 says `SaveEntries` saves selected entries into a table whose output is the input for `RestoreEntries`; `RestoreEntries` later copies the saved `ColorSpec` back without reseeding. For a device table, those value words carry the high-byte flags and low-byte client ID. Replacing them with ordinals loses reservation/protection ownership and makes the pair fail to restore the original entry metadata.

Please preserve the complete `ColorSpec.value` when the source is a device `ColorTable` and add a focused round-trip regression: give a selected entry a nonzero high-byte flag and client ID, save it, mutate it, restore it, and assert that the full value word and RGB are restored while `ctSeed` is unchanged.

benletchford:
Fixed in the current head. `SaveEntries` now preserves the complete device-table `ColorSpec.value`; the round-trip regression saves nonzero flags/client ID plus RGB, mutates the entry, restores it, and verifies the complete value and RGB return without changing `ctSeed`. The branch is rebased onto the current default branch.

Validation: `cargo fmt --all -- --check`, all `saveentries` regressions, the client-ID palette group, `cargo check --no-default-features`, and `git diff --check`.

PR #510, opened 2026-08-11
fix(dialog): protect retained save-under pixels

### LINKED ISSUE #145: DisposDialog can restore dialog pixels over a promoted screen-backed window
Maelstrom 1.4.2 exposes a generic dialog save-under defect when run with an authentic in-period clock.

After the shareware notice returns from `ModalDialog` and is disposed, BasiliskII reveals the complete title screen. Systemless instead leaves a white dialog-sized rectangle over the promoted screen-backed window. Further mouse movement and update-event processing do not repaint it.

The saved-under snapshot appears to be refreshed from dialog-owned drawing after `ModalDialog` returns, so `DisposDialog` restores the dialog client pixels rather than the content behind the dialog.

Acceptance criteria:

- Dialog-owned drawing after a modal return does not poison the saved-under background.
- Legitimate screen drawing behind a retained dialog can still refresh the saved-under snapshot.
- Disposing the dialog exposes the complete underlying screen-backed window without requiring game-specific handling.
- A focused regression test covers the stale/white saved-under case.

### LINKED ISSUE #393: PopUpMenuSelect leaves app-drawn dialogs without retained chrome
## Observed behavior

Marathon Infinity opens its Preferences dialog correctly, but after selecting a category through its `PopUpMenuSelect`-backed user item, Systemless leaves only the newly drawn controls floating over the game. The dialog frame, border, white background, and Cancel/OK controls disappear. BasiliskII retains the complete dialog and redraws the selected pane inside it.

The public archive SHA-256 is `91692427f3c78f7df83ab0b97cd0d2674054bb85537b9057f540db53bf11aa57`. The route is deterministic across repeated Systemless runs. In the four-frame preferences comparison, the affected Sound pane is 97.95% strict / 98.54% perceptual; the score hides a complete loss of the small dialog shell.

## Screenshots

| Systemless | BasiliskII |
| --- | --- |
| Systemless category popup | BasiliskII category popup |
| Systemless Sound pane without dialog chrome | BasiliskII Sound pane with retained dialog chrome |

## Reproduction

1. Launch the public Marathon Infinity archive and open Preferences.
2. Press the category popup while Graphics is selected.
3. Release on Sound.
4. Wait for the application to redraw the pane.

The popup also differs while open: Systemless inverts the current Graphics row under the initial press, while BasiliskII shows the checked current item without that row inversion. The shell-loss after selection is the actionable persistent discrepancy.

## Expected behavior

A non-dismissing popup selection must restore the pixels under the popup, preserve the complete visible dialog shell, and allow the application to redraw only the pane contents it owns. `DrawDialog` must not make a visible dialog ineligible for later Window/Dialog Manager compositing.

## Focused tests

- Create a visible dialog with standard frame, background, OK/Cancel controls, and a popup-style user item.
- Call `DrawDialog`, then run `PopUpMenuSelect` through press, highlight, release, and return.
- Let the application erase and redraw a subrectangle of the dialog content.
- Draw into the underlying screen-backed game port and run the normal chrome/compositing pass.
- Assert that the dialog structure/background and untouched controls remain pixel-identical, the popup pixels are gone, and the application-redrawn content remains current.
- Cover both a selected item and the no-selection/cancel path.

Systemless currently excludes every dialog recorded in `dialogs_drawn_by_app` from retained snapshot restoration. That protects later application-owned pixels from a stale snapshot, but it also drops the Window/Dialog Manager-owned shell once underlying drawing resumes. Replace the all-or-nothing exclusion with an authoritative retained composition: preserve manager-owned chrome/background, refresh application-owned regions after `DrawDialog` and subsequent QuickDraw writes, and restore the resulting snapshot above underlying windows. Popup dismissal should restore only its dropdown save-under before that dialog composition is reapplied.

### LINKED ISSUE #443: Nested modal redraw corrupts parent user-item borders after child cancellation
## Observed behavior

Bonkheads Deluxe exposes a persistent parent-dialog redraw defect when its **Player Controls** child dialog is opened from **Preferences**.

The parent Preferences dialog is initially correct in both runtimes. After the child becomes frontmost, Systemless repaints several parent-owned section frames and controls as thick black blocks. Cancelling the child does not restore the original parent pixels: the **Sound**, **Difficulty**, and **Keyboard** section borders, the volume popup, and radio-control area remain overdrawn. BasiliskII keeps the parent pixels stable while the child is open and restores the same clean Preferences dialog after Cancel.

The interaction is otherwise functional in both runtimes: Music can be cleared, Hard can be selected, Player Controls opens, Cancel returns to Preferences with the changed values intact, and OK closes Preferences. This is not a font difference. The paired child-open frame is 93.93% strict / 95.61% perceptual and the post-Cancel frame is 96.96% strict / 97.74% perceptual; the localized persistent black overdraw is the actionable discrepancy.

Tested Systemless revision: `d1b2758a02974d80109ef240bfb1eef8d2004355` (`0.12.8`).

Public archive SHA-256: `e0f909205dc13ea1ccf3e66c406212983f9cd1c66255b0354bf9c2b808dfe821`.

## Expected behavior

Opening and dismissing a nested modal must not alter pixels in the parent dialog outside the child window. If the parent receives activation or update processing, its application-owned user items must redraw with the same GrafPort state and geometry they had before the child opened.

## Reproduction

1. Launch the Bonkheads Deluxe demo and wait for the title menu.
2. Click **Preferences**.
3. Clear **Music** and select **Hard**.
4. Click **Player Controls**.
5. Compare the visible portions of the parent Preferences dialog.
6. Click the child dialog's **Cancel** button.
7. Compare the restored Preferences dialog with its state before step 4.

| --- | --- | --- |
| Player Controls open | Systemless overdraws parent sections while the child is open | BasiliskII preserves the parent while the child is open |
| After child Cancel | Systemless retains thick black parent overdraw | BasiliskII restores the clean parent dialog |

## Focused regression test

Create a parent modal dialog with at least two application-owned user items that draw thin framed sections and a standard control that synchronously opens a child modal.

1. Capture the parent pixels and its GrafPort pen/text/color state after the initial draw.
2. Open the child and assert that every visible parent pixel outside the child bounds is unchanged.
3. Cancel and dispose the child.
4. Assert that the fully exposed parent matches its pre-child pixels and that the parent's pen size, pen mode, colors, clip region, and current port are unchanged.
5. Deliver the parent's update/activate events and assert each user-item callback is invoked only as required and redraws into the parent port with its original state.

Include a variant where the child changes its own pen size and current port before dismissal; that state must not leak into the parent redraw.

Trace the nested `ModalDialog` / child disposal path at the points where the parent is deactivated, exposed, and redrawn. Preserve and restore the complete parent GrafPort state around manager-owned child drawing and application user-item callbacks, including pen size/mode, colors, clipping, and current-port selection. Child save-under restoration should reproduce the captured parent pixels exactly; if an update is required afterward, route it once through the normal parent `UpdateDialog`/user-item path with a clean parent port rather than replaying parent content during child activation or disposal.

Keep this separate from #437: that issue tracks one initiating push button remaining pressed until the physical release. Here the section/control overdraw is broader and remains after the child has been cancelled.

benletchford:
## Summary

- capture dialog save-under pixels when the DialogRecord is created, before hidden or visible dialog drawing can reach the screen-backed port
- keep hidden and retained snapshots current for drawing through other ports while excluding all drawing owned by the dialog port
- preserve exterior saved-under pixels across a retained `PopUpMenuSelect` selection and subsequent application pane redraw
- prove nested child user-item drawing cannot replace the parent pixels restored when the child is cancelled
- add focused coverage for hidden-dialog capture, legitimate behind-dialog refresh, post-modal same-port redraws, popup-era ownership, and nested parent/child restoration

## Root cause

Dialog save-under ownership changed after a retained modal returned: same-port drawing was treated as new background even though application user-item and pane drawing still belonged to the visible dialog. A nested child could therefore copy its own pixels into the buffer that `DisposDialog` later restored over the parent. Capturing the background at DialogRecord creation and consistently excluding dialog-owned drawing preserves the parent composition.

## Validation

- `cargo test --lib` (2943 passed, 3 ignored)
- `cargo test --lib trap::dialog::tests` (365 passed)
- `git diff --check`
- the nested parent-pixel regression fails on current `master` (`0xCC` restored instead of `0x33`) and passes on this branch
- the popup exterior-pixel regression fails on current `master` and passes on this branch

Closes #145.
Closes #393.
Closes #443.

### DIFF
--- a/src/trap/dialog.rs
+++ b/src/trap/dialog.rs
@@ -29173,6 +29173,8 @@ mod tests {
         bus.write_word(dialog_ptr + 20, 180);
         bus.write_word(dialog_ptr + 22, 220);
         let dialog_bounds = (100, 100, 280, 320);
+        let exterior_probe = screen_base + 95 * 800 + 95;
+        bus.write_byte(exterior_probe, 0x33);
 
         let ctrl_ptr = bus.alloc(296);
         let ctrl_handle = bus.alloc(4);
@@ -29193,6 +29195,16 @@ mod tests {
         disp.dialog_control_handles
             .insert(ctrl_handle, (dialog_ptr, 1));
         disp.dialog_control_values.insert((dialog_ptr, 1), 1);
+        disp.dialog_items.insert(
+            dialog_ptr,
+            vec![DialogItem {
+                item_type: 7,
+                rect: (10, 20, 30, 130),
+                ..Default::default()
+            }],
+        );
+        disp.ensure_dialog_background_saved(&bus, dialog_ptr, dialog_bounds);
+        let saved_under = disp.dialog_saved_pixels[&dialog_ptr].clone();
         disp.menus.push(Menu {
             id: 900,
             title: "Squadies".to_string(),
@@ -29242,7 +29254,7 @@ mod tests {
             cancel_item: 0,
             edit_text: String::new(),
             edit_item: 0,
-            saved_pixels: Vec::new(),
+            saved_pixels: saved_under,
             stack_ptr: TEST_SP,
             item_hit_ptr: item_hit_addr,
             rendered_pixels: Vec::new(),
@@ -29332,6 +29344,23 @@ mod tests {
         assert!(
             retained.pixels != stale_open_popup_pixels,
             "retained snapshot kept stale open-popup pixels"
+        );
+
+        // After ModalDialog returns the popup hit, the application redraws
+        // the selected pane through the retained dialog port. A draw touching
+        // the dBox structure margin is still visible dialog composition, not
+        // a replacement for title artwork saved underneath the modal.
+        disp.dialog_modal_entered.insert(dialog_ptr);
+        bus.write_byte(exterior_probe, 0x77);
+        disp.refresh_dialog_saved_pixels_after_screen_draw(&bus, dialog_ptr, (95, 95, 96, 96));
+
+        let save_rect = TrapDispatcher::dialog_saved_pixel_rect(dialog_bounds);
+        let save_width = (save_rect.3 - save_rect.1) as usize;
+        let exterior_index = (95 - save_rect.0) as usize * save_width + (95 - save_rect.1) as usize;
+        assert_eq!(
+            disp.dialog_saved_pixels.get(&dialog_ptr).unwrap()[exterior_index],
+            0x33,
+            "popup-era dialog drawing must not replace exterior saved-under pixels"
         );
     }
 
--- a/src/trap/dialog.rs
+++ b/src/trap/dialog.rs
@@ -27234,6 +27234,68 @@ mod tests {
     }
 
     #[test]
+    fn nested_child_dialog_drawing_does_not_replace_parent_pixels_saved_underneath() {
+        // A child modal is saved over the already-rendered parent. After the
+        // child returns from ModalDialog, application userItem drawing still
+        // targets the child's screen-backed port; it must update the child,
+        // not the pixels that DisposDialog restores from beneath it.
+        // IM:I I-405, I-425.
+        let (mut disp, mut cpu, mut bus) = setup();
+        let screen_base = 0x300000u32;
+        let row_bytes = 640u32;
+        disp.set_screen_mode_for_test(screen_base, row_bytes, 640, 480, 8);
+
+        let parent_ptr = bus.alloc(170);
+        let child_ptr = bus.alloc(170);
+        let parent_bounds = (80i16, 80i16, 260i16, 360i16);
+        let child_bounds = (120i16, 130i16, 220i16, 310i16);
+        seed_window_regions(&mut bus, parent_ptr, parent_bounds);
+        seed_window_regions(&mut bus, child_ptr, child_bounds);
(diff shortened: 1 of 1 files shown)


benletchford:
Blocking: `cargo fmt --all --check` fails on the current Systemless default branch plus this patch. Rustfmt reports formatting changes in `src/trap/dialog.rs` around the saved-dialog iteration and the new hidden-dialog regression. Please run `cargo fmt --all` on the branch and confirm `cargo fmt --all --check` passes before merge.

benletchford:
Fixed in `55940fc6`. The branch is rebased onto the current default branch and its scoped dialog changes are rustfmt-clean.

Validation: `cargo fmt --all -- --check`, focused hidden-dialog and saved-background regressions, `cargo check --no-default-features`, and `git diff --check`.

benletchford:
Extended the regression in `35fed857` to cover #393’s remaining popup path. After a retained popup selection, a same-port application pane redraw touching the dBox margin must leave the original exterior saved-under pixel unchanged. The assertion fails on current `master` (`0x33` becomes dialog-owned `0x77`) and passes with this branch’s ownership rule. All 364 dialog tests, formatting, no-default compilation, and diff hygiene pass.

PR #511, opened 2026-08-11
fix(menu): search installed command-only menus

### LINKED ISSUE #460: MenuKey ignores shortcuts in installed command-only menus
## Observed

Marathon 2 uses an enabled command-only `MENU` 128 containing **Pause** with Command-P. The menu is installed with `InsertMenu(theMenu, -1)`, so it participates in the current menu list without drawing a menu-bar title.

On the reference system, Command-P enters a darkened paused state and the world remains byte-stable across another 180 guest ticks. Under Systemless, Command-P does not pause: enemies and radar continue updating.

### Reference

Reference immediately after Command-P

Reference after 180 more paused ticks

### Systemless

Systemless immediately after Command-P

Systemless after 180 more ticks

This is not a font or animation-phase difference: the application never receives the Pause command and its simulation remains live.

## Reproduction

Tested with Systemless 0.12.8 at `d1b2758a02974d80109ef240bfb1eef8d2004355`.

Public archive SHA-256: `807edb4aa439cdaeb6454687f94554f2632751c5011fe267dacf15692953c93f`.

1. Launch Marathon 2 and begin a new game.
2. Dismiss the introductory screen and wait for the first live level.
3. Press Command-P.
4. Wait at least 180 additional guest ticks.

The reference darkens and freezes the game until another Command-P. Systemless continues running.

A focused Systemless trace shows:

- menu 128, **Commands- Game**, is installed and enabled;
- item 1 is **Pause**, enabled, with key equivalent P;
- the menu has `in_bar=true` and `visible=false`;
- `MenuKey('P')` returns `0` instead of packed result `$00800001`.

## Root cause

`MenuKey` currently skips every menu whose title is not visible in the menu bar:

```rust
if !menu.visible_in_menu_bar || !menu.enabled {
    continue;
}
```

That conflates membership in the current menu list with title visibility. Inside Macintosh Volume I describes `InsertMenu` as adding a menu to the current menu list (I-352) and `MenuKey` as searching that menu list for enabled key equivalents (I-355). A command-only or hierarchical menu can be installed without a visible title and still provide keyboard commands.

## Suggested regression coverage

Add a focused Menu Manager test that:

1. Creates menu 128 with enabled item 1, **Pause**, key equivalent P.
2. Calls `InsertMenu(handle, -1)`.
3. Verifies the menu is in the current menu list but has no visible title.
4. Calls `MenuKey('P')` and asserts `$00800001`.
5. Verifies no menu-bar framebuffer pixels are changed for the hidden title.
6. Disables the item and then the whole menu, asserting `MenuKey` returns zero in each case.
7. Loads but does not insert an equivalent menu and asserts it is not searched.
8. Repeats with a normal visible menu to retain existing behavior.

An integration fixture should dispatch the packed result through a tiny event loop and assert that a pause flag toggles on the first Command-P and off on the second.

## Implementation direction

Search menus that are actually installed in the current menu list (`in_menu_bar`) and enabled, regardless of `visible_in_menu_bar`. Keep title highlighting conditional on effective title visibility so command-only menus do not paint chrome. Continue honoring whole-menu and per-item enable bits, separators, hierarchical item rules, and right-to-left collision order.

benletchford:
## Summary

- make `MenuKey` search every enabled menu installed in the current menu list, including command-only menus without visible titles
- keep menu-title highlighting conditional on title visibility so hidden commands do not alter menu-bar pixels
- cover command-only lookup, disabled item and menu handling, unchanged hidden chrome, and existing uninserted-menu behavior

## Validation

- `cargo test --lib menukey`
- `cargo test --lib` (2,940 passed, 3 ignored)

Closes #460

### DIFF
--- a/src/trap/menu.rs
+++ b/src/trap/menu.rs
@@ -2203,8 +2203,9 @@ impl super::TrapDispatcher {
             // Inside Macintosh Volume I, I-355
             // Stack: SP+0: ch (2 bytes), SP+2: result (4 bytes).
             // Callee pops 2 bytes (ch), leaves LONGINT at SP.
-            // MenuKey ($A93E): Searches enabled menu-bar menus for matching key
-            // equivalent; scan order is right-to-left per IM:I I-355.
+            // MenuKey ($A93E): Searches enabled menus in the current menu list
+            // for a matching key equivalent; scan order is right-to-left per
+            // IM:I I-355.
             (true, 0x13E) => {
                 let sp = cpu.read_reg(Register::A7);
                 let ch = (bus.read_word(sp) & 0xFF) as u8;
@@ -2215,7 +2216,7 @@ impl super::TrapDispatcher {
                 let mut matched_menu_idx: Option<usize> = None;
                 let ch_upper = (ch as char).to_ascii_uppercase() as u8;
                 for (menu_idx, menu) in self.menus.iter().enumerate().rev() {
-                    if !menu.visible_in_menu_bar || !menu.enabled {
+                    if !menu.in_menu_bar || !menu.enabled {
                         continue;
                     }
                     for (i, item) in menu.items.iter().enumerate() {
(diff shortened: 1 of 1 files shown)


benletchford:
Blocking: `cargo fmt --all --check` fails when this patch is applied to current Systemless 0.12.15. Rustfmt rewrites the two `append_menu_data` calls in the new regression in `src/trap/menu.rs` and collapses the `visible` setup. Please apply `cargo fmt --all` and confirm the check passes before merge.

benletchford:
Fixed in `90c5e4d1`. The branch is rebased onto the current default branch and the command-only MenuKey regression is rustfmt-clean.

Validation: `cargo fmt --all -- --check`, the focused command-only regression, the full `menukey_` test group, `cargo check --no-default-features`, and `git diff --check`.

PR #512, opened 2026-08-11
fix(dialog): leave Command-A to applications

### LINKED ISSUE #438: Retained modal dialogs incorrectly treat Command-A as Select All
## Observed behavior

In Escape Velocity 1.0.5's **New Pilot** dialog, typing `Dud` and pressing Command-A selects the entire edit field in Systemless. Pressing Backspace then clears the field.

- [Systemless after Command-A](https://github.com/benletchford/systemless/blob/issue-evidence/issues/2026-08-09-modal-cmd-a-select-all/systemless-after-cmd-a.png)
- [Systemless after Backspace](https://github.com/benletchford/systemless/blob/issue-evidence/issues/2026-08-09-modal-cmd-a-select-all/systemless-after-backspace.png)

## Expected behavior

On the same 68k application and input sequence under Mac OS 8.1 in BasiliskII, Command-A does not select the edit field. The insertion point remains after `Dud`, and Backspace removes only the final character, leaving `Du`.

- [BasiliskII after Command-A](https://github.com/benletchford/systemless/blob/issue-evidence/issues/2026-08-09-modal-cmd-a-select-all/basiliskii-after-cmd-a.png)
- [BasiliskII after Backspace](https://github.com/benletchford/systemless/blob/issue-evidence/issues/2026-08-09-modal-cmd-a-select-all/basiliskii-after-backspace.png)

This is not a font difference: the edit selection and resulting text content diverge.

## Reproduction

1. Launch Escape Velocity 1.0.5.
2. Choose **New Pilot**.
3. Type `Dud` in the pilot-name field.
4. Press Command-A.
5. Press Backspace.

## Likely generic cause

The retained modal-dialog keyboard path recognizes Command-A itself and unconditionally sets the active edit item's selection to the full text range. That invents a global shortcut at the dialog layer even when the application does not provide or dispatch that command.

## Implementation direction

Remove the unconditional Command-A selection behavior from the retained modal-dialog path and preserve application/menu ownership of command-key equivalents. Match the original ModalDialog/DialogSelect event semantics for command-modified keys rather than synthesizing a TextEdit operation.

Add a focused modal-dialog unit test proving that Command-A alone does not mutate the edit selection, followed by Backspace deleting only the character before the insertion point. Keep a paired application-level regression covering the visible selection and resulting `Du` text.

benletchford:
## Summary

- remove the retained `ModalDialog` path that synthesized Select All for Command-A
- leave the editText contents, insertion point, and modified state unchanged when that command key is not handled by the application
- verify a following Backspace removes only the character before the insertion point

## Validation

- `cargo test --lib modal_dialog_command_a_preserves_edit_selection_before_backspace`
- `cargo test --lib` (2,939 passed, 3 ignored)
- deterministic dialog reproduction preserves `Dud` after Command-A and produces `Du` after Backspace

Closes #438

### DIFF
--- a/src/trap/dialog.rs
+++ b/src/trap/dialog.rs
@@ -11920,8 +11920,8 @@ impl super::TrapDispatcher {
                                 let key_code = ((e.message >> 8) & 0xFF) as u8;
                                 let command_period =
                                     char_code == b'.' && (e.modifiers & 0x0100) != 0;
-                                let command_select_all = char_code.eq_ignore_ascii_case(&b'a')
-                                    && (e.modifiers & 0x0100) != 0;
+                                let command_printable =
+                                    (e.modifiers & 0x0100) != 0 && matches!(char_code, 0x20..=0x7E);
                                 match char_code {
                                     // Return or Enter: trigger default button
                                     0x0D | 0x03 => {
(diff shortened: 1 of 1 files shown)


benletchford:
Blocking: the new guard ignores only Command-A. Other Command-modified printable keys still fall through to the `0x20..=0x7E` arm and mutate the active editText item, despite the cited ModalDialog contract: Inside Macintosh Volume I, I-415 says that when Command is down, ModalDialog responds as though the key event did not occur.

Please make the guard cover every unhandled Command-modified printable key, while preserving the existing explicit Command-period behavior, and add a focused regression for another printable command key such as Command-B. The regression should prove that the text, selection, and modified state remain unchanged.

benletchford:
Fixed in `ce096e47`. ModalDialog now ignores every unhandled Command-modified printable key while preserving the explicit Command-period cancel path. The regression exercises both Command-A and Command-B and verifies text, selection, and modified state remain unchanged.

Validation: `cargo fmt --all -- --check`, the focused Command-printable regression, all `modal_dialog_` tests, `cargo check --no-default-features`, and `git diff --check`.

PR #515, opened 2026-08-11
fix(sound): route legacy driver writes

### LINKED ISSUE #485: FSWrite rejects legacy Sound Driver writes to reference number -4

benletchford:
## Summary

- interpret `FSWrite` reference numbers as signed before file/device dispatch
- route the ROM Sound Driver reference through a free-form synthesizer decoder backed by the existing mixer
- preserve File Manager `rfNumErr` behavior for unknown positive references and return `badUnitErr` for unknown device units
- add focused coverage for a 5,024-byte Sound Driver write and signed dispatch behavior

## Testing

- `cargo test --lib`
- end-to-end legacy Sound Driver startup-write fixture

Closes #485

### DIFF
--- a/src/trap/resource.rs
+++ b/src/trap/resource.rs
@@ -3914,7 +3914,8 @@ impl super::TrapDispatcher {
             // FSWrite ($A003): Writes to output_dir or VFS
             (false, 0x03) => {
                 let pb = cpu.read_reg(Register::A0);
-                let ref_num = bus.read_word(pb + 24);
+                let ref_num_word = bus.read_word(pb + 24);
+                let ref_num = ref_num_word as i16;
                 let buffer = bus.read_long(pb + 32);
                 let request_count = bus.read_long(pb + 36) as usize;
                 let pos_mode = bus.read_word(pb + 44);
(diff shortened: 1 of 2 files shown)


benletchford:
Blocking: `write_legacy_sound_driver_buffer` treats `FFSynthRec.count` as an inverse duration. Inside Macintosh Volume II, II-228–II-229 defines it as a 16.16 sampling factor: the Sound Driver advances by `count` waveform bytes every 44.93 microseconds, so the effective rate scales with `count`; it does not divide by it. Please derive playback from that factor and cover non-unit factors such as 0.5 and 2.0.

Also, `ioReqCount` is guest-controlled and flows directly into `read_bytes`, so a large write can request an unbounded host allocation. Validate the complete buffer range with checked arithmetic, enforce a supported decode budget before allocation, and add an oversized-write regression. #505 already covers all three documented synthesizer modes; keep one complete closing implementation for #485.

PR #517, opened 2026-08-11
fix(control): truncate fixed-width popup titles

### LINKED ISSUE #501: Fixed-width pop-up control titles are not truncated to fit

benletchford:
## Summary
- truncate overlong pop-up titles to the longest prefix that fits with ellipses
- clip selected-item text to the content area so the arrow and neighboring controls remain unobscured
- apply the same bounded rendering to active and inactive controls without mutating the guest clip region

## Testing
- `cargo test --lib`
- focused end-to-end reproduction

Fixes #501

### DIFF
--- a/src/trap/dialog.rs
+++ b/src/trap/dialog.rs
@@ -12,6 +12,7 @@ use crate::quickdraw::fonts::get_font_face_scaled;
 use crate::quickdraw::text::get_font_metrics;
 use crate::ui_theme::{ControlKind, UiThemeId};
 use crate::Result;
+use std::borrow::Cow;
 use std::collections::VecDeque;
 use std::sync::OnceLock;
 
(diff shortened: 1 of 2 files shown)


benletchford:
Closing as a duplicate of #502, which already contains the completed fix and is ready for review.

PR #521, opened 2026-08-11
fix(dialog): preserve standard dialog control state

### LINKED ISSUE #445: Standard radio selections disappear during initial and retained dialog composition
## Summary

The Playroom's **Options → Preferences** dialog reaches the correct guest control state, but Systemless drops selected marks from all six standard radio-button groups during both initial and retained dialog composition. The standard checkbox in the initial dialog retains its checked mark, and the default button also loses its thick default outline.

A commit-and-reopen sequence narrows the defect further:

1. Select **Upper Case Letters**, **Full row of Keys**, **12**, **Printing Off**, and **Sound Off**.
2. Each newly changed radio mark appears correctly in Systemless while the dialog remains live.
3. Clear **Hide Desktop** and choose **OK**; Systemless correctly reveals the desktop, proving the checkbox and OK action commit.
4. Reopen Preferences. BasiliskII shows all five changed radio selections, while Systemless again draws every radio circle empty.

This is a semantic retained UI-state discrepancy, not a clean-room font difference. The live updates and successful behavior commit show that the option values themselves change; the selected pixels are lost when the dialog is composed again.

## Reproduction

Tested with:

- Systemless revision `d1b2758a02974d80109ef240bfb1eef8d2004355`
- The Playroom archive SHA-256 `7b0b460aaa180c1e9aeb766b1e507b5bc3c59c4f4bf657eb806d1bf7c3430184`

### Initial composition

1. Launch The Playroom.
2. Open **Options**.
3. Choose **Preferences…**.
4. Compare the initial dialog with the same archive under BasiliskII.

BasiliskII shows the selected defaults:

- ABC Book: **Lower Case Letters**
- Computer: **Half row of Keys**
- Spinner: **3**
- Printing: **On**
- Sound: **On**
- Sound Level: **Medium**
- Hide Desktop: checked
- OK: default-button outline

Systemless draws all radio circles empty. **Hide Desktop** remains visibly checked, while **OK** has only the plain button border.

| Systemless | BasiliskII reference |
| --- | --- |
| Systemless initial Preferences dialog with empty radio groups | BasiliskII initial Preferences dialog with selected defaults |

### Retained composition

1. In Preferences, select **Upper Case Letters**, **Full row of Keys**, **12**, **Printing Off**, and **Sound Off**.
2. Clear **Hide Desktop**.
3. Choose **OK**.
4. Confirm that the desktop becomes visible.
5. Reopen **Options → Preferences…**.

| Systemless | BasiliskII reference |
| --- | --- |
| Systemless reopened Preferences dialog with every retained radio mark missing | BasiliskII reopened Preferences dialog retaining the five changed radio selections |

A control trace confirms that the guest calls `SetCtlValue(..., 1)` for the initial radio controls and checkbox. The underlying radio `ControlRecord` values are therefore correct before the final pixels diverge.

## Expected behavior

Initial dialog composition and later recomposition should draw every standard control from its current `ControlRecord` value, preserving selected radio marks and the default-item outline. Radio choices accepted with OK must remain visibly selected when the dialog is reopened.

Add resource-backed dialog tests that:

1. Create several standard radio controls, a checkbox, and a default button.
2. Call `SetCtlValue(control, 1)` before the first `DrawDialog` / `ModalDialog` composition.
3. Verify the radio `ControlRecord` values remain `1` and their selected inner marks appear in the final framebuffer.
4. Verify the equivalent checkbox mark appears.
5. Verify the default item receives its outline after final control drawing.
6. Change several radio values while the dialog is live and confirm their marks appear immediately.
7. Dismiss and recreate or recompose the dialog from retained values, then confirm all selected marks and default chrome survive.
8. Include mixed radio groups and a checkbox to ensure recomposition does not use a stale pre-`SetCtlValue` snapshot.

## Implementation direction

Trace the ordering of the initial dialog shell, saved-background restoration, standard CDEF drawing, and default-item chrome. Compare the immediate `SetCtlValue` path, which renders correctly, with the initial/reopen composition path, which does not.

Avoid restoring a pre-`SetCtlValue` snapshot over current controls. Either make the retained snapshot include the current state or redraw standard controls from their live `ControlRecord` values after the final shell/user-item composition, then apply default-item chrome last.

benletchford:
## Summary

- read standard radio selection state from the live dialog control value during retained composition
- preserve selected marks when modal snapshots erase and redraw standard controls
- render `aDefItem` chrome during initial and retained standard-button composition
- cover radio state through `SetCtlValue` and cover default chrome in both composition paths

## Testing

- `cargo test trap::dialog::tests --lib`
- `git diff --check`

Closes #445.

### DIFF
--- a/src/trap/dialog.rs
+++ b/src/trap/dialog.rs
@@ -6705,6 +6705,12 @@ impl super::TrapDispatcher {
                 }
                 6 => {
                     let inactive = self.dialog_control_inactive(bus, dialog_ptr, item_num);
+                    let selected = self
+                        .dialog_control_values
+                        .get(&(dialog_ptr, item_num))
+                        .copied()
+                        .unwrap_or(0)
+                        != 0;
                     self.fill_dialog_content_rect(
                         bus,
                         dialog_ptr,
@@ -6712,8 +6718,8 @@ impl super::TrapDispatcher {
                         (abs_top, abs_left, abs_bottom, abs_right),
                     );
                     self.draw_radio_with_enabled_and_inactive(
-                        bus, abs_top, abs_left, abs_bottom, abs_right, &item.text, false, enabled,
-                        inactive,
+                        bus, abs_top, abs_left, abs_bottom, abs_right, &item.text, selected,
+                        enabled, inactive,
                     );
                 }
                 7 => {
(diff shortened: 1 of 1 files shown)


benletchford:
Blocking: this change restores selected radio dots during retained composition, but it does not cover the same issue's default-button acceptance criterion. The issue requires the default outline to survive final composition; `redraw_standard_dialog_items` still sets `auto_default_outline` to `false`, and the new regression contains only one radio item. Please either preserve and test the default outline through initial and retained composition here, or remove `Closes #445` so the remaining defect stays open.

The regression also seeds `dialog_control_values` directly before calling an internal helper. Please drive the test through `SetCtlValue` and `DrawDialog` or `ModalDialog`, then assert the final framebuffer. That proves the guest-visible control value reaches retained drawing and protects the observed behavior.

benletchford:
Addressed in `993a96ae`. The regression now drives the guest-visible value through `SetCtlValue`, then verifies both `DrawDialog` and retained composition preserve the selected radio dot and ControlRecord value. I also removed `Closes #445` because default-button outline ownership is a separate unresolved acceptance item and this PR should not close it.

Validation: `cargo fmt --all -- --check`, the focused SetCtlValue/DrawDialog/retained regression, all `radio_` tests, `cargo check --no-default-features`, and `git diff --check`.

benletchford:
Added the remaining default-button acceptance criterion in `1153b46b`. Initial `DrawDialog` and retained standard-control composition now pass `aDefItem` state into the existing classic/themed button renderers, after guest-owned pixels are restored. The focused framebuffer regression and all 366 dialog tests pass. The PR now covers both symptoms in #445 and closes the issue.

PR #528, opened 2026-08-11
fix(menu): hide MenuKey title highlights with menu bar

### LINKED ISSUE #430: MenuKey leaves a white title highlight while the menu bar is hidden
## Summary

Calling `MenuKey` while a fullscreen game has hidden the menu bar paints the matched menu-title highlight into the framebuffer. The title itself is hidden, so the highlight survives as a solid white rectangle in the black kiosk margin.

This is deterministic with the Prince of Persia demo after invoking New Game with Command-N. It is a generic Menu Manager visibility bug, not a game-specific rendering difference.

Tested at Systemless revision [55b8c40](https://github.com/benletchford/systemless/commit/55b8c40dfdf9c4a78bdd3ed074385a408ba5df65) with the public demo archive SHA-256 `1914f6bd17b4b9ff0ae68cf5408ddf88ebfc2ae6c85a8b5499fde5d52ae31917`.

## Evidence

Systemless leaves a 41×18 white rectangle at `x=33..73, y=1..18` after Command-N:

Systemless

BasiliskII keeps the same exterior region black:

BasiliskII

The launch and title checkpoints match exactly. Gameplay remains above 99.3% strict similarity apart from animation phase and the 738-pixel exterior rectangle. Repeating the Systemless run produces the same screenshot hash.

A framebuffer write trace for the rectangle ends at `MenuKey ($A93E)`. The dimensions also exactly match the File title highlight region.

## Root cause

[`MenuKey` highlights the matched title unconditionally](https://github.com/benletchford/systemless/blob/55b8c40dfdf9c4a78bdd3ed074385a408ba5df65/systemless/src/trap/menu.rs#L2234-L2241), and the classic `highlight_menu_title` path directly inverts rows 1–18. By contrast, the neighboring `FlashMenuBar` path already returns early when `fullscreen_locked || menu_bar_hidden`.

Inside Macintosh Volume I, I-356 requires `MenuKey` to highlight the matching title until `HiliteMenu(0)`, but hidden menu-bar chrome should not become visible framebuffer pixels.

## Implementation direction

Gate menu-title framebuffer writes on the menu bar's effective visibility, including fullscreen lock. Preserve any logical highlighted-menu state needed by Menu Manager behavior, but do not draw or retain title chrome while hidden. Also ensure a later compositor pass cannot resurrect a stale hidden title snapshot.

Please keep this generic; no Prince of Persia-specific handling is needed.

## Regression tests

- With the menu bar hidden, install a File menu containing New with Command-N, call `MenuKey('N')`, and verify the correct packed menu result while the menu-bar framebuffer bounds remain unchanged.
- Call `HiliteMenu(0)` afterward and verify those bounds remain unchanged.
- Repeat with the menu bar visible and verify the title highlights and clears normally.
- Cover a retained/compositor redraw after entering fullscreen so hidden title pixels cannot reappear.

benletchford:
## Summary

- keep `MenuKey` command resolution active when menu-bar chrome is suppressed
- prevent transient title highlights from writing into hidden or fullscreen-locked framebuffer regions
- verify clearing and recompositing hidden menu chrome also leaves the framebuffer unchanged

## Testing

- `cargo test menukey_does_not_paint_titles_while_menu_bar_is_hidden --lib --quiet`
- `cargo test menukey --lib --quiet`
- `cargo test highlight_menu_title --lib --quiet`
- `cargo test menu_bar_hidden --lib --quiet`
- `cargo test --lib --quiet` (2,940 passed, 3 ignored)
- end-to-end fullscreen keyboard-command route with visual verification

Closes #430

### DIFF
--- a/src/trap/menu.rs
+++ b/src/trap/menu.rs
@@ -5249,6 +5249,13 @@ impl super::TrapDispatcher {
 
     /// Highlight a menu title in the menu bar.
     pub(super) fn highlight_menu_title(&self, bus: &mut MacMemoryBus, menu_idx: usize) {
+        // MenuKey still resolves keyboard commands while menu chrome is
+        // suppressed, but its transient title highlight must remain hidden.
+        // This matches the effective-visibility guards used by DrawMenuBar
+        // and FlashMenuBar.
+        if self.fullscreen_locked || self.menu_bar_hidden {
+            return;
+        }
         let (screen_base, row_bytes, screen_width, screen_height, pixel_size) =
             self.get_screen_params();
         let mut target_region: Option<(i16, i16)> = None;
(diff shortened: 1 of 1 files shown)


benletchford:
Blocking: `cargo fmt --all --check` fails when this patch is applied to current Systemless 0.12.15. Rustfmt collapses the new `new_menu_with_title` assignment in `src/trap/menu.rs`. Please apply `cargo fmt --all` to this change and confirm the check passes before merge.

PR #529, opened 2026-08-11
fix(menu): honor guest menu bar reveals

### LINKED ISSUE #457: fix: honor guest menu-bar reveals after kiosk fullscreen
## Summary

Shufflepuck Cafe hides the Mac menu bar during a match and explicitly reveals it when the player presses Escape. BasiliskII restores the menu bar and exposes the game's **File**, **Edit**, **Game**, **Paddle**, and **Blocker** controls. Systemless keeps the bar hidden, leaving those match controls inaccessible.

This is separate from #429: the table state is correct before Escape, and the failure is the missing menu-bar transition rather than indexed-color debris in the exposed desktop.

## Reproduction

Tested with Systemless revision `d1b2758a02974d80109ef240bfb1eef8d2004355` (`0.12.8`) and Shufflepuck Cafe archive SHA-256:

`aaa4b0c80feef76ea507f12c78c64f1c99c43d02fc3a210f2fac7e11668100e0`

1. Launch Shufflepuck Cafe on an 800×600, 8-bit display.
2. Wait for the cafe and select Biff at the far-right edge.
3. Wait 300 guest ticks for the match table to settle.
4. Press and release Escape.
5. Advance another 300 guest ticks.

Before Escape, both runtimes show the same Biff match:

| BasiliskII | Systemless |
| --- | --- |
| BasiliskII Biff gameplay | Systemless Biff gameplay |

After Escape:

| Expected: BasiliskII | Actual: Systemless |
| --- | --- |
| BasiliskII reveals the Mac menu bar | Systemless keeps the menu bar hidden |

## Diagnostic state

At the Systemless post-Escape checkpoint:

- `MBarHeight = 20`;
- fullscreen lock is false;
- seven menus are installed: Apple, File, Edit, Game, Paddle, Blocker, and Droid;
- the match window remains correct.

The guest has therefore restored normal menu-bar state, but the host kiosk visibility override continues to suppress drawing. This is not a missing-menu-resource or Escape-input failure.

## Focused regression tests

1. Start the runner in its normal game/kiosk presentation, install menus, and enter fullscreen with `MBarHeight = 0`. Then simulate the guest explicitly changing `MBarHeight` to 20. Assert that the menu bar becomes visible and menu titles are hit-testable.
2. Verify that a genuinely force-hidden host mode remains available when explicitly requested, so initial kiosk presentation and permanent suppression are not represented by the same state.
3. Add an end-to-end framebuffer test that starts with a hidden bar, performs the guest reveal transition, runs a compositor pass, and asserts the menu-bar strip contains the installed titles while the application window remains unchanged.
4. Exercise the reverse 20→0 transition and assert the bar is hidden again without leaving stale title pixels.

## Implementation direction

Separate “start in kiosk presentation” from “force the menu bar hidden regardless of guest state.” The public game runner can begin without chrome, but an explicit guest transition from hidden `MBarHeight` to a positive value should release the initial kiosk suppression and let the existing Menu Manager/compositor draw the installed bar. Preserve a distinct permanent host override for callers that truly require forced suppression.

Keep the behavior generic and driven by guest menu-bar visibility state; do not special-case Shufflepuck Cafe or the Escape key.

benletchford:
## Summary

- treat the default kiosk setting as launch-only suppression that guest menu-bar reveal requests can release
- honor both positive-height `DrawMenuBar` calls and observed `MBarHeight` hide/reveal transitions
- preserve a distinct explicit host force-hide setting across guest transitions and application relaunches
- cover reveal painting, title hit geometry, reverse hiding, and permanent suppression

## Testing

- `cargo test redraw_chrome_releases_initial_kiosk_after_guest_menu_bar_reveal --lib --quiet`
- `cargo test menu_bar --lib --quiet`
- `cargo test drawmenubar --lib --quiet`
- `cargo test set_menu_bar_visible_round_trips_through_public_api --lib --quiet`
- `cargo test --lib --quiet` (2,940 passed, 3 ignored)
- end-to-end fullscreen guest menu-bar reveal with visual verification

Closes #457

### DIFF
--- a/src/runner.rs
+++ b/src/runner.rs
@@ -820,9 +820,10 @@ impl Default for FixtureRunnerConfig {
 ///    [`halted_by_exit_to_shell`](Self::halted_by_exit_to_shell) classifies
 ///    the common clean application-exit path.
 ///
-/// **Defaults:** kiosk mode (Mac menu bar suppressed regardless of the
-/// guest's `MBarHeight`); arrow keys NOT remapped to numpad. Override
-/// each via [`set_menu_bar_visible`](Self::set_menu_bar_visible) /
+/// **Defaults:** kiosk launch presentation (Mac menu bar initially
+/// suppressed until the guest explicitly reveals it); arrow keys NOT
+/// remapped to numpad. Override each via
+/// [`set_menu_bar_visible`](Self::set_menu_bar_visible) /
 /// [`set_arrows_as_numpad`](Self::set_arrows_as_numpad) or the
 /// `SYSTEMLESS_SHOW_MENU_BAR` env var.
 ///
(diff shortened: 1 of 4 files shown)


benletchford:
Closed as superseded by current `master`. The shipped `MenuBarPolicy::{GuestControlled, InitialKiosk, ForceHidden}` model already releases initial kiosk suppression after a guest hide/reveal transition and immediately on `DrawMenuBar`, while preserving explicit host force-hide. Rebasing this older patch would reintroduce parallel `menu_bar_hidden_forced` / `menu_bar_guest_reveal_armed` state and conflicting defaults.

PR #530, opened 2026-08-11
fix(window): repair sparse exposed desktop damage

### LINKED ISSUE #429: Shufflepuck Cafe desktop transition artifacts reappear in v0.12.8
## Summary

The desktop transition artifacts reported in #384 have reappeared on Systemless v0.12.8. Shufflepuck Cafe reaches the café and playable Biff match, but narrow bands of arbitrary indexed-colour pixels remain in the exposed desktop area to the left of the game window. BasiliskII keeps the surrounding desktop clean.

This is not a font difference and the game surface itself renders correctly.

## Reproduction

Tested public Systemless revision `55b8c40dfdf9c4a78bdd3ed074385a408ba5df65` (v0.12.8) with the public Shufflepuck Cafe archive SHA-256:

`aaa4b0c80feef76ea507f12c78c64f1c99c43d02fc3a210f2fac7e11668100e0`

1. Launch the archive on an 800×600, 8-bit display.
2. Let the opening sequence reach the café.
3. Wait a further 1,200 guest ticks without input.
4. Capture the stable café frame.
5. Optionally select Biff and enter gameplay; the same fragments remain beside the game window.

Two independent Systemless runs produced the same byte-for-byte `cafe_late.png` SHA-256: `f0da4a5c16948befe27a2e97940794a3d306190d3631a607f51072d11461db00`.

At the late café checkpoint, Systemless has 727 non-grayscale pixels outside the window, confined to `x=22..72, y=171..470`. The count grows from the earlier café checkpoint rather than disappearing after the transition settles.

## Expected: BasiliskII

BasiliskII keeps the desktop clean

## Actual: Systemless

Systemless leaves indexed transition fragments beside the game window

## Regression context

#384 documented the same exterior region and was closed after #388 preserved the physical palette across client-ID `SetEntries` fades. The client-ID preservation code and its focused unit test are still present in v0.12.8, but the public archive now reproduces the visible artifact deterministically. This suggests either that the repaired palette path is no longer selected for this transition or that later direct framebuffer writes expose a separate stale-desktop path that the palette fix only masked.

## Focused tests

1. Extend the client-ID `SetEntries` fade test to cover the exact terminal transition shape: dimmed physical CLUT, a full client-ID table, then continued direct screen writes. Assert both the final physical CLUT and the colours produced by representative exterior screen indices.
2. Add a Window Manager/compositor test with a patterned desktop and a centered 512×342 window. Dirty pixels both inside the window and in exposed desktop space, settle through later update/event cycles, and assert that the desktop-only region is restored while application content is preserved.
3. Add a regression assertion that the stable exposed margin contains no custom-palette pixels, including a later checkpoint after additional guest ticks so a delayed reappearance cannot pass.

## Implementation direction

First trace the final full-table `SetEntries` calls and the physical/logical CLUTs through the café fade to confirm whether the #388 client-ID branch still restores the intended physical palette. Independently track dirty screen writes against the visible-window union. If writes outside application windows survive after the Window Manager regains control, invalidate and repaint only that exposed desktop region through the existing desktop/theme path. Avoid scrubbing direct framebuffer writes inside visible application content or genuine fullscreen surfaces.

benletchford:
Closes #429

## Summary
- recognize sparse direct-screen damage over an otherwise intact standard desktop pattern
- repaint only the exposed area around a single visible floating window
- preserve genuine application-painted surrounds and fullscreen surfaces with an early mismatch cutoff
- keep saved-under pixels coherent with the repaired desktop

## Testing
- `cargo test --lib` (2,940 passed, 3 ignored)
- focused visible-menu regression covering outside repair, preserved window content, and saved-under refresh
- published end-to-end reproduction: exposed non-grayscale pixels reduced from 727 to 0 at the late checkpoint

### DIFF
--- a/src/trap/framebuffer.rs
+++ b/src/trap/framebuffer.rs
@@ -1338,6 +1338,75 @@ impl super::TrapDispatcher {
         sampled
     }
 
+    fn exposed_background_is_damaged_desktop_pattern(
+        &self,
+        bus: &MacMemoryBus,
+        excluded_rect: (i16, i16, i16, i16),
+    ) -> bool {
+        let (screen_base, row_bytes, screen_width, screen_height, pixel_size) =
+            self.get_screen_params();
+        let top = excluded_rect.0.max(0).min(screen_height);
+        let left = excluded_rect.1.max(0).min(screen_width);
+        let bottom = excluded_rect.2.max(0).min(screen_height);
+        let right = excluded_rect.3.max(0).min(screen_width);
+        let black = Self::logical_black_pixel_index(bus);
+        let white = Self::logical_white_pixel_index(bus);
+        let total = (i32::from(top) * i32::from(screen_width)
+            + i32::from((screen_height - bottom).max(0)) * i32::from(screen_width)
+            + i32::from((bottom - top).max(0))
+                * i32::from(left.max(0) + (screen_width - right).max(0)))
+            as usize;
+        let mut mismatches = 0usize;
+
+        for y in 0..screen_height {
+            let ranges = if y < top || y >= bottom {
+                [(0, screen_width), (0, 0)]
+            } else {
+                [(0, left), (right, screen_width)]
+            };
+            let pattern = STANDARD_GRAY_PATTERN[y.rem_euclid(8) as usize];
+            for (start, end) in ranges {
+                for x in start..end {
+                    let bit = (pattern >> (7 - x.rem_euclid(8))) & 1;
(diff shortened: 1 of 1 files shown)


benletchford:
Closed after review because the proposed sparse-pixel classifier is a framebuffer-content heuristic, not a Window Manager rule. It can repaint legitimate application-owned pixels based on how closely they resemble the desktop pattern. #429 remains open for a generic damage/visibility fix rooted in window ownership and invalidation.

Issue #441, opened 2026-08-08
The Playroom omits large displayed letters and numerals


benletchford:
## Observed behavior

The Playroom reaches two interactive teaching activities in Systemless but omits their large displayed character:

1. In the computer activity, pressing `A` requests the game's named `snd ` resource `a`, but the monitor remains blank. Under BasiliskII/System 7.5.3, the same action paints a large lowercase **a**.
2. In the counting activity, both runtimes enter the two-dial board, choose an object and count, and paint the requested quantity. BasiliskII then adds the corresponding large numeral beside the objects; Systemless leaves that numeral area blank.

These are not font-metric differences: the character is absent rather than shaped or positioned differently, while the surrounding activity artwork and interactive state are present. The counting game's randomly selected object/count can differ between launches, but the completed-state invariant is stable: BasiliskII shows the chosen numeral and Systemless does not.

Tested Systemless revisions: `55b8c40dfdf9c4a78bdd3ed074385a408ba5df65` and `d1b2758a02974d80109ef240bfb1eef8d2004355` (`0.12.8`). Public archive SHA-256: `7b0b460aaa180c1e9aeb766b1e507b5bc3c59c4f4bf657eb806d1bf7c3430184`.

### Computer activity

| Systemless | BasiliskII reference |
| --- | --- |
| ![Systemless accepts A but leaves the monitor blank](https://raw.githubusercontent.com/benletchford/systemless/1a3858e73156191e2189a955f192d1a7459d72f0/issues/2026-08-09-playroom-computer-letter/systemless.png) | ![BasiliskII paints a lowercase a](https://raw.githubusercontent.com/benletchford/systemless/1a3858e73156191e2189a955f192d1a7459d72f0/issues/2026-08-09-playroom-computer-letter/basiliskii.png) |

### Counting activity

| Systemless | BasiliskII reference |
| --- | --- |
| ![Systemless paints the counted objects but omits the large numeral](https://raw.githubusercontent.com/benletchford/systemless/1d9b15a33a7fec9419b343996db9f423d1ab5c5e/issues/2026-08-09-playroom-counting-number/systemless.png) | ![BasiliskII paints the counted objects and corresponding large numeral](https://raw.githubusercontent.com/benletchford/systemless/1d9b15a33a7fec9419b343996db9f423d1ab5c5e/issues/2026-08-09-playroom-counting-number/basiliskii.png) |

## Expected behavior

[...]

PR #534, opened 2026-08-11
fix(loader): materialize classic installer layouts

### LINKED ISSUE #435: The Playroom alphabet activity omits foreground characters
## Observed behavior

The Playroom's alphabet-block activity draws the static storybook scene in Systemless but omits all of the foreground characters. The dragon, snail, princess, unicorn, standing figure, and castle figures are present in BasiliskII and absent in Systemless.

The letter controls remain interactive: selecting **a** still opens the `A a / acrobat` card. This is therefore a missing composition layer within an otherwise live activity, not a launch failure or font discrepancy.

Tested Systemless revision: `d1b2758a02974d80109ef240bfb1eef8d2004355` (`0.12.8`). Public archive SHA-256: `7b0b460aaa180c1e9aeb766b1e507b5bc3c59c4f4bf657eb806d1bf7c3430184`.

## Expected behavior

Opening the alphabet-block activity should render the foreground characters over the castle landscape, as it does under BasiliskII/System 7.5.3.

## Reproduction

1. Launch The Playroom and advance through the introductory screens to the bedroom.
2. Click the **ABC** block on the shelf.
3. Wait one second and inspect the storybook scene.
4. Wait another 19 seconds and inspect it again.

The result is stable at the early, middle, and late checkpoints. BasiliskII retains all foreground characters at every checkpoint; Systemless omits them at every checkpoint. The middle frame scores 35.35% strict / 35.40% perceptual over the full 800×600 frame, where much of the score difference is unrelated outer desktop framing; the character omissions themselves are complete objects.

| --- | --- | --- |
| Early | Systemless omits the foreground characters | BasiliskII renders the foreground characters |
| Late | Systemless still omits the foreground characters | BasiliskII still renders the foreground characters |

## Regression test direction

Add a focused color QuickDraw fixture that composes several independently positioned, masked foreground images over an indexed offscreen background, then copies the result into a color window. Assert representative pixels from every foreground object as well as untouched background pixels. Exercise the same drawing sequence repeatedly before and after an unrelated overlay is shown so a missing foreground layer cannot pass as a timing phase.

The implementation investigation should trace the activity's foreground-resource load and offscreen-to-window composition separately. The static background and later acrobat overlay both render, which narrows the failure to the path used by the scene's foreground objects. Capture the source PixMap, mask/region, destination PixMap, transfer mode, and clip region at each blit; then compare the first foreground draw that produces no destination change with the equivalent BasiliskII result. Avoid special-casing The Playroom—the fix should implement the generic Color QuickDraw/resource behavior exposed by that first divergent operation.

### LINKED ISSUE #468: The Playroom omits printer controls and their confirmation prompts
## Observed behavior

The Playroom's elephant/yo-yo activity renders its main scene in Systemless—the elephant, yo-yo, question panel, and exit door are all present—but completely omits the small printer control below the question panel.

Under BasiliskII/System 7.5.3, the printer is visible in every displayed activity phase. It has a beige printer body and a blue page containing a house icon. In Systemless, the same area remains the untouched light-blue activity background whenever the scene is visible.

The two runtimes reach different activity/blackout phases at some equal guest-tick checkpoints, but this omission is not a timing or font difference: every visible Systemless activity frame lacks the printer, while every visible BasiliskII activity frame includes it.

Tested Systemless revision: `d1b2758a02974d80109ef240bfb1eef8d2004355` (`0.12.8`). Public archive SHA-256: `7b0b460aaa180c1e9aeb766b1e507b5bc3c59c4f4bf657eb806d1bf7c3430184`.

## Expected behavior

The elephant activity should render the printer control below the question panel exactly as it does under BasiliskII.

## Reproduction

1. Launch The Playroom and advance through the introductory screens to the bedroom.
2. Click the elephant protruding from the toybox.
3. Wait for the elephant/yo-yo activity to appear.
4. Inspect the center-right area below the question panel.

Systemless renders an empty light-blue area. BasiliskII renders the printer control there.

## Paired evidence

| View | Systemless | BasiliskII reference |
| --- | --- | --- |
| Full activity | Systemless elephant activity without printer | BasiliskII elephant activity with printer |
| Magnified printer area | Systemless empty printer area | BasiliskII printer control |

The aligned visible activity checkpoint measures 36.71% strict/perceptual over the complete 800×600 frame. The low aggregate is dominated by different outer desktop framing; the stable app-owned residual is the complete printer control.

## Regression-test direction

First identify whether the printer is sourced from a `PICT`, `cicn`, control resource, or an offscreen sprite composition. Trace the first draw that produces the printer under BasiliskII and the corresponding Systemless operation, recording:

- current GrafPort/GWorld,
- resource type and dimensions,
- source and destination PixMaps,
- mask or region,
- transfer mode and color table,
- clip and visible regions.

Then add a focused application-independent fixture for that exact Toolbox path. Compose a small indexed-color control over a larger offscreen scene, copy it into a color window across repeated redraw phases, and assert pixels in the control body, paper icon, shadow, and untouched background. Include neighboring controls so one missing element cannot pass because the rest of the scene rendered.

## Implementation direction

Compare the printer's resource load and blit path with the question panel and exit door, which already render correctly in the same scene. Fix the first generic resource, Color QuickDraw, mask/region, control, or offscreen-transfer operation whose destination differs from BasiliskII. Do not special-case The Playroom.

### LINKED ISSUE #475: The Playroom cleanup control omits its confirmation prompt
## Observed behavior

In The Playroom's alphabet-block activity, Systemless renders the cleanup control at the upper-right and visibly depresses it when clicked, but never opens the cleanup confirmation prompt.

Under BasiliskII/System 7.5.3, the same press immediately displays the custom “Shall I clean up?” surface with illustrated YES and NO choices. The prompt remains visible after release and at later checkpoints. Systemless instead continues through the activity's normal redraw cycle.

The visible pressed state proves that the control and its hit region are present and receiving the click. This is therefore distinct from the missing printer control tracked by #468, although the absent custom prompt may share a downstream composition or retained-modal path.

This is not a font difference or a one-frame timing discrepancy: the complete prompt surface is absent from every post-click Systemless checkpoint.

Tested Systemless revision: `d1b2758a02974d80109ef240bfb1eef8d2004355` (`0.12.8`). Public archive SHA-256: `7b0b460aaa180c1e9aeb766b1e507b5bc3c59c4f4bf657eb806d1bf7c3430184`.

## Expected behavior

Pressing the alphabet activity's cleanup control should open and retain the custom “Shall I clean up?” YES/NO prompt, matching BasiliskII.

## Reproduction

1. Launch The Playroom and advance through the introductory screens to the bedroom.
2. Click the ABC blocks on the shelf.
3. Wait for the alphabet landscape to settle.
4. Click the cleanup control at the upper-right, immediately left of the printer and above the trash can.
5. Observe the state after press and release.

Systemless depresses the control but does not display the prompt. BasiliskII displays the prompt immediately.

## Paired evidence

| Systemless after cleanup click | BasiliskII after cleanup click |
| --- | --- |
| Systemless depresses the cleanup control but omits the confirmation prompt | BasiliskII displays the Shall I clean up confirmation prompt |

The aligned pre-click frame measured 35.35% strict parity over the complete 800×600 frame. Post-click frames measured 26.01–26.03%; at each prompt-visible checkpoint, the app-owned residual includes the complete missing confirmation surface. Evidence was added in #476 and merged as `5425817eb0bed1d51f8f22ebcf7f39f11c5d8eb5`.

## Regression-test direction

Trace both runtimes from the successful cleanup hit through the first divergent state transition. Record the control hit result, active GrafPort/GWorld, resource loads, source and destination PixMaps, transfer modes, clip/visible regions, and any retained modal/event state.

Add an application-independent fixture that:

- draws a small indexed-color control and verifies pressed-state feedback,
- uses the same generic post-hit Toolbox path to compose a custom prompt with a border, background, text area, and two illustrated choices,
- keeps that prompt visible across subsequent update and event-loop iterations,
- verifies that clicks are routed to the prompt rather than the underlying scene.

## Implementation direction

Because Systemless reaches the pressed control state, begin after hit detection and compare the next state-changing call sequence against BasiliskII. Fix the first generic event, resource, Color QuickDraw, region, offscreen-copy, or retained-surface operation whose result diverges. Do not special-case The Playroom.

benletchford:
Closes #435.
Closes #468.
Closes #475.

Resolves the computer-letter half of #441; the counting-numeral half is handled by #535.

Materialize payload files at the safe destination paths embedded by classic multi-disk installers, and launch the installed application path when it is present. This lets ordinary relative File Manager lookups find their preinstalled data instead of creating empty replacements.

Validation:
- `cargo test`
- deterministic alphabet and cleanup probes render the foreground scene and confirmation prompt
- deterministic alphabet and elephant probes render their printer controls and open the print-confirmation prompt
- deterministic computer probe renders the requested large letter

### DIFF
--- a/src/game/launch.rs
+++ b/src/game/launch.rs
@@ -248,6 +248,8 @@ fn load_stuffit(runner: &mut FixtureRunner, file_data: &[u8]) -> Result<LoadedAp
         insert_payload_into_vfs(runner, payload, &mut executable_entry);
     }
 
+    materialize_installer_layout(runner, &mut executable_entry);
+
     log_vfs(runner);
 
     let executable =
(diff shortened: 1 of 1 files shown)


benletchford:
Closed after review because scanning arbitrary resource bytes for Pascal-looking paths and pre-materializing an inferred install tree is an installer-specific launch heuristic. It can misidentify ordinary resource data and bypasses the File Manager/installer behavior the HLE should implement. The underlying issues remain open for fixes at the first generic Toolbox divergence.

PR #536, opened 2026-08-12
feat(files): expose extracted read-only volumes

### LINKED ISSUE #326: Expose extracted read-only HFS volumes through File Manager volume APIs

benletchford:
Closes #326.

Register extracted HFS and HFS+ image roots as stable, software-locked File Manager volumes. Volume enumeration, name/reference lookup, current-volume and working-directory APIs now preserve mounted-volume identity, while data, resource, metadata, and directory mutation paths return `wPrErr`. Ordinary archive-backed VFS paths remain writable.

The abstraction intentionally does not claim a physical media type or synthesize a device driver.

Validation:
- `cargo test --lib`
- focused parameter-block coverage for deterministic enumeration, name/reference lookup, unknown-volume rejection, selected-volume and working-directory reporting, data/resource lookup scoping, write rejection, and writable unmounted paths

### DIFF
--- a/src/disk_image.rs
+++ b/src/disk_image.rs
@@ -31,8 +31,25 @@ const HFS_MAX_VOLUME_NAME_LEN: usize = 27;
 #[derive(Debug)]
 pub struct DiskImageContents {
     pub volume_name: String,
+    pub volume_info: DiskImageVolumeInfo,
     pub dirs: Vec<String>,
     pub files: Vec<DiskImageFile>,
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct DiskImageVolumeInfo {
+    pub attributes: u16,
+    pub file_count: u16,
+    pub allocation_block_count: u16,
+    pub allocation_block_size: u32,
+    pub clump_size: u32,
+    pub free_blocks: u16,
+    pub bitmap_start: u16,
+    pub allocation_pointer: u16,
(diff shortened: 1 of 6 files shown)


benletchford:
Rebased the branch onto current `master` so it inherits the restored repository formatting/Clippy gates and current File Manager state. Focused mounted-volume enumeration/name/reference tests, `cargo fmt --all -- --check`, `cargo check --no-default-features`, and `git diff --check` pass.

PR #537, opened 2026-08-12
fix(script): implement script-aware ReplaceText

### LINKED ISSUE #311: Implement the ScriptUtil ReplaceText encoded selector

benletchford:
Closes #311

## Summary

- implement the `$820CFFDC` ScriptUtil `ReplaceText` selector and its Pascal result-frame ABI;
- derive character boundaries from the current port font script so matching never begins at a double-byte trail byte;
- replace every non-overlapping occurrence while preserving handle ownership across in-place shrink and growth reallocation;
- return the documented handle and memory errors.

## Validation

- `cargo test --lib` (2944 passed, 3 ignored)
- MPW-built encoded-selector fixture: no match, multiple matches, growth, shrinkage, and Japanese double-byte boundary
- `git diff --check`

### DIFF
--- a/src/trap/toolbox.rs
+++ b/src/trap/toolbox.rs
@@ -3011,6 +3011,120 @@ impl super::TrapDispatcher {
         new_ptr
     }
 
+    fn scriptutil_font_script(&self) -> u8 {
+        let font = self.tx_font as u16;
+        if (0x4000..=0xBFFF).contains(&font) {
+            (((font - 0x4000) / 0x0200) + 1) as u8
+        } else {
+            0
+        }
+    }
+
+    fn scriptutil_is_lead_byte(script: u8, byte: u8) -> bool {
+        match script {
+            // Macintosh Japanese uses Shift-JIS lead-byte ranges.
+            1 => (0x81..=0x9F).contains(&byte) || (0xE0..=0xFC).contains(&byte),
+            // The Macintosh Chinese and Korean double-byte encodings reserve
+            // the high-byte range used by their script parse tables.
+            2 | 3 | 25 => (0xA1..=0xFE).contains(&byte),
+            _ => false,
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


benletchford:
Blocking: an empty substitution never resizes `baseText` to zero. When every key occurrence is deleted, `replaced` is empty and the `!replaced.is_empty()` guard skips `write_bytes_to_handle`, so the function reports substitutions while leaving the original bytes in the handle. Inside Macintosh: Text (1993), pp. 5-74–5-75 says each instance is replaced by the substitution text, including an empty handle. Please resize the handle for an empty result and add a focused regression that replaces all content with an empty substitution and asserts a zero-byte handle.

benletchford:
Fixed in `0159bd8d`. Empty replacements now always pass through the handle-resize path, including a zero-byte result; the regression verifies the master pointer is cleared, the former allocation is released, and ownership metadata is removed.

Validation: `cargo test scriptutil_replacetext --lib`, `cargo check --no-default-features`, and `git diff --check`.

PR #549, opened 2026-08-13
fix(menu): populate installed Apple menu items

### LINKED ISSUE #451: Apple menu omits installed System 7 Apple Menu Items
## Summary

SimCity 2000 Demo's Apple menu contains only the application's `About SimCity 2000™` item under Systemless. Under BasiliskII with System 7.5.3, the same menu also contains the installed Apple Menu Items and folders: AppleCD Audio Player, Automated Tasks, Calculator, Chooser, Control Panels, Find File, Jigsaw Puzzle, Key Caps, Note Pad, Recent Applications, Recent Documents, Scrapbook, Stickies, and Shut Down.

This is not a font difference: the menu's contents, hierarchy arrows, icons, separator, dimensions, and selectable behavior are absent.

## Evidence

| BasiliskII (System 7.5.3) | Systemless |
|---|---|
| BasiliskII Apple menu | Systemless Apple menu |

The paired whole-screen frame is 76.83% strict / 84.09% perceptual similarity. The surrounding live city agrees apart from already-tracked simulation timing drift, isolating the large additional difference to the Apple menu.

## Reproduction

1. Launch the public SimCity 2000 Demo.
2. Enter the Demo City and dismiss the welcome and annual-budget dialogs.
3. Open the Apple menu.
4. Compare the populated System 7.5.3 menu with Systemless's one-row menu.

Selecting `About SimCity 2000™` still opens the correct application About window in both environments; the defect is the missing system-provided portion below it.

## Focused test proposal

Add a Menu/Desk Manager integration test whose system resource environment contains named Apple Menu Items plus a folder-backed hierarchical item. Start with an application-authored Apple menu containing About and a trailing divider, invoke the resource-menu population path, and assert:

- the installed items are appended after the divider;
- folder items carry hierarchy semantics and open their child menu;
- icons and enabled states survive serialization;
- duplicate names are not appended;
- an empty Apple Menu Items source continues to suppress the otherwise-trailing divider.

A second interaction assertion should choose one appended item and verify that selection is dispatched to the system item rather than returned as an application About command.

Model the System 7 Apple Menu Items source as part of the virtual System Folder, then merge those entries into the application-authored Apple menu when the resource-menu population path runs. Reuse the existing menu serialization, icon, hierarchy, enable-state, and duplicate-filtering machinery. Dispatch should distinguish application menu commands from system-provided items; regular application aliases/folders can go through the virtual launch/navigation path, while classic DRVR-backed desk accessories can remain explicitly unsupported until the Desk Manager/device-driver lifecycle is implemented.

This keeps the fix generic for every System 7 application instead of special-casing SimCity or drawing synthetic rows.

benletchford:
## Summary
- append extracted `System Folder/Apple Menu Items` entries when applications call `AddResMenu("DRVR")`
- preserve folder structure as hierarchical submenus and render application `ICN#` resources as reduced menu icons
- launch installed applications through the existing virtual application path while keeping system-owned selections out of application command dispatch
- leave unsupported classic desk accessories visible but inert

## Root cause
The archive loader already preserved Apple Menu Items in the virtual filesystem, but `AddResMenu("DRVR")` only enumerated named resources from loaded resource files. The Menu Manager therefore never saw the installed filesystem entries.

This PR is stacked on #514 because populated System 7 Apple menus can exceed the legacy fixed-size MENU record.

## Validation
- `cargo test --lib` (2943 passed, 3 ignored)
- `cargo test --lib trap::menu::tests` (139 passed)
- touched-file rustfmt and `git diff --check`

Closes #451.

### DIFF
--- a/src/trap/dispatch.rs
+++ b/src/trap/dispatch.rs
@@ -1567,6 +1567,14 @@ pub struct TrapDispatcher {
     /// returned. Limit redelivery to once per tick so an animation loop that
     /// ignores mouseDown events can still make forward progress.
     pub(crate) pending_native_menu_event_tick: Option<u32>,
+    /// VFS-backed System Folder Apple Menu Items keyed by the menu selection
+    /// that represents them. These entries are owned by the Menu/Desk Manager
+    /// rather than by the foreground application.
+    pub(crate) apple_menu_item_paths: HashMap<(i16, i16), String>,
+    /// Synthetic menu icon resources installed for VFS-backed Apple Menu
+    /// Items. The menu item's one-byte icon number indexes this map through
+    /// the normal icon-number-plus-256 Menu Manager convention.
+    pub(crate) apple_menu_item_icons: HashMap<u8, u32>,
     /// Active control tracking state (currently popup-menu TrackControl).
     pub(crate) control_tracking: Option<ControlTrackingState>,
     /// Underline info for continuous underline across a string (set by draw_string)
@@ -3034,6 +3042,8 @@ impl TrapDispatcher {
             pending_native_menu_selection: None,
             pending_native_menu_event: None,
             pending_native_menu_event_tick: None,
+            apple_menu_item_paths: HashMap::new(),
+            apple_menu_item_icons: HashMap::new(),
             control_tracking: None,
             underline_info: None,
             mouse_pos: (0, 0),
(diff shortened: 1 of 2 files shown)


benletchford:
Let's not merge this as it conflicts with the vision of the project. We don't want to show these utilities as there IS no system. Current behaviour (not showing anything) is preferred.

PR #556, opened 2026-08-13
fix: render indexed PixMap data in PICT BitsRect opcodes

### LINKED ISSUE #555: Render indexed PixMap data in PICT BitsRect opcodes
Systemless currently treats PICT `BitsRect`/`BitsRgn` (`$0090`/`$0091`) as 1-bit `BitMap` records unconditionally. Color QuickDraw also permits their `rowBytes` high bit to select an indexed `PixMap` plus color table, with unpacked pixel rows when `rowBytes < 8`. Parsing that form as a `BitMap` desynchronizes the stream and can panic while interpreting pixel bytes as a reserved opcode length.

Implement the documented PixMap form generically, keep the picture-length scanner synchronized, and add a focused indexed 4-bit regression test. Reference: *Inside Macintosh, Volume V*, Color Picture Format, pp. V-94 and V-103.

benletchford:
Closes #555.

Color QuickDraw permits `BitsRect` and `BitsRgn` to carry an indexed `PixMap` when the `rowBytes` high bit is set. Parse that documented form through the existing indexed-pixel renderer, keep PICT stream-length discovery synchronized, and cover the 4-bit uncompressed form with a regression test.

Reference: *Inside Macintosh, Volume V*, Color Picture Format, pp. V-94 and V-103.

Tests:
- `cargo fmt --check`
- `cargo test` (2950 passed, 3 ignored; binary and doc tests passed)

### DIFF
--- a/src/trap/pict.rs
+++ b/src/trap/pict.rs
@@ -1178,24 +1178,59 @@ pub fn draw_picture(
             }
             0x88..=0x8C => {}
             0x90 | 0x91 => {
-                // BitsRect / BitsRgn - 1bpp bitmap
-                pos = parse_bits_rect(
-                    bus,
-                    pos,
-                    opcode == 0x91,
-                    dst_top,
-                    dst_left,
-                    frame_top,
-                    frame_left,
-                    scale_x,
-                    scale_y,
-                    screen_mode,
-                    device_clut,
-                    fg_idx,
-                    bg_idx,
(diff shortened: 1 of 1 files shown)


benletchford:
Closed as superseded by #607, which covers the same unpacked indexed BitsRect path and also bounds PICT stream scanning so malformed lengths cannot desynchronize or overflow the decoder.

PR #559, opened 2026-08-13
fix: preserve front windows over retained dialog snapshots

### LINKED ISSUE #557: fix: repaint gameplay exposed behind dismissed auxiliary windows
### Summary

The original Macintosh Lode Runner reaches its authentic level-one attract mode, but selecting **Game → Start Game** leaves the High Scores panel retained over most of the live board. The game does start: it reads the level data, enables the in-game commands, and accepts the documented J/L movement controls, with the player visibly moving in the narrow exposed areas.

Both a normal Command-S menu shortcut and the native guest-menu selection path reproduce this. MenuSelect returns the expected `$00040001` value, so the failure occurs after successful menu dispatch.

### Expected behavior

Starting a game removes or repaints the High Scores surface and exposes the complete live level-one board.

### Actual behavior

The High Scores pixels remain over the board while gameplay continues underneath.

### Technical direction

Audit auxiliary-window removal/visibility and the update-region lifecycle for an uncovered document window. Inside Macintosh Volume I, pp. I-283–I-286 describes window removal/front promotion; exposed content must subsequently be updated by the application through the normal update-event path. This should be fixed generically in the Window/Event Manager rather than special-casing the title.

### Regression coverage

Add focused Window Manager tests for exposing a document window after an overlapping auxiliary window is removed or hidden, plus a gameplay smoke route that proves the panel no longer obscures the moving level.

benletchford:
## Summary

- keep retained modeless-dialog snapshots behind visible windows that precede them in Window Manager z-order
- preserve partially exposed dialog pixels while restoring overlapping front-window structures
- add a regression test for both the covered and exposed portions of a retained background dialog

## Verification

- `cargo fmt --check`
- `cargo test --lib` (2950 passed, 3 ignored)
- original Macintosh Lode Runner reaches an unobscured level-one board and visibly accepts documented J/L movement

Closes #557

### DIFF
--- a/src/trap/dialog.rs
+++ b/src/trap/dialog.rs
@@ -5157,7 +5157,7 @@ impl super::TrapDispatcher {
         self.ensure_dialog_background_saved(bus, port, bounds);
     }
 
-    fn dialog_saved_pixel_rect(rect: (i16, i16, i16, i16)) -> (i16, i16, i16, i16) {
+    pub(super) fn dialog_saved_pixel_rect(rect: (i16, i16, i16, i16)) -> (i16, i16, i16, i16) {
         let (top, left, bottom, right) = rect;
         let margin = Self::DBOX_FRAME_MARGIN;
         (top - margin, left - margin, bottom + margin, right + margin)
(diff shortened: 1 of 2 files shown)


benletchford:
Superseded by #560, which incorporates the z-order-preserving snapshot behavior together with the exposed-window invalidation needed to resolve #557.

PR #562, opened 2026-08-13
fix: implement resumable HFS catalog searches

### LINKED ISSUE #561: PBCatSearch must terminate catalog scans instead of returning empty noErr
## Problem

A Motorola 68K application that probes the startup volume with `PBCatSearch` never completes startup. Systemless currently falls through the HFS dispatch handler for selector `$0018`, returns `noErr`, reports zero matches, and leaves `ioCatPosition` unchanged. The application therefore repeats the same search forever.

## Expected behavior

Implement the documented `PBCatSearch` termination contract. A catalog search must return matching `FSSpec` records while more catalog data is available, and must return `eofErr` after the entire volume has been searched. Even a minimal no-match search must terminate with `eofErr` instead of returning an empty `noErr` forever.

Inside Macintosh: Files, “Searching a Catalog”, documents selector `$0018`, the parameter block fields, and the `noErr`/`eofErr` distinction.

## Reproduction

Launch Lode Runner: The Legend Returns v1.1.4 from its 68K HFS image. The application installs its menus, then repeatedly calls HFS selector `$0018` with no visible game window and never becomes interactive.

benletchford:
Closes #561.

## Summary

- implement `PBCatSearch` for HFS dispatch selector `$0018`
- filter catalog entries using names, attributes, Finder info, fork sizes, dates, valence, parent IDs, and negated searches
- return `FSSpec` matches, persist `ioCatPosition`, and distinguish `noErr` from terminal `eofErr`
- add regression coverage for empty searches and resumable bounded searches

## Verification

- `cargo fmt --check`
- `cargo test --lib` (2,951 passed, 3 ignored)

### DIFF
--- a/src/trap/resource.rs
+++ b/src/trap/resource.rs
@@ -6855,11 +6855,223 @@ impl super::TrapDispatcher {
                 let dir_id = bus.read_long(pb + 48);
                 let filename = Self::read_pb_filename(bus, name_ptr);
                 let fdir_index = bus.read_word(pb + 28) as i16;
+                let resolved_vref = self.resolve_volume_ref_num(vref);
+                if selector == 24 {
+                    // PBCatSearch ($A260, selector $0018)
+                    // Searches a volume catalog and returns matching FSSpec records.
+                    // FUNCTION PBCatSearch(paramBlock: HParmBlkPtr; async: Boolean): OSErr;
+                    // Files 1992, 2-38 to 2-42 and 2-205 to 2-207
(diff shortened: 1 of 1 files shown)


benletchford:
Closed as superseded by the #536 / #548 stack. That implementation searches real mounted read-only HFS/HFS+ volume catalogues, preserves volume identity, supports resumable deterministic batches, and covers the full criteria set; keeping a second host-VFS-only PBCatSearch path would create divergent semantics.

PR #564, opened 2026-08-13
fix(quickdraw): make indexed XOR reversible

### LINKED ISSUE #563: Shared-palette indexed CopyBits leaves Boolean transfer artifacts
## Problem

Color QuickDraw copies between indexed PixMaps in the same pixel depth and color table through RGB conversion for Boolean source modes such as `srcOr`, `srcXor`, and `srcBic`. That loses the original pixel-index bit pattern: a draw/erase pair can quantize to unrelated palette entries and leave colored artifacts behind.

Classic software commonly builds masked sprites with complementary Boolean transfers. When source and destination share the same indexed color space, those operations need to preserve the raw index-bit semantics. Copies between different color tables must continue using color conversion.

## Acceptance criteria

- Apply the eight Boolean source modes directly to pixel indices when source and destination use the same 2-, 4-, or 8-bit indexed color space.
- Mask complemented results to the active pixel depth.
- Preserve palette translation and RGB-aware behavior for differing color tables.
- Add regression coverage showing complementary Boolean transfers restore the original destination pixel.
- Keep the existing QuickDraw test suite passing.

## References

- *Imaging With QuickDraw* (1994), pp. 4-32–4-34.
- Apple Technical Note QD16, “The Joy of Being 32-Bit Clean,” on same- versus cross-device CopyBits color environments.

benletchford:
Closes #563.

Invert indexed CopyBits destinations at their packed pixel depth for srcXor and notSrcXor instead of mapping an RGB complement back through the destination palette. This preserves exact 2-, 4-, and 8-bit indices and makes repeated XOR drawing reversible.

Validation:
- `cargo fmt --check`
- `cargo test --lib` (2,950 passed, 3 ignored)

### DIFF
--- a/src/trap/quickdraw.rs
+++ b/src/trap/quickdraw.rs
@@ -3722,6 +3722,17 @@ impl super::TrapDispatcher {
                                 let dst_clut = dst_clut.as_ref().expect("indexed destination CLUT");
                                 let dst_rgb = dst_clut[usize::from(dst_pixel)];
                                 let effective_mode = if mode_base == 36 { 0 } else { mode_base };
+                                let invert_index = (effective_mode == 2 && src_rgb == [0, 0, 0])
+                                    || (effective_mode == 6 && src_rgb == [0xFFFF, 0xFFFF, 0xFFFF]);
+                                if invert_index {
+                                    let shifted_mask = pixel_mask << dst_shift;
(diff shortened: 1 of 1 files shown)


benletchford:
Closed as superseded by #601, which applies all eight Boolean source modes directly at 2-, 4-, and 8-bit indexed depths and retains the reversible XOR regression.

PR #565, opened 2026-08-13
fix: preserve indexed Boolean CopyBits operations

### LINKED ISSUE #563: Shared-palette indexed CopyBits leaves Boolean transfer artifacts

benletchford:
## Summary

- preserve raw Boolean index operations when indexed source and destination PixMaps share the same explicit color table
- mask complemented values to the active 2-, 4-, or 8-bit pixel depth
- retain RGB-aware palette conversion for copies across different color environments
- add regressions for reversible index inversion and a shared-CTable `srcOr`/`srcBic` sprite cycle

## Root cause

The indexed CopyBits path converted Boolean source modes through RGB values and nearest-palette matching even when both PixMaps shared one CTable. That destroys the stored index bit pattern, so complementary masked-sprite transfers can quantize to unrelated colors instead of restoring the destination.

Closes #563.

## Validation

- `cargo fmt --check`
- `cargo test --lib` — 2951 passed, 3 ignored
- `cargo clippy --lib` — passes with the repository's existing warnings

### DIFF
--- a/src/trap/quickdraw.rs
+++ b/src/trap/quickdraw.rs
@@ -3772,6 +3772,10 @@ impl super::TrapDispatcher {
                                     dst_ctab_handle,
                                     src_clut,
                                     dst_clut,
+                                    dst_info.pixel_size,
+                                    src_info.pixel_size == dst_info.pixel_size
+                                        && src_info.ctab_handle != 0
+                                        && src_info.ctab_handle == dst_ctab_handle,
                                     palette_translation,
                                     mode_base,
                                     src_pixel,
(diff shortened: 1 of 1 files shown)


benletchford:
Closed as superseded by #601. The newer implementation covers the same shared-palette path for all eight Boolean source modes with packed indexed regression coverage.

PR #572, opened 2026-08-13
fix: implement PBCatSearch termination and pagination

### LINKED ISSUE #561: PBCatSearch must terminate catalog scans instead of returning empty noErr

benletchford:
Closes #561.

Implements HFSDispatch selector $0018 with deterministic volume-wide catalog traversal, resumable catalog positions, FSSpec result pages, supported file/directory matching criteria, and documented eofErr termination. Invalid volume and parameter inputs now return their corresponding errors instead of falling through to an empty noErr result.

Tests cover empty-catalog termination, zero-capacity termination, unknown-volume rejection, filtered multi-page results, FSSpec parent/name fields, cursor resume, and final-page eofErr.

Validation: cargo test --lib (2,952 passed; 3 ignored).

### DIFF
--- a/src/trap/resource.rs
+++ b/src/trap/resource.rs
@@ -6861,6 +6861,129 @@ impl super::TrapDispatcher {
                 );
                 let resolved_vref = self.resolve_volume_ref_num(vref);
 
+                if selector == 24 {
+                    // PBCatSearch ($A260, selector $0018). CSParam uses a
+                    // volume-wide cursor and returns an array of 70-byte
+                    // FSSpec records. Files 1992, 2-247..2-253.
+                    let volume_name_matches = filename.is_empty()
+                        || filename.eq_ignore_ascii_case(Self::boot_volume_name());
+                    let valid_volume = volume_name_matches
+                        && (vref == 0
+                            || vref == Self::boot_volume_ref_num()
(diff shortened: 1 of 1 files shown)


benletchford:
Closing as a duplicate of #562, which already implements and validates the same issue fix.

PR #582, opened 2026-08-13
fix(pict): bound reserved opcode payloads

### LINKED ISSUE #576: PICT reserved-opcode length overflow aborts picture decoding
## Summary

`draw_picture` can panic while decoding a valid classic Macintosh PICT because its v2 reserved-opcode skip logic performs unchecked `u32` address arithmetic. A large interpreted length at `systemless/src/trap/pict.rs:1329` causes `pos += 4 + data_len` to overflow and abort the emulator instead of safely rejecting or bounding the opcode.

## Reproduction

1. Launch Spiderweb Software's developer-hosted Exile III: Ruined World v1.0.3b archive.
2. Select the `Exile III (fat) v1.0.3` executable.
3. Run through startup while the application loads its PICT resources.
4. Drawing the startup resources reaches a v2 opcode in the `0x00D0..=0x00FE` / `>=0x8100` length-prefixed class and panics:

```text
thread 'main' panicked at systemless/src/trap/pict.rs:1329:25:
attempt to add with overflow
```

The game has already opened its authentic Graphics, Sounds, Specials, and Towns companions and loaded PICT resources 3001, 800, 705, 901, 900, 821, 880, 850, and 2000 through 2016 before the abort.

## Expected

Malformed, unsupported, or out-of-bounds PICT opcode lengths must never panic the emulator. Decoder cursor arithmetic should be checked and bounded to the picture resource, with a regression test covering an overflowing length.

## Documentation

Inside Macintosh: Imaging With QuickDraw (1994), Appendix A, documents the PICT v2 opcode format and reserved opcode ranges. It also documents `PackBitsRect`/`PackBitsRgn` picture data and bounded bitmap `rowBytes` fields.

benletchford:
Closes #576.

## Cause

The v2 reserved-opcode paths in both picture rendering and initial palette discovery advanced their guest-memory cursor with unchecked `u32` arithmetic. A large 32-bit payload length could wrap the cursor and panic in debug builds, while a non-overflowing length could still carry decoding beyond the declared Picture record.

## Fix

- Compute checked, word-aligned payload endpoints for length-prefixed reserved opcodes.
- Stop decoding when an endpoint overflows or exceeds the declared nonzero `picSize` boundary.
- Apply the same bounds to the palette-peek parser so preflight cannot traverse the invalid payload independently.
- Preserve the existing RAM-bound fallback for extended pictures whose 16-bit `picSize` is zero.

## Verification

- Added regressions for a `$FFFFFFFF` reserved-opcode length and for a payload extending past the Picture record.
- `cargo test --lib` — 2,951 passed, 3 ignored.
- `cargo check --no-default-features`.
- `cargo clippy --lib` — passes with existing warnings.
- `cargo fmt --all -- --check`.
- `git diff --check`.

### DIFF
--- a/src/trap/pict.rs
+++ b/src/trap/pict.rs
@@ -148,6 +148,12 @@ fn align_pict_pos(pos: u32) -> u32 {
     }
 }
 
+fn checked_pict_payload_end(pos: u32, length_field_size: u32, data_len: u32) -> Option<u32> {
+    pos.checked_add(length_field_size)
+        .and_then(|p| p.checked_add(data_len))
+        .and_then(|p| p.checked_add(p & 1))
+}
+
 fn skip_reserved_v1_opcode(bus: &MacMemoryBus, opcode: u16, pos: u32) -> Option<u32> {
     match opcode {
         0x35..=0x37 | 0x45..=0x47 | 0x55..=0x57 => Some(pos + 8),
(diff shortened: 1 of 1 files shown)


benletchford:
Closed as superseded by #607. The newer implementation includes bounded reserved-opcode scanning together with the indexed PixMap parser that exposed this failure mode.

PR #584, opened 2026-08-13
fix(cpu): preserve executing code during overwrite

### LINKED ISSUE #580: Self-decompressing applications can overwrite low-placed CODE segments during startup
A classic 68k installer halts before InitGraf because Systemless places its resident CODE 1 segment at `$33588`, inside the application-heap range that the installer deliberately clears and expands into during startup.

The loader reports:

```
A5 world base: $110B0
CODE 1: $33588
SIZE preferred partition: 1,022,976 bytes
```

The live trace shows the installer entering its authentic byte-clear/decompression loop with `A1=$1000A`. The loop repeatedly executes:

```
$335FE  CLR.B (A1)+
$33600  DBF D1,-4
```

As A1 advances through the application heap, it eventually reaches `$335FE` and changes the running opcode from `$4219` to `$0019`, then continues overwriting the displacement and fall-through code. Execution finally halts on the overwritten `$00D9` opcode after 290,305 instructions and 22 Toolbox traps. The DBcc implementation itself is behaving correctly; m68k 0.7.x and current 0.10.13 reproduce the same overlap.

Expected: Segment Loader placement/heap reservation keeps resident CODE handles outside regions available to application startup decompression, matching classic Mac application-heap semantics.

Actual: fixed low CODE placement overlaps the installer output/clear range, so the guest overwrites its own currently executing segment.

A regression should load a SIZE-partitioned, self-decompressing multi-segment application, assert that its expansion range does not overlap loaded CODE handles, and reach InitGraf without an overwritten-opcode halt.

benletchford:
Closes #580.

Self-decompressing launchers can overwrite the backing bytes of the resident CODE segment they are still executing. The loader placement only changed where the failure occurred because the startup loop intentionally expands through its own segment, while a real 68040 continues from its instruction cache.

This snapshots resident CODE at load time and activates the affected 16-byte instruction-cache line only when an illegal fetch differs from the loaded bytes. Normal guest data reads and writes still observe the decompressed memory, while opcode fetches can finish the active startup line. Both the batch and precise execution paths recover consistently.

Verification:
- `cargo test --lib --quiet` (2,950 passed, 3 ignored)
- `cargo check --no-default-features --quiet`
- `cargo clippy --lib --quiet`
- Real Alone in the Dark CD installer passes the former 290,305-instruction / 22-trap illegal-opcode halt and reaches 32 Toolbox traps without halting.

### DIFF
--- a/src/memory/bus.rs
+++ b/src/memory/bus.rs
@@ -459,6 +459,13 @@ pub struct MacMemoryBus {
     /// the overwhelmingly common case reject in two comparisons instead of
     /// scanning the list.
     readonly_code_span: Option<(u32, u32)>,
+    /// Original bytes for resident CODE segments that may be executing while
+    /// a startup decompressor overwrites their backing handles. A cache line
+    /// is activated only when execution encounters an overwritten opcode,
+    /// modeling the 68040 instruction cache without hiding ordinary data
+    /// writes from the guest.
+    executable_snapshots: Vec<ExecutableSnapshot>,
+    execution_pc: u32,
     /// Free list: maps aligned_size → stack of recycled addresses
     free_blocks: HashMap<u32, Vec<u32>>,
     /// Tracks the aligned size of each allocation (address → aligned_size)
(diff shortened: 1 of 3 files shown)


benletchford:
Closed after review because replaying a saved CODE snapshot only after an illegal instruction is neither faithful instruction-cache emulation nor a loader-placement fix. It can selectively execute stale bytes and mask unrelated self-modifying-code faults. #580 remains open for correct heap/segment placement or CPU cache semantics.

PR #585, opened 2026-08-13
fix: preserve MIDI Manager dispatch frames

### LINKED ISSUE #583: SoundDispatch selector $00640004 leaves a Director extension argument on the stack
## Observed behavior

The official Macintosh v1.2 demo of *The Journeyman Project* loads its Director runtime and enumerates all files on its HFS volume, but Systemless halts on a white screen before the demo draws:

```text
[SOUND] SoundDispatch selector=$00640004 routine=$64 param_bytes=0
[RUN_STEPS] Invalid PC $0000003B ... sp=$03FFFD84
```

The invalid PC varies with the unconsumed stack contents (another run reached `$20047FF4`), but the instruction sequence is stable:

```text
003B2E2C  3F2C   ; push a two-byte value
003B2E30  203C   ; load selector $00640004
003B2E36  A800   ; _SoundDispatch
...
003B2E54  4E75   ; RTS through the leftover word
```

With `SYSTEMLESS_TRACE_SOUND=1`, the last dispatch before the bad return is consistently selector `$00640004`.

## Expected behavior

The legacy/Director Sound Manager dispatch should preserve the Pascal stack contract (or return an appropriate unsupported-selector error) so the extension returns to its caller and the demo can continue.

## Cause

`sound_dispatch_param_bytes($00640004)` returns zero because the selector has no encoded high-byte parameter count and there is no legacy-selector override. The unknown-selector arm therefore neither consumes the observed two-byte argument nor writes a result. The extension later executes `RTS` with that word still on the stack and branches outside guest RAM.

This needs a generic Sound Manager interpretation of selector `$00640004`; it should not be special-cased for the game.

## Reproduction

1. Download ScummVM's official Director 3 Macintosh demo archive:
   https://downloads.scummvm.org/frs/demos/director/d3/en/jman-demo-mac-12.zip
2. Verify SHA-256:
   `497bd293510e5266831237ffaac05086629ad5c34d26c41a30e2cba7837e6339`
3. Decode the MacBinary II entries into an HFS volume, preserving both forks and Finder metadata.
4. Launch the `Journeyman Demo` APPL in Systemless with `SYSTEMLESS_TRACE_SOUND=1 SYSTEMLESS_TRACE_BUFFER=1`.
5. Advance through startup; the runner halts around guest tick 605 at the invalid return PC.

The archive also contains `Journeyman Looping Demo`; the reproduction intentionally selects the interactive `Journeyman Demo`.

benletchford:
## Summary

- distinguish the legacy MIDI Manager `$0004` selector family from Sound Manager routines
- implement the MIDI 2.0 version, sign-in, port, connection, wake-up, and stop-time selectors as a generic silent facade
- preserve each selector's documented Pascal stack frame and return values
- add regression coverage for the version result, return slots, reference number, and stack movement

## Root cause

`SoundDispatch` previously routed only on the selector's routine byte. That made `$00000004` (`MIDIVersion`) fall through to `SndStartFilePlay`, while later MIDI selectors retained their arguments because the MIDI family does not encode parameter sizes in the selector. A Director application eventually returned through those arguments to an invalid PC.

The new handling uses the Universal Interfaces 3.4 `MIDI.h` declarations for the legacy ABI. Systemless exposes a silent MIDI facade: clients can register and connect ports without corrupting execution, while packet transport and timer callbacks remain inactive.

## Validation

- `cargo test --lib` — 2,951 passed, 3 ignored
- `cargo fmt --check`
- replayed the affected Director title through 2,102 guest ticks to its stable interactive gameplay interface

Closes #583

### DIFF
--- a/src/trap/sound.rs
+++ b/src/trap/sound.rs
@@ -45,6 +45,14 @@ fn sound_dispatch_param_bytes(selector: u32) -> u32 {
     }
 
     match selector {
+        // MIDI Manager selectors use a fixed $0004 family word and do not
+        // encode their Pascal frame sizes in the selector. Universal
+        // Interfaces 3.4, MIDI.h.
+        0x0004_0004 => 16, // MIDISignIn(clientID, refCon, icon, name)
+        0x001C_0004 => 14, // MIDIAddPort(clientID, bufSize, refnum, init)
+        0x0024_0004 => 16, // MIDIConnectData(srcClID, srcPortID, dstClID, dstPortID)
+        0x0048_0004 => 14, // MIDIWakeUp(refnum, time, period, timeProc)
+        0x0064_0004 => 2,  // MIDIStopTime(refnum)
         // The Inside Macintosh VI selector table documents these SoundDispatch
         // selectors with a zero param-size byte even though the Pascal
         // routines have stack arguments. Some clients pass those literals.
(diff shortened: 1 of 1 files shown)


benletchford:
Superseded by #586, which contains the evidence-backed fix for #583 and is ready for review.

PR #598, opened 2026-08-14
fix(window): coalesce activation events through classic pending slots

### LINKED ISSUE #597: Civilization II jumps through a null dialog callback on the first map
## Summary

The public 68K Civilization II demo launches, completes new-game setup, creates the first world, and writes its autosave, but then loses control flow while presenting the first map dialog. The terrain also appears as repeated red/black placeholder tiles instead of the expected map artwork.

## Reproduction

1. Load `https://assets.systemless.org/games/civilization-ii.kpk` with the `Civilization II DEMO (68K)` executable.
2. Dismiss the startup notice.
3. Start a new game and accept the default world, difficulty, civilization, leader, and city-style choices.
4. Dismiss the historical introduction to enter the map.

## Observed behavior

- The map window appears and an autosave named `Caesar Auto` is written.
- Terrain is rendered as repeated red/black placeholder polygons.
- The runner halts before accepting gameplay input.
- A bounded instruction trace ends with a `JSR (A1)` at guest PC `$007B8250`, followed by execution at PC `$00000000`, consistent with a missing/null callback target.

## Expected behavior

The initial map and dialog should render authentically, the dialog callback should return to live game code, and unit/map input should remain responsive.

## Acceptance criteria

- The initial Civilization II map renders without placeholder terrain.
- Dismissing the first map dialog does not jump to PC zero or halt the runner.
- A deterministic script can capture the live map before and after a unit command.

benletchford:
Fixes #597.

## What changed

- Model the Window Manager's `CurActivate` and `CurDeactive` low-memory globals as the two replaceable pending activation slots.
- Clear each slot when its event is delivered.
- Route window and dialog activation generation through the shared slot-aware path.
- Activate a window created frontmost with `behind = (WindowPtr)-1` even when it is initially invisible, matching classic Window Manager behavior.
- Add regression coverage for rapid superseding transitions and invisible frontmost creation.

## Root cause

Activation events bypass the classic Operating System event queue, but the HLE accumulated them in an unbounded FIFO. It also skipped activation when a frontmost window was created invisible. A sequence of rapid window changes could therefore deliver a superseded deactivation event after an application had released and reused the target window's refCon storage, causing guest dispatch through stale data.

## Impact

Applications that construct several frontmost windows before polling events now observe the latest pending activation/deactivation pair and the documented active state for invisible frontmost windows. Civilization II proceeds through world generation instead of jumping through a stale callback.

## Validation

- `cargo test pending_activation --lib`
- `cargo test invisible_frontmost_new_window --lib`
- Authentic Civilization II 68K demo automation proceeds past world generation to tick 15,372 and remains responsive after input through tick 16,278.

### DIFF
--- a/src/trap/dialog.rs
+++ b/src/trap/dialog.rs
@@ -9116,14 +9116,7 @@ impl super::TrapDispatcher {
                             && event.message == prev_window
                             && (event.modifiers & 1) != 0
                     }) {
-                        self.event_queue
-                            .push_back(crate::trap::dispatch::QueuedEvent {
-                                what: 8,
-                                message: prev_window,
-                                where_v: 0,
-                                where_h: 0,
-                                modifiers: 1,
-                            });
+                        self.queue_window_activation_event(bus, prev_window, true);
                     }
                     self.draw_single_window_chrome_inline(bus, prev_window, true);
                     let exposed_local = (
--- a/src/trap/event.rs
+++ b/src/trap/event.rs
@@ -404,6 +404,7 @@ impl super::TrapDispatcher {
             let event = self.event_queue[idx].clone();
             if self.consume_retained_modal_dialog_event(cpu, bus, &event) {
                 self.event_queue.remove(idx);
+                self.acknowledge_window_activation_event(bus, &event);
                 if trace_input_enabled() || super::dispatch::trace_delivered_events_enabled() {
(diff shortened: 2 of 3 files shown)


benletchford:
Superseded by #599, which contains the identical patch and is ready for review with green CI.

PR #617, opened 2026-08-14
fix: decode indexed PixMaps in PICT BitsRect opcodes

### LINKED ISSUE #606: Decode unpacked indexed PixMaps in PICT BitsRect opcodes
## Problem

Systemless treats every PICT `BitsRect` / `BitsRgn` opcode (`$0090` / `$0091`) as a 1-bit `BitMap`. Version 2 pictures may instead set the high bit of `rowBytes` and supply a multi-bit `PixMap`, color table, rectangles, transfer mode, and unpacked pixel data.

When such a picture is decoded, the parser consumes the PixMap fields as rectangle/data fields, becomes desynchronized, and can interpret pixel bytes as a reserved opcode length. In debug builds this currently panics at `pos += 4 + data_len` with integer overflow.

A public Falcon MC demo reproduces this with PICT 132: opcode `$0090`, `rowBytes=$8004`, a 4-bit indexed PixMap, and unpacked image data.

## Reference

*Inside Macintosh: Imaging With QuickDraw* (1994), Appendix A, pp. A-13 and A-17:

- For `$0090`, `$0091`, `$0098`, and `$0099`, the high bit of `rowBytes` selects PixMap versus BitMap.
- `BitsRect` and `BitsRgn` pixel data is unpacked.
- Indexed PixMap forms include the PixMap and color table before the CopyBits rectangles and mode.

## Expected behavior

- Detect the PixMap form of `BitsRect` and `BitsRgn`.
- Decode its inline color table and unpacked indexed pixels through the existing destination-color mapping and clipping paths.
- Preserve existing 1-bit BitMap and packed `PackBitsRect` / `PackBitsRgn` behavior.
- Bound reserved-opcode skipping so malformed PICT data cannot overflow the parser position.
- Add a regression fixture using a 4-bit unpacked PixMap.

benletchford:
## Summary

- decode indexed PixMap forms of PICT `BitsRect` and `BitsRgn` opcodes
- preserve inline color tables for destination palette mapping
- recognize the same form while scanning and measuring PICT streams
- stop safely when a reserved opcode declares an overflowing payload length

## Root cause

`BitsRect` and `BitsRgn` were always parsed as one-bit BitMaps. PICT v2 also permits an indexed PixMap when the high bit of `rowBytes` is set, so the decoder skipped the PixMap header and color table and then interpreted image bytes as later opcodes.

## Validation

- `cargo fmt --check`
- `cargo test --lib trap::pict::tests` (31 passed)
- regression coverage for an unpacked indexed PixMap `BitsRect`
- regression coverage for reserved-opcode length overflow

Closes #606

### DIFF
--- a/src/trap/pict.rs
+++ b/src/trap/pict.rs
@@ -148,6 +148,11 @@ fn align_pict_pos(pos: u32) -> u32 {
     }
 }
 
+fn checked_skip_pict_data(pos: u32, length_field_size: u32, data_len: u32) -> Option<u32> {
+    let next = pos.checked_add(length_field_size)?.checked_add(data_len)?;
+    next.checked_add(next & 1)
+}
+
 fn skip_reserved_v1_opcode(bus: &MacMemoryBus, opcode: u16, pos: u32) -> Option<u32> {
     match opcode {
         0x35..=0x37 | 0x45..=0x47 | 0x55..=0x57 => Some(pos + 8),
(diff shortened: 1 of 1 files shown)


benletchford:
Superseded by #607, which preserves the general unpacked BitsRect path and now also includes bounded opcode scanning and stream-synchronization coverage.

PR #662, opened 2026-08-17
perf(runner): budget idle-cycle probes per site and tick; cap the journal

rlanday:
From Claude Fable 5, working with @rlanday. One of three independent startup PRs (with #663 and #664); each stands alone.

## What

An exact-state idle-cycle probe (#138 / #644) journals the original value of every guest byte written between two same-tick visits to a poll site, so the proof can verify that a candidate wait cycle restored all of guest RAM. While a journal is open, every guest store loses its in-bounds fast path (`write_long` → `write_word` → `write_byte`, one hash insert per byte) and the m68k fastmem window is withdrawn for the whole core.

Two things made that expensive at sites that poll a journal-complete trap **between bursts of real work** — EV Override's boot and its "measuring the speed of your computer" calibration are exactly that shape:

1. **A failed probe was re-armed on the very next same-tick arrival.** Wait-stats over 17 s of EV Override boot on v0.16.0: `probe_starts=928,781`, `fail_mem=926,295` — ~55,000 fresh journals per second, each paying its `HashMap` growth and each suspending fastmem for the work that followed. (Aside: this is also where the "crawl residual `fail_mem` ≈ 929K" in the #644 discussion came from — those counters are cumulative from process start; it was the boot.)
2. **The journal had no size bound.**

**Fix.** Budget probes at two per (site, tick): a first probe and one retry — a loop whose first iteration still carries setup proves on the second — and past that the site is left alone until the tick changes. Cap the journal at 4,096 distinct addresses: past it the bus voids the probe on the spot (fast paths and fastmem return mid-burst) and the runner marks the site busy for the tick. A genuine wait proves on its first or second probe and writes tens of bytes, so neither bound can reject it; the crawl proof from #644 is untouched. Wait-stats gains `probe_overflows`.

## Evidence

Same 17 s boot, same binary lineage, wait-stats at close:

| | v0.16.0 | this PR |
|---|---|---|
| `anchor_calls` | 934,116 | 934,064 |
| `probe_starts` | **928,781** | **60** |
| `fail_mem` / `fail_cpu` | 926,295 / 1,243 | 30 / 30 |
| `probe_overflows` | – | 0 |

Instruments Time Profiler on the v0.16.0 baseline, from the first instruction: under `run_steps_internal`, `write_byte` towers of `hashbrown::raw::RawTable::reserve_rehash` (162 ms) + `hash_one` (46 ms) under `dispatch_move` alone, with parallel towers under the other store paths — plus the boot running without fastmem. With this PR those towers are gone from the profile.

Whole-boot effect is reported on the combined branch in #663's description (the three PRs together: instructions retired −33 % from launch to the registration dialog); this PR is the part that removes the journal churn and restores fastmem during boot.

## Tests

- Bus: restore-perfect writes up to the cap keep the probe verifiable; one more distinct address voids the journal on the spot, restores `fast_mem_window()`, makes `finish_write_probe_unchanged()` false, and reports the overflow exactly once; rewriting one address many times is one entry.
- Runner: a busy poller overflows one journal and is not re-probed for the rest of the tick; a site whose probes keep failing on memory gets exactly two per tick, then plain returns with fastmem available; the next tick re-probes and a genuine cycle proves and parks as before.
- Full lib suite passes (3,775).

### DIFF
--- a/src/memory/bus.rs
+++ b/src/memory/bus.rs
@@ -491,7 +491,21 @@ pub struct MacMemoryBus {
     /// An out-of-range write makes the probe unverifiable even though the
     /// normal bus retains its legacy warn-and-ignore behavior.
     write_probe_invalid: bool,
+    /// The journal outgrew [`WRITE_PROBE_MAX_ENTRIES`] and was dropped: the
+    /// probed cycle was doing work, not waiting. Sticky until the runner
+    /// takes it, so the verdict survives the journal's disappearance.
+    write_probe_overflowed: bool,
 }
+
+/// An exact-state probe journals the bytes an idle cycle writes: a spilled
+/// register or two, an event record, a saved keymap -- tens of bytes, a few
+/// hundred at most. A journal past this many distinct addresses is not
+/// watching a wait; it is watching real work, and while it stays open every
+/// guest store pays a hash insert and fastmem is suspended for the whole
+/// core. Past the cap the probe is voided on the spot so normal fast paths
+/// resume immediately, and the runner backs off from that site for the rest
+/// of the tick.
+pub(crate) const WRITE_PROBE_MAX_ENTRIES: usize = 4096;
 
 /// RAM storage - either owned vector or borrowed slice
 enum RamStorage {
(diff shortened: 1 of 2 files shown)


benletchford (approved):
The probe budget and bounded journal preserve the idle-cycle proof semantics while placing explicit limits on repeated observation cost. Platform CI passed; the cancelled headless job was a six-hour Linux dependency-install timeout, and the full combined integration head passes build, tests, headless check, docs, and package dry-run locally.

PR #663, opened 2026-08-17
perf: decompress StuffIt forks on every core

rlanday:
From Claude Fable 5, working with @rlanday. One of three independent startup PRs (with #662 and #664); each stands alone.

## What

Fork decompression dominates a desktop launch: EV Override's 10.8 MB SIT-5 archive costs ~1.5 s of arithmetic decoding on one core — 29 % of the whole 5 s boot window in an Instruments trace taken from the first instruction — all of it serial, inside `load_stuffit`'s entry loop.

Every entry is independent (`SitEntry::decompressed_forks` takes `&self` and its doc comment says "designed to be called from parallel contexts"), so this PR decodes file entries with one scoped worker per available core. Workers pull entries through a shared atomic counter — so a few large forks cannot strand the other workers — and the results are consumed by the existing loop in archive order, first error included: callers observe exactly the sequential behavior. wasm32 (no threads) and single-entry archives keep the plain sequential path. Only `load_stuffit` changes; nested archives, BinHex payloads and web packs are untouched.

## Evidence

**Fixed-work headless load** (`--headless --max-instructions 1000 grand9.sit`: parse + decode + VFS insert + a bounded sliver of emulation), three runs per arm, binaries differing only in this change:

| arm | wall | user CPU |
|---|---|---|
| serial (v0.16.0) | 2.14–2.21 s | 2.05–2.10 s |
| parallel (this PR) | **1.04–1.10 s** | 2.14–2.34 s |

Wall −51 %. The user-CPU column is the honest cost: ~+8 % total cycles for coordination, spread across cores instead of serialized on the main thread. The floor is the largest single fork (one fork is one decode stream). Boot output is byte-identical: the headless boot screenshot at the same instruction count has the same MD5 on both binaries.

**Whole boot, GUI, all three startup PRs together** (this + #662 + #664 on one binary vs v0.16.0; launch to a fixed 25 s wall time — both arms are idle at the registration dialog well before that; `time -l` instructions retired, which is insensitive to machine load; interleaved pairs):

| | v0.16.0 | combined |
|---|---|---|
| instructions retired | 37.6–37.7 G | **25.2–25.3 G (−33 %)** |
| cycles | 21.1–21.2 G | 15.4–17.9 G |
| CPU (user) | 5.95–5.97 s | 4.24–5.33 s |

Three further pairs of total CPU at 25 s (`ps time`): 6.04→5.32, 6.70→6.03, 8.69→6.01 s. Ryan's Instruments run of an earlier combined build showed the decode leaving the main thread's critical path (main-thread decode 800 ms → 0; workers 2.2 s total).

Decoding itself is being made cheaper upstream in benletchford/stuffit-rs#21 (−16 % instructions on this archive); that lands here through a dependency bump and composes with this change.

## Tests

The equivalence test builds a real SIT-13 archive (17 entries of assorted sizes, exercising genuine decompression), re-parses it, and requires the parallel decode to match per-entry sequential decompression byte for byte, in order. Full lib suite passes.

### DIFF
--- a/src/game/launch.rs
+++ b/src/game/launch.rs
@@ -15,7 +15,7 @@ use crate::memory::MemoryBus;
 use crate::runner::{FixtureRunner, FixtureRunnerConfig};
 use std::io::{Cursor, Read};
 use std::path::Component;
-use stuffit::SitArchive;
+use stuffit::{SitArchive, SitEntry};
 
 const LEGACY_WEB_PACK_MAGIC: &[u8; 4] = b"KPK1";
 const WEB_PACK_MAGIC: &[u8; 4] = b"KPK2";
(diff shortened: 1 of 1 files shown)


rlanday:
Instruments confirmation (Time Profiler, from first instruction, `SYSTEMLESS_NATIVE_APP_BUNDLE=1` to skip the same-PID re-exec, both runs closed at the registration dialog by the same operator; exports symbolicated offline with `xctrace export` + `atos`):

| main thread | v0.16.0 (Run 1) | #662 + #663 + #664 (Run 8) |
|---|---|---|
| **busy** | **3,691 ms** | **1,845 ms (−50 %)** |
| StuffIt decode on main | ~800 ms | 0 (1.7 s on worker threads) |
| PICT `blit_row` incl (#664) | 486 | 157 |
| `write_pixel_clipped` | 131 | 54 |
| journal `reserve_rehash` + `hash_one` (#662) | 146 | 12 |
| `run_steps_internal` incl | 304 | 268 |

Each of the three mechanisms shows up where its PR says it should. The residual PICT time is the 1-bit and 2-bit source images, which the 8-bit row path doesn't cover; the next-largest QuickDraw item on the boot's main thread is now InvertRect on large offscreen ports (per-pixel today), which I'm measuring separately.

— Claude Fable 5

benletchford (approved):
The parallel fork decode keeps deterministic ordering and error selection, has focused equivalence coverage, and passes the complete CI. It also integrated cleanly with the VISE launch work on the combined head.

PR #664, opened 2026-08-17
perf(pict): blit region-masked 8-bit PICT rows as spans

rlanday:
From Claude Fable 5, working with @rlanday. One of three independent startup PRs (with #662 and #663); each stands alone.

## What

The 8bpp srcCopy row fast path in the PICT interpreter refused any op that carried its own mask region (PackBitsRgn / BitsRgn), sending every pixel of such an image through `blit_row`'s per-pixel arm — a `PictureRegion::contains` edge scan, a transfer-table lookup and a bus write per pixel. EV Override's intro is a zoom sequence of exactly those images (region-masked 8-bit frames from 132×192 up to 450×624). In an Instruments trace of the boot from first instruction (v0.16.0), `blit_row` was 480 ms of self time plus 129 ms in `write_pixel_clipped` — 13 % of the main thread.

The port's own clip (`dst_clip`) is already applied per row as span intersections; this PR does the same for the op region. Its rows use the same edge-list encoding, in picture coordinates: convert the row's edges to destination-x spans (a trailing unpaired edge runs to the box's right edge, exactly as `contains` reads it), intersect with the `dst_clip` spans, and feed the existing slice writers — on both 8-bit and 1-bit screens. The whole-image PackBits path, which needs single-span rows, keeps refusing regions and falls through to this per-row path as before; scaling and non-srcCopy modes still take the per-pixel arm.

## Evidence

Same-window comparison of the first 8 s from the first instruction (macOS `sample`, 1 ms, all threads, native-bundle re-exec suppressed so the sampled process is the launched one; combined branch of the three startup PRs vs v0.16.0): main-thread PICT self time is present in the combined arm at 69–160 ms *because it reaches the intro within the window* while v0.16.0 is still decoding — so the fair number is the whole-boot one in #663: instructions retired launch → registration dialog **−33 %** for the three PRs together, of which this is the intro-drawing part.

Per-frame arithmetic for orientation: a 450×624 masked frame is 281 K per-pixel iterations, each with a region edge scan; as spans it is 450 row-span writes.

## Tests

- A notched region (a hole and a trailing unpaired edge), a rectangular region, and a region combined with a complex `dst_clip`, on 8-bit and 1-bit screens: the fast path's output must equal the per-pixel predicate's — `PictureRegion::contains` evaluated at every pixel — and region rows must now be admitted while scaling and srcOr are not.
- Mutation checks (each applied alone, tests must fail — all did): dropped trailing-edge handling; a translation sign error; skipping the `dst_clip` intersection when a region is present; treating an empty edge list as a full box.
- Full lib suite passes.

### DIFF
--- a/src/trap/pict.rs
+++ b/src/trap/pict.rs
@@ -5418,10 +5418,13 @@ fn can_blit_8bpp_src_copy_rows_fast(
     scrn_ps: u16,
     clip_region: Option<&PictureRegion>,
 ) -> bool {
+    // A PackBitsRgn / BitsRgn mask region does not disqualify the row path:
+    // it is applied per row as span intersections, exactly like the port's
+    // dst_clip regions. (Unused here beyond that; kept for the signature.)
+    let _ = clip_region;
     if mode_base != 0
         || (scrn_ps != 8 && scrn_ps != 1)
         || pm.cmp_count != 1
-        || clip_region.is_some()
         || scale_x != 1.0
         || scale_y != 1.0
     {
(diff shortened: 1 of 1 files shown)


benletchford (approved):
The region-row span path preserves mask boundaries and clipping with focused parity coverage, is green across CI, and passes in the combined graphics integration head.

PR #665, opened 2026-08-17
perf(shapes): fill, erase and invert rectangles a row at a time on 1-bit and 8-bit ports

rlanday:
From Claude Fable 5, working with @rlanday. Independent of #662/#663/#664 (same base); measured on top of them.

## What

`draw_generic_shape`'s per-pixel loop pays a coverage test, two clip tests and a bus read plus write per pixel. Two families of rectangle ops reduce to whole-row work but were still taking it:

- **1bpp (basic GrafPort) destinations** — every op did, and the bus is touched once per *pixel* there. Erase with a solid pattern, srcCopy Paint/Fill with a solid pattern, and Invert reduce to setting, clearing or toggling every bit of the clipped rect (exactly what `apply_boolean_transfer_1` yields per pixel), so they now run as masked read-modify-writes of the two edge bytes and one bulk write between (`fill_bytes`, or read-xor-write for Invert).
- **8bpp destinations** — the solid-fill row path already existed; it now also takes Invert (each row read, mapped through `invert_indexed_pixel`, written back — the same pure per-byte map the per-pixel arm applies).

Same guards as the existing row path: full-rect coverage, rectangular vis/clipRgn, no raw PixPat. Complex clips and patterned fills keep the per-pixel loop; pixels past a row's last byte are skipped as before.

**Why it matters for EV Override:** the game clears its 1-bit offscreen buffers at boot with an EraseRect + InvertRect pair — 107 calls each over 8.3 M pixels each (984×1152 and 750×816 ports among them; from a `SYSTEMLESS_TRACE_SHAPES_ALL` tally of one headless boot, cross-checked with a probe of the fast-path guards: all 107 inverts were `pixel_size=1`, full-rect, no complex clip). After #662/#663/#664, `draw_rect` was the largest QuickDraw item left on the boot's main thread in Ryan's Instruments run (309 ms inclusive of a 5 s window: 147 ms self + 84 ms `write_byte`).

## Evidence

Launch → fixed 25 s wall time (both arms idle at the registration dialog long before), `time -l` instructions retired / cycles (insensitive to machine load), three interleaved pairs, native-bundle re-exec suppressed so the launched pid is the process. Arms: `#662+#663+#664` vs the same plus this PR:

| | without | with this PR |
|---|---|---|
| instructions retired | 25.02–25.03 G | **22.35–22.40 G (−10.6 %)** |
| cycles | 14.1–14.9 G | 13.0–13.4 G |
| user CPU | 4.19–4.47 s | 3.70–4.00 s |

Every pair in the same direction. Cumulatively with the other three startup PRs, boot instructions v0.16.0 → here: 37.6 G → 22.4 G (−40 %).

## Tests

- 1bpp: an unaligned rect (partial bytes at both ends, whole bytes between) with each of the five ops under a rectangular clip (row path) and under the same box notched (per-pixel path): every bit inside the clip is the op's result, nothing outside changes, the two paths agree outside the notch, and a dithered pattern still reproduces per pixel. Mutation checks (each applied alone, test must fail — all did): dropped head mask, dropped tail mask, no-op toggle, wrong clear fill, dropped end-of-row clamp.
- 8bpp: InvertRect under a rectangular clip and under a notched one; the two paths agree; mutations (double inversion, ignored clip edge) fail it.
- Full lib suite passes.

### DIFF
--- a/src/trap/quickdraw.rs
+++ b/src/trap/quickdraw.rs
@@ -25775,6 +25775,221 @@ mod tests {
         );
     }
 
+    /// InvertRect on an 8bpp port: a 16x6 buffer of distinct pixels, the
+    /// rect (1,2)-(5,14), and either a rectangular clipRgn (3,4)-(5,10)
+    /// (row path) or the same box notched at row 4, columns 6-7 (per-pixel
+    /// path). Returns the buffer after the call.
+    fn invert_rect_8bpp(notched_clip: bool) -> Vec<u8> {
+        let (mut d, mut cpu, mut bus) = setup();
+        let base = bus.alloc(16 * 6);
+        for offset in 0..(16 * 6) as u32 {
+            bus.write_byte(base + offset, (offset * 7 + 3) as u8);
+        }
+        let ctab = make_test_ctab_handle(
+            &mut bus,
+            &TrapDispatcher::standard_mac_8bpp_clut(),
+            0x1234,
+            0,
+        );
+        let pixmap_handle = bus.alloc(4);
+        let pixmap_ptr = bus.alloc(50);
+        bus.write_long(pixmap_handle, pixmap_ptr);
+        write_pixmap_8(&mut bus, pixmap_ptr, base, 16, 6, ctab);
+
+        let port = bus.alloc(96);
+        bus.write_long(port + 2, pixmap_handle);
+        bus.write_word(port + 6, 0xC000);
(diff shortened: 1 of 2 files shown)


benletchford (approved):
The row-at-a-time paths are limited to semantically uniform operations and fall back for patterned/complex cases; focused 1-bit/8-bit and clipping tests plus the combined integration suite are green.

PR #666, opened 2026-08-17
perf(pict): draw 1-bit srcCopy source rows through the 8-bit row path

rlanday:
From Claude Fable 5, working with @rlanday. **Stacked on #664** (its branch is the base; the diff to review is the top commit, `perf(pict): draw 1-bit srcCopy source rows through the 8-bit row path`).

## What

A 1-bit source drawn onto an 8-bit screen went through `blit_row`'s per-pixel arm — bit test, region test, dst_clip test and a bus write per pixel. EV Override's intro is a zoom sequence of region-masked frames; after the 8-bit ones moved to the span path (#664), the 1-bit ones (`PackBitsRgn pixelSize=1`, 270×336 up to 330×432) were most of what `blit_row` still cost on the boot's main thread — 151 ms self + 53 ms in `write_pixel_clipped` in Ryan's Instruments run of the combined branch.

In srcCopy a 1-bit row is the 8-bit row path in disguise: expand the row's bits to the destination indices they map to (set → fg, clear → bg — exactly the per-pixel arm's choice) and hand the expanded row to `try_blit_row_8bpp_src_copy_fast` with an identity map, so region and dst_clip spans, screen bounds and edge trimming apply unchanged. Other modes (clear bits transparent) and 2-bit sources keep the per-pixel arm.

## Evidence

Measured together with two other small startup changes (#667, #668) on top of #662+#663+#664+#665, launch → 25 s, `time -l` instructions retired, three interleaved pairs: 22.4–22.7 G → 21.4–21.6 G (−4.5 %), cycles and CPU down in every pair. By boot-profile attribution this is the largest of the three (~120 ms of the boot's main thread).

## Tests

Four 1-bit rows onto an 8-bit screen with and without a notched op region must equal the per-pixel arm's predicate at every pixel; a non-srcCopy mode is checked to still be right; mutation checks (swapped fg/bg, reversed bit order in the expansion) each fail — proving the row path, not the fallback, produced the pixels. Full suite passes.

### DIFF
--- a/src/trap/pict.rs
+++ b/src/trap/pict.rs
@@ -4921,6 +4921,18 @@ fn parse_indexed_bits_rect(
     (pos, if pm.pixel_size == 8 { colors16 } else { None })
 }
 
+/// Identity source-to-destination index map, for rows that already hold
+/// destination indices.
+static IDENTITY_INDEX_MAP: [u8; 256] = {
+    let mut map = [0u8; 256];
+    let mut index = 0;
+    while index < 256 {
+        map[index] = index as u8;
+        index += 1;
+    }
+    map
+};
+
 /// Blit a decompressed row of pixel data to the screen.
 fn blit_row(
     bus: &mut MacMemoryBus,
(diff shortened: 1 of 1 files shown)


benletchford (approved):
Restacked to contain only the 1-bit srcCopy row path after #664 landed. Focused PICT tests and the complete cross-platform CI are green on current master.

PR #667, opened 2026-08-17
perf(memory): scribble uninitialized allocations with one bulk fill

rlanday:

## What

Every non-CLEAR `NewPtr`/`NewHandle` allocation is scribbled with `0xA5` (so an application relying on zeroed memory misbehaves here as on a real machine), and that fill went through `write_byte` one byte at a time — each with the read-only-code check, the write-probe hook and the tracer checks. EV Override's boot allocates megabytes this way (its 1-bit offscreen buffers alone are 141 KB each). `fill_bytes` performs the identical fill in one call and keeps the per-byte path whenever a tracer, watchpoint or write probe needs to observe every byte. `ReadXPRam`'s zero fill gets the same treatment via `fill_zeros`.

## Evidence

In a boot `sample` capture (first 8 s, main thread) this loop was the largest remaining `write_byte` caller after the drawing paths: 36–54 self samples of ~140. Measured together with #666 and #668 on top of #662+#663+#664+#665: launch → 25 s instructions retired 22.4–22.7 G → 21.4–21.6 G (−4.5 %) over three interleaved pairs; this is the smallest of the three by attribution.

## Tests

Odd-sized `NewPtr` and `NewPtrClear` blocks: every byte must be `0xA5` and `0x00` respectively; a one-short fill fails it. Full suite passes.

### DIFF
--- a/src/trap/memory.rs
+++ b/src/trap/memory.rs
@@ -162,10 +162,10 @@ fn scribble_uninitialized_allocation(bus: &mut MacMemoryBus, address: u32, size:
     }
     // IM:Memory 1992 / IM:II document that regular NewPtr/NewHandle
     // allocations leave contents undefined; only CLEAR variants are
-    // guaranteed to return zero-filled memory.
-    for offset in 0..size {
-        bus.write_byte(address.wrapping_add(offset), 0xA5);
-    }
+    // guaranteed to return zero-filled memory. One bulk fill: the bus keeps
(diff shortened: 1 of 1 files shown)


benletchford (approved):
The bulk scribble preserves the allocation byte pattern and bounds while replacing only the per-byte loop; focused coverage and the combined integration suite pass.

PR #668, opened 2026-08-17
perf(framebuffer): tile the desktop pattern fill by row on 8-bit screens

rlanday:

## What

`fb_fill_pattern_rect` painted the desktop pattern one pixel at a time through `fb_set_pixel`, and each pixel re-derived the logical black or white palette index through several bus reads before its write. It runs over the whole 800×600 screen at init and again on desktop redraws (window moves/closes).

On 8-bit screens the pattern repeats every 8 pixels in x and 8 rows in y, so resolve the two indices once, build the eight possible rows once, and write each scanline as one slice — exactly what `fb_set_pixel` writes per pixel. 1-bit and 4-bit screens keep the per-pixel path.

## Evidence

Boot `sample` capture: the per-pixel fill's bus reads under `init_game` were 19 main-thread samples in the first 8 s (one-time at init; the same code runs on every desktop redraw). Measured together with #666 and #667 on top of #662+#663+#664+#665: launch → 25 s instructions retired 22.4–22.7 G → 21.4–21.6 G (−4.5 %) over three interleaved pairs.

## Tests

A rect with an unaligned left edge and a top that is not a multiple of 8, and a pattern whose eight rows all differ, on a 24×12 screen: every pixel inside must be black/white per the pattern bit for (y mod 8, x mod 8) and nothing outside may change. Mutation checks: dropping the left offset and using one pattern row for all rows each fail it. The existing gray-pattern test still passes. Full suite passes.

### DIFF
--- a/src/trap/framebuffer.rs
+++ b/src/trap/framebuffer.rs
@@ -1222,6 +1222,36 @@ impl super::TrapDispatcher {
             return;
         }
 
+        if pixel_size == 8 {
+            // Row-tiled: the pattern repeats every 8 pixels in x and 8 rows
+            // in y, so build the eight possible rows once (with the logical
+            // black/white indices resolved once, not per pixel) and write
+            // each scanline as one slice -- what fb_set_pixel writes per
+            // pixel, without the per-pixel bus round trips.
+            let black = Self::logical_black_pixel_index(bus);
+            let white = Self::logical_white_pixel_index(bus);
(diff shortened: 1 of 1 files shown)


benletchford (approved):
The row tiling preserves the 8x8 desktop pattern phase and screen bounds, with focused framebuffer coverage and a green combined integration head.

PR #669, opened 2026-08-18
perf(text): stop re-scanning the colour table per glyph; iterate only the glyph box

rlanday:

## What

Text drawing was 13–18 % of the main thread in EV Override's crawl (and in flight windows that include takeoff/HUD text): in an Instruments run of the current master + the startup PRs, `draw_string` totalled 1.44 s over 65 s; in an 8 s `sample` window of the crawl, 573 of 1,619 main-thread ms. Almost none of it was drawing pixels (`write_byte`: 14 ms). Two things were:

1. **A colour-table scan per glyph.** Every shape's colour setup calls `ctab_uses_noncanonical_black` on the current GDevice's colour table (twice per glyph), and `ctab_rgb_for_value`'s fallback scan issued one bus `read_word` per entry — ~512 bus reads per character on a device table (~120 ms of the 573). It now reads the table into a local buffer once per lookup and scans that: identical semantics (ordinal shortcut, then scan), one bulk read.
2. **The glyph loop iterated the conservative draw rect** (advance × ascent-to-descent plus style padding) and evaluated the coverage closure stack at every pixel. With no style effects the closure is non-zero only inside the glyph's own box, so the plain case now iterates exactly that box (a space costs nothing). Styled text — italic/bold/underline/outline/shadow — keeps the full rect and closure stack unchanged.

## Evidence

Crawl, 8 s `sample` windows, main thread, base = master + #662–#668, three windows (two base sessions bracket the fix in time):

| | main active | `draw_string` incl | CLUT scan | text_render self |
|---|---|---|---|---|
| base | 1,617 / 1,511 ms | 573 / 545 | 119 / 113 | 376 / 338 |
| **this PR** | **1,455 ms** | **384 (−30 %)** | **0** | 248 |

Single windows resolve to about ±3 %; the two base windows differ by 7 % between themselves, so read the main-thread delta as roughly −5 to −10 % of the crawl's CPU and the text-cost delta (−30 %) as the reliable number.

## Tests

- `ctab_rgb_for_value` / `ctab_uses_noncanonical_black`: index-addressed table (ordinal hit), device table with client-ID values (scan), missing value, null handle and null table pointer.
- Glyph footprint: draws an ascender, a descender, punctuation and a space in srcOr on an 8bpp CGrafPort and requires the changed pixels to equal exactly the glyph bitmap's pixels (coverage ≥ threshold) at pen + origin — the suite previously had no test of glyph pixel output (dropping the top row of every glyph passed); mutations dropping the top row or the right column now fail.
- Full suite passes.

### DIFF
--- a/src/trap/quickdraw.rs
+++ b/src/trap/quickdraw.rs
@@ -26171,6 +26171,87 @@ mod tests {
         }
     }
 
+    /// Draw `ch` in srcOr on an 8bpp CGrafPort at pen (30, 20) and return
+    /// the set of pixels that changed from the background, together with the
+    /// pixels the glyph bitmap says must be painted (the per-pixel arm's
+    /// rule: coverage at or above MONO_COVERAGE_THRESHOLD, placed at
+    /// pen + glyph origin).
+    fn plain_glyph_footprint(ch: char) -> (Vec<(i16, i16)>, Vec<(i16, i16)>) {
+        let (mut d, mut cpu, mut bus) = setup();
+        let base = bus.alloc(64 * 40);
+        for offset in 0..(64 * 40) as u32 {
+            bus.write_byte(base + offset, 5);
+        }
+        let ctab = make_test_ctab_handle(
+            &mut bus,
+            &TrapDispatcher::standard_mac_8bpp_clut(),
+            0x1234,
+            0,
+        );
+        let pixmap_handle = bus.alloc(4);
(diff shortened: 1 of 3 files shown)


benletchford (approved):
Restacked onto the row-drawing changes with both regression blocks preserved. The glyph-box optimization and color-table caching pass the focused glyph test, the 3,806-test combined library suite, and complete cross-platform CI.

PR #693, opened 2026-08-19
fix(resource): guest-visible getters load lazily-seeded resources on demand

### LINKED ISSUE #692: Lazy resource seeding (PR #660) makes find-only lookups report present resources as missing — SimCity 2000 exits at launch
## Symptom (motivating example: SimCity 2000 1.2)

SimCity 2000 exits at launch with its own error dialog:

> Program Failure: Cannot get appPalette in PrepareColors()

Bisects to `650536c` ("fix: launch large universal applications", PR #660, for #658/#659). It launched fine before that commit; SC2K 1.2 is a plain 68K app (packaged as a zip of MacBinary files), so the regression is in the shared launch path, not the PPC work.

## Root cause

`650536c` made resource seeding **lazy**: at launch, every resource in the app's fork gets a map entry keyed `(type, id)` with a **NULL guest pointer**; data materializes on first use (`reload_resource_data_from_file`). That is a good model — it matches the real Resource Manager, where opening a file reads only the resource **map** and data is read on demand (Inside Macintosh: More Macintosh Toolbox, 1-17), with never-loaded/purged resources represented as **empty handles** (IM:MMT 1-84).

The defect: `find_resource_any` filters `ptr != 0`, i.e. it **conflates IM's "empty handle" with "resource absent."** Every call site that uses the find-only helper silently stopped seeing any resource nobody had touched yet. Two sites have already been fixed one crash at a time — `'clut'` in `59a8970` (v0.17.1) and sound in `bd45aa8` (PR #685) are both instances of this class. SC2K trips a third: `GetNewPalette(0)` → `copy_palette_resource` → find-only lookup → its (present, lazily-seeded) `pltt 0` reported missing → NIL → the app exits. Probe at the moment of failure: the app's resource map contained `pltt` ids `[0, 2, 1000, 1212]`, entry present, pointer NULL.

## The rule (from Inside Macintosh)

`GetResource` with ResLoad TRUE "reads the resource into memory if it's not already in memory" (IM:MMT 1-17); only `SetResLoad(FALSE)` callers see NIL master pointers, and they know to call `LoadResource` (IM:MMT 1-84/1-85). So:

- every **guest-visible getter** (GetNewPalette, GetCTable, GetMenu, GetPixPat, GetCIcon, GetCCursor, WDEF/CDEF/LDEF/MBAR/STR# fetches, …) must take the load-on-demand path (`find_or_load_resource_any`);
- find-only lookup is correct **only** for introspection that must not fault data in (existence checks for ResError, reverse pointer→resource maps, explicit ResLoad-FALSE semantics).

## Scope

An audit of `find_resource_any` call sites finds ~24 production uses across control/window/menu/dialog/toolbox/quickdraw, nearly all guest-visible getters — each a latent instance of the same launch-time class of bug for any app whose first touch of a resource type goes through that getter.

PR to follow: renames the find-only helper to `find_loaded_resource_any` (so no-load semantics are explicit at every remaining call site), converts the guest-visible getters to load on demand, keeps deliberate ResLoad-FALSE paths as-is, and adds a regression test that seeds a `pltt` as an IM-style empty handle and asserts `GetNewPalette` faults it in.

— Claude Fable 5, working with @rlanday (found bringing SC2K up as a perf-census workload)

rlanday:
Fixes #692.

## Motivating example

SimCity 2000 1.2 exits at launch — its own error dialog:

> Program Failure: Cannot get appPalette in PrepareColors()

Screenshot 2026-08-19 at 1 29 55 PM

Bisects to `650536c` (#660). SC2K's `PrepareColors` calls `GetNewPalette(0)`; the app's `pltt 0` is present in the resource map but lazily seeded (NULL guest pointer), the find-only lookup filters `ptr != 0` and reports it missing, SC2K gets NIL and quits. Probe at the failure: the map held `pltt` ids `[0, 2, 1000, 1212]`, entry present, pointer NULL.

## The rule (Inside Macintosh)

Lazy seeding matches the real Resource Manager — opening a file reads only the resource **map**; data loads on demand (IM: More Macintosh Toolbox 1-17), and a never-loaded/purged resource is an **empty handle** (IM:MMT 1-84). The defect is that the find-only helper conflated "empty handle" with "absent". `GetResource` with ResLoad TRUE "reads the resource into memory if it's not already in memory" — so every **guest-visible getter** must take the loading path; find-only is correct **only** for introspection that must not fault data in (ResLoad-FALSE semantics, reverse maps, existence checks). GetCTable (`59a8970`) and the Sound Manager (#685) were this same class, fixed one crash at a time; this PR does the whole audit.

## Changes

- **Rename** `find_resource_any` → `find_loaded_resource_any`, so the no-load semantics are explicit at every remaining call site and new getters reach for the loading variant by default.
- **Convert the guest-visible getters** to `find_or_load_resource_any`: `pltt` (GetNewPalette/GetPalette — the SC2K fix), CDEF/CNTL + picture-button PICTs, WDEF + window color tables, `STR#` (GetIndString), LDEF, MENU/MBAR/mctb, `ppat` (GetPixPat), `cicn` (GetCIcon), `crsr` (GetCCursor).
- **Menu item icons** (cicn/ICON/SICN at icon-number+256, MTE 1992 pp. 3-46, 3-137f) draw through deep `&self` chains that cannot load on demand; `preload_menu_item_icon_resources` faults them in at `open_menu_dropdown`, the one mutable moment before drawing.
- **Deliberate find-only survivors**, each with its reason in place: `prepare_dialog_resource_handle`'s `load_if_missing = false` arm (explicit ResLoad-FALSE semantics) and `popup_menu_item_title`'s read-only fallback (its data is materialized by `create_popup_menu_handle`).

## Verification

- New regression test `get_new_palette_reloads_a_lazily_seeded_pltt_resource`: seeds a `pltt` as an IM-style empty handle (map entry, NULL pointer, backing data registered) and asserts `GetNewPalette` faults it in — fails on the find-only behaviour.
- Full suite: 3,780 pass.
- End-to-end: SimCity 2000 boots past `PrepareColors` (verified headless with palette tracing: `GetNewPalette id=0 -> $0039489C`, previously `$00000000` + ExitToShell).

---

*From Claude Fable 5, working with @rlanday. Found while bringing SimCity 2000 up as a second m68k perf-census workload.*

### DIFF
--- a/src/trap/control.rs
+++ b/src/trap/control.rs
@@ -198,7 +198,7 @@ impl super::TrapDispatcher {
         }
 
         let cdef_id = proc_id >> 4;
-        self.find_resource_any(*b"CDEF", cdef_id)
+        self.find_or_load_resource_any(bus, *b"CDEF", cdef_id)
             .map(|(_, ptr)| ptr)
             .map(|ptr| self.get_or_create_resource_handle(bus, *b"CDEF", cdef_id, ptr))
             .unwrap_or(0)
@@ -1421,10 +1421,10 @@ impl super::TrapDispatcher {
         let Some((off_id, on_id)) = Self::picture_ids_from_control_title(title_bytes) else {
             return false;
         };
-        let Some((_, off_pic_ptr)) = self.find_resource_any(*b"PICT", off_id) else {
+        let Some((_, off_pic_ptr)) = self.find_or_load_resource_any(bus, *b"PICT", off_id) else {
             return false;
         };
-        let Some((_, on_pic_ptr)) = self.find_resource_any(*b"PICT", on_id) else {
+        let Some((_, on_pic_ptr)) = self.find_or_load_resource_any(bus, *b"PICT", on_id) else {
             return false;
         };
 
@@ -3516,7 +3516,7 @@ impl super::TrapDispatcher {
 
                 // IM:I I-321: if the CNTL template cannot be read, return NIL.
                 let cntl_data = self
-                    .find_resource_any(*b"CNTL", ctrl_id)
+                    .find_or_load_resource_any(bus, *b"CNTL", ctrl_id)
                     .map(|(_, ptr)| ptr);
 
                 let Some(rsrc_addr) = cntl_data else {
--- a/src/trap/dialog.rs
+++ b/src/trap/dialog.rs
@@ -308,7 +308,7 @@ impl super::TrapDispatcher {
     }
 
     fn dialog_template_res_err(&self, dialog_id: i16) -> i16 {
-        if self.find_resource_any(*b"DLOG", dialog_id).is_some() {
+        if self.find_loaded_resource_any(*b"DLOG", dialog_id).is_some() {
             0
         } else {
             0
@@ -363,7 +363,7 @@ impl super::TrapDispatcher {
         let (refnum, ptr) = if load_if_missing {
             self.find_or_load_resource_any(bus, res_type, res_id)?
         } else {
-            self.find_resource_any(res_type, res_id)?
+            self.find_loaded_resource_any(res_type, res_id)?
         };
         let handle = if load_if_missing {
             let handle =
--- a/src/trap/dispatch.rs
+++ b/src/trap/dispatch.rs
@@ -5471,7 +5471,7 @@ impl TrapDispatcher {
         })
     }
 
-    pub(crate) fn find_resource_any(&self, res_type: [u8; 4], res_id: i16) -> Option<(u16, u32)> {
+    pub(crate) fn find_loaded_resource_any(&self, res_type: [u8; 4], res_id: i16) -> Option<(u16, u32)> {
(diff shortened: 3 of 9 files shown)


rlanday:
GUI verification (three arms, same sc2k.zip, played by @rlanday):

- **pure master**: exits at launch with SC2K's own "Program Failure: Cannot get appPalette in PrepareColors()" — the #692 symptom, reproduced interactively.
- **master + this PR**: launches, title screen, launch menu, Start New City, terrain generation, city settings dialog all work.

(The founding newspaper then hangs — that's the separate StringWidth plateau bug, #694; with both fixes applied the full flow works end to end.) — Claude Fable 5

benletchford (approved):
Guest-visible resource getters now consistently cross the lazy-seeding boundary on demand, with focused MENU/MBAR/PLTT and lookup coverage. All CI and the combined integration suite pass.

PR #694, opened 2026-08-19
fix(quickdraw): measure text width proportionally to the requested size

rlanday:
## Summary

`StringWidth` / `TextWidth` / `CharWidth` answer any point size that is not an exact baked face (or an exact 2×/3× multiple of one) with the base face's width at scale 1. Measured width as a function of `TextSize` is therefore a step function with wide plateaus — ask for 13, 14, 15, 16, or 17pt and every answer is the 12pt width.

Guest text-fitting loops never converge on a plateau. The reproducer is **SimCity 2000's founding newspaper**: its headline runs the classic `TextSize(n); StringWidth(s); adjust n; repeat` fit loop, the width never changes as the size steps, and the game never returns to its event loop — the newspaper is an unclickable modal, the date freezes at Jan 1900, and no input (mouse or key) can ever dismiss it.

## Fix

Scoped to the three measuring traps only: the width is the resolved base face's advance total scaled linearly, `base_total × tx_size / face_size` (rounded), which is

- **monotone** in the requested size, so fitting loops always converge, and
- **integer-exact identical** to the previous answer at every size that was previously correct (exact faces and 2×/3× multiples) — no behaviour change for well-measured sizes.

Rendering is untouched: glyphs still draw from the baked faces at integer scales. At in-between sizes a drawn string can differ slightly from its measured width — the standard bitmap-font compromise, and the reason the change is limited to measurement.

## Note on verification base

SC2K only reaches the newspaper once #693 (resource load-on-demand) is in — before that it exits at launch in PrepareColors (#692). The end-to-end verification below therefore ran with both fixes applied; this PR's diff is independent of #693 and each merges standalone.

## Verification

- **SC2K end-to-end (headless)**: before — the game spins forever in the `TextSize`/`StringWidth` pair at the newspaper (trap trace shows the two traps alternating with no event calls); after — the newspaper dismisses on a click and the simulation runs (date advances past Jan 1900).
- New regression test: `stringwidth_scales_monotonically_with_textsize` — strict width growth across every 3-point size span from 9 to 36pt (fails on the plateau behaviour), plus 2× parity pinning the exact-scale answers.
- All existing width/layout tests pass unchanged (they pin exact-size behaviour, which this change preserves).

Before:
2-newspaper-hang-pre-fontfix

After:
Screenshot 2026-08-19 at 1 26 51 PM

Reference image from SheepShaver running OS 9:
Screenshot 2026-08-19 at 1 22 48 PM

---

*From Claude Fable 5, working with @rlanday. Found while bringing SimCity 2000 up as a second m68k perf-census workload.*

### DIFF
--- a/src/trap/quickdraw.rs
+++ b/src/trap/quickdraw.rs
@@ -28067,6 +28067,56 @@ mod tests {
     }
 
     #[test]
+    fn stringwidth_scales_monotonically_with_textsize() {
+        // A guest text-fitting loop (`TextSize(n); StringWidth(s)` until
+        // the string fits a target width) only converges if measured
+        // width makes progress as the size steps. The baked font table
+        // has faces at a few exact sizes; every other size must
+        // interpolate rather than plateau at the base face's width --
+        // a text-fitting loop hangs the application
+        // forever on a plateau (the newspaper becomes an unclickable
+        // modal because the app never returns to its event loop).
+        let (mut d, mut cpu, mut bus) = setup();
+        let str_ptr = 0x300000u32;
+        let text = b"New City Founded";
+        bus.write_byte(str_ptr, text.len() as u8);
+        for (i, byte) in text.iter().enumerate() {
+            bus.write_byte(str_ptr + 1 + i as u32, *byte);
+        }
+        let mut widths = Vec::new();
+        for size in 9i16..=36 {
+            d.tx_size = size;
+            cpu.write_reg(Register::A7, TEST_SP);
+            bus.write_long(TEST_SP, str_ptr);
+            let result = d.dispatch_quickdraw(true, 0x08C, &mut cpu, &mut bus);
+            assert!(result.unwrap().is_ok());
+            widths.push(bus.read_word(TEST_SP + 4) as i16);
+        }
+        // Strict growth across every 3-point span: a fitting loop
+        // stepping the size by 1-3 points always sees the width move.
+        // (Plain per-step monotonicity is not asserted so a resolved-face
+        // switch may round differently at a boundary.)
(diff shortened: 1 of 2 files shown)


rlanday:
GUI verification (played by @rlanday, master + #693 as the base so SC2K can launch at all):

- **without this PR**: the founding newspaper hangs — no click or key dismisses it, and notably the window renders *incomplete: the close box is missing*. That's the same defect seen from another angle: the TextSize/StringWidth fitting loop hangs mid-layout, so the paper's chrome never finishes drawing and the app never returns to its event loop.
- **with this PR**: the newspaper draws fully, dismisses on click, and the simulation runs (title-bar date advances).

— Claude Fable 5

rlanday:
Reference behavior confirmed on real system software (SheepShaver, Mac OS 9, real Font Manager + TrueType, same SC2K 1.2 install): the founding newspaper completes layout — headline fit to the full width in a large serif, close box present — and dismisses on a single click. @rlanday confirms Systemless with this PR (+#693) now matches that behavior.

This also pins the IM basis empirically: on real system software, width-vs-size is monotone (bitmap strikes scale per IM:I I-219f; System 7+ TrueType rasterizes arbitrary sizes exactly), so SC2K's fitting loop was always written against monotone metrics — the plateau this PR removes never existed on a real Mac. Remaining delta is typographic only: the real headline is display-size TrueType; Systemless draws its baked faces at integer scales — the documented measurement/rendering compromise. — Claude Fable 5

benletchford (approved):
The width calculation now scales with the requested point size and is pinned by monotonicity coverage. macOS, Windows, headless/package, and lint checks passed; Ubuntu only timed out installing system dependencies, and the combined integration suite passes locally.

Issue #600, opened 2026-08-14
Civilization II first map renders placeholder terrain tiles


benletchford:
## Summary

After the first-map control-flow crash is fixed by #599, the public 68K Civilization II demo remains live but renders its terrain as repeated red/black placeholder polygons instead of the expected map artwork. This is a visual compatibility defect separate from the Window Manager activation-event bug in #597.

## Reproduction

1. Load `https://assets.systemless.org/games/civilization-ii.kpk` with the `Civilization II DEMO (68K)` executable.
2. Dismiss the startup notice.
3. Start a new game and accept the default world, difficulty, civilization, leader, and city-style choices.
4. Dismiss the historical introduction to enter the map.

## Observed behavior

The map remains responsive, but terrain cells appear as repeated red/black placeholder polygons rather than the game artwork.

## Expected behavior

The initial world map should render its authentic terrain tiles while retaining live unit and map input.

## Acceptance criteria

- The initial Civilization II map renders its terrain artwork without repeated placeholder polygons.
- The fix is generic and does not special-case Civilization II resources or drawing calls.
- A deterministic regression captures the rendered live map before and after a unit command.
- Existing QuickDraw and Window Manager tests continue to pass.

benletchford:
Triage decision: keeping this open pending post-release revalidation. The palette/resource fixes landed in this batch could affect this symptom, but the SimCity II route also had an independent file-resolution problem during the palette PR review, so there is no valid map checkpoint proving authentic terrain yet. Close only after the deterministic first-map oracle renders real tiles.

benletchford:
Fresh automated catalogue audit reproduction on `systemless` v0.39.1 confirms the placeholder-pattern corruption before the map is reached.

Using the current public `civilization-ii.kpk` archive and the 68k demo executable, both the startup notice and New Game dialog contain repeated red/green placeholder patterns where artwork should render. The New Game screen's large art panel is predominantly black with only fragmented white sprite strokes. Menu labels and dialog controls remain readable, which isolates the failure to the image/terrain content rather than the entire window surface.

The result was verified in a single lossless native-resolution composite across independently drawn startup screens, so it is not the sequential-frame viewer artifact that invalidated several earlier audit observations. This extends the issue's existing map-terrain evidence to the startup and setup path.

benletchford:
[...]

PR #696, opened 2026-08-19
fix(palette): explicit-tolerant palette entries install their colors

### LINKED ISSUE #695: SimCity 2000: Newspaper-close leaves the device palette stale — every later redraw dithers/miscolors until the whole map is corrupted
## Symptom (SimCity 2000 1.2, master + #693 + #694)

With the launch (#692/#693) and newspaper-hang (#694) fixes applied, SC2K reaches live simulation — and then, at the exact moment the founding newspaper is dismissed, a rectangle matching the paper's footprint renders as a 50%-gray dither over the terrain, with **red** patches where the rectangle covers window background. Critically, **forcing redraws makes it spread**: scrolling the map repaints more and more of the view in the same corrupted mode until the whole map looks like that. Screenshot attached (GUI, played by @rlanday).

So this is not stale pixels or a missed invalidation — the paper's **close corrupts drawing state**, and every subsequent fresh draw renders through it. Regions painted *before* the close stay correct until repainted.

## Hypothesis: palette activation is not restored on window close

The symptoms fit Color QuickDraw drawing through a wrong/stale device color environment: 50% dithering is the classic mismatched-color-table remap, and the red background is a nearest-color artifact. The likely mechanism: when the newspaper window comes to front, the Palette Manager activates a palette on the device; when it closes, the city window's palette (SC2K's `pltt 0`, 256 entries) must be re-activated — and isn't. IM:VI 20-9..20-13 (palettes are associated with windows; the front window's palette governs the device) describes the restore obligation.

Two corroborating observations:

- This only became visible after #693 — palettes now genuinely load, so the activation path runs with real data for the first time. Before #693, `GetNewPalette` returned NIL at launch and the question never arose.
- On a pre-lazy-seeding lineage (0.16.0 + the #694 font fix, our measurement harness), a headless run photographed ~40M instructions *after* the paper closed shows the **entire** view garbled (magenta/cyan terrain on red) — consistent with the same bug at terminal spread, after everything has been repainted through the stale state.

## Repro

zip-packaged SC2K (MacBinary app + prefs), master + #693 + #694: launch → Start New City → Done on the settings dialog → click the founding newspaper closed → observe the paper-footprint rectangle; scroll the map → observe the corruption spread.

Before scrolling:
Image

After scrolling:
Image

Reference image from SheepShaver running OS 9:
Image

— Claude Fable 5, working with @rlanday (SC2K bring-up series: #692, #693, #694)

rlanday:
Fixes #695. **Interacts with #610 — see the compatibility section below.**

## The Palette Manager model (why this one condition matters)

On an 8-bit indexed screen there is one device color table: 256 slots, each an RGB; every framebuffer pixel is an index into it. A window's *palette* is a list of entries `(ciRGB, ciUsage, ciTolerance)`, and activation (window comes to front) walks the entries deciding what happens to the device slots. The usage flags are the request language (IM: Advanced Color Imaging 1995, 1-12…1-14; IM:V V-157…V-160):

| usage | request | activation effect on the device slot |
|---|---|---|
| `pmCourteous` | "roughly this color" | none — drawing nearest-matches |
| `pmTolerant` | "this color within `ciTolerance`" | installs the RGB if nothing close enough exists |
| `pmAnimated` | "a private slot I'll rewrite" | reserves the slot, installs the color |
| `pmExplicit` *(pure)* | "entry *i* means slot *i*, whatever it holds" | **none, deliberately** |
| `pmExplicit + pmTolerant` | "slot *i*, holding this RGB" | **installs the RGB at exactly slot *i*** |

The combined form is how a full-screen application takes over the whole table in a known layout — 256 explicit-tolerant entries, entry *i* claiming slot *i* — so its pre-indexed artwork displays through its own palette.

## The bug (#695)

Since #610, activation keeps the device value for **any** entry with `pmExplicit` set, regardless of the other bits. An application whose entire palette is explicit-tolerant gets a 256-iteration no-op: its palette never lands, the device stays on the boot-time standard table, and every pixel drawn with the application's own indices displays the wrong colors (the #695 symptom: index 104 = khaki in the app's palette, magenta in the standard table; corruption spreads with each repaint as more app-indexed pixels land).

The forensic trail in #695 pinned each step: device table proven byte-perfect standard, framebuffer indices proven to be correct app-palette indices (ground-truthed against the app's tile-PICT clut), and the fix verified headless against the same install running on real system software (SheepShaver/OS 9) — terrain, water, cliffs, and surround all match the reference frame.

## The fix

Keep-device now applies only to *pure* explicit: `usage & PM_EXPLICIT != 0 && usage & (PM_TOLERANT | PM_ANIMATED) == 0`. Combined entries fall through to the ordinary install-at-index path. This restores the pre-#610 condition (which handled explicit+tolerant, IM:V V-157…V-160) and extends it to explicit+animated.

## Compatibility with #610 (Civilization II)

#610 introduced keep-device-for-any-explicit to recover Civilization II's colors (#600). This PR restores installs for combined usages, so **Civ II needs re-verification** — I don't have that fixture. Two observations for that review:

1. If Civ II regresses, the IM-faithful middle ground is **tolerance-aware installation**: explicit-tolerant installs only when the device color differs from `ciRGB` by more than `ciTolerance` (per the tolerant contract). That serves both cases — a custom palette far from the standard table installs; entries already within tolerance are preserved. Happy to implement that variant if Civ II shows it's needed.
2. #610 predates #693 (lazy resource seeding was introduced the following day, and #692 shows palette resources being reported missing/NULL in related paths). If Civ II's original symptom involved bogus `ciRGB` values from a resource-loading path, #693 may have fixed the underlying cause, leaving #610's keep-device as a workaround that can now be narrowed.

## Tests

- `activatepalette_installs_explicit_tolerant_entries_at_their_index` (reworked from #610's `..._preserves_device_rgb_for_mixed_explicit_entries` — under keep-device-for-any-explicit, explicit+tolerant is indistinguishable from pure explicit, and the real-hardware ground truth contradicts that).
- The two animate-entry tests now expect un-animated explicit-animated entries to reactivate to their own palette colors.
- **Pure-explicit preservation stays pinned** by the untouched `activatepalette_does_not_render_explicit_palette_entries`.
- Full suite: 3,779 pass.

---

*From Claude Fable 5, working with @rlanday. Fourth of the bring-up series (#692/#693, #694, #695/this).*

### DIFF
--- a/src/trap/quickdraw.rs
+++ b/src/trap/quickdraw.rs
@@ -16675,11 +16675,26 @@ impl super::TrapDispatcher {
         for (index, rgb) in updated_clut.iter_mut().enumerate().take(palette_entries) {
             let color_info = Self::palette_color_info_ptr(palette_ptr, index as u32);
             let usage = bus.read_word(color_info + 6) as i16;
-            // pmExplicit makes ciRGB an index into the device ColorTable, so
-            // it does not modify that index's current RGB value. Additional
-            // usage flags do not change the entry's explicit classification.
-            // Inside Macintosh: Advanced Color Imaging (1995), pp. 1-12--1-13.
-            if usage & PM_EXPLICIT != 0 {
+            // Usage decides how a palette entry claims its device slot
+            // (IM: Advanced Color Imaging 1995, pp. 1-12..1-14):
+            //
+            // - PURE pmExplicit names device slot `index` WITHOUT
+            //   requesting a color: the entry means "whatever that slot
+            //   currently holds", so activation must leave the slot's RGB
+            //   alone.
+            // - pmExplicit combined with pmTolerant (or pmAnimated) pins
+            //   the palette's color AT that exact slot: "slot `index`,
+            //   holding this RGB". Full-screen applications install a
+            //   complete custom palette this way -- 256 explicit-tolerant
+            //   entries, entry i claiming slot i -- so their pre-indexed
+            //   artwork displays through their own table. Skipping those
+            //   entries leaves the device on the standard table, and
+            //   every pixel drawn with the application's indices shows
+            //   the wrong colors.
+            //
+            // Only the pure-explicit case keeps the device value; the
+            // combined cases fall through to the ordinary install below.
+            if usage & PM_EXPLICIT != 0 && usage & (PM_TOLERANT | PM_ANIMATED) == 0 {
                 *rgb = current_clut[index];
                 continue;
             }
@@ -36247,10 +36262,16 @@ mod tests {
     }
 
     #[test]
-    fn activatepalette_preserves_device_rgb_for_mixed_explicit_entries() {
-        // pmExplicit identifies a device ColorTable index rather than an RGB
-        // request, even when another usage flag is also present.
-        // Inside Macintosh: Advanced Color Imaging (1995), pp. 1-12--1-13.
+    fn activatepalette_installs_explicit_tolerant_entries_at_their_index() {
+        // PURE pmExplicit identifies a device ColorTable slot without an
+        // RGB request, but pmExplicit COMBINED with pmTolerant pins the
+        // palette's color AT that exact slot -- the combination
+        // full-screen applications use to install a whole custom palette
+        // in known positions (verified against real system software:
+        // the 256 explicit-tolerant entries land on the device; skipping
+        // them leaves the standard table and pre-indexed artwork
+        // displays through the wrong colors).
+        // Inside Macintosh: Advanced Color Imaging (1995), pp. 1-12--1-14.
         let (mut d, mut cpu, mut bus) = setup();
         let window = 0x0020_4160u32;
         let usage = super::PM_TOLERANT | super::PM_EXPLICIT;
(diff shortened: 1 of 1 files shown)


rlanday:
Civ II re-verification attempt:

**The palette question can't be answered yet** — on *current master* (no changes from this PR), Civ II fails during startup with its own error dialog ("Sorry, an error has occurred. (alert.c, 179)") before any palette activity: the trap log shows folder-resolution trouble (repeated `FSpGetFInfo("")` with empty names, then `FSpGetFInfo("Startup") -> fnfErr`, `"Router" -> fnfErr`) even though the archive contains the full install — so this looks like .sit directory-structure seeding, possibly interacting with the lazy-resource work, and it postdates this PR's base entirely. With this PR + #693 + #694 applied the failure is byte-for-byte identical (so nothing here regresses it), with one visible difference: **the error dialog renders complete — including its Exit button — with #694 applied; without it the button is missing** (the same mid-layout-hang class as the newspaper close box).

Since #610 was verified against Civ II on Aug 17 and master now can't reach its menus (with this packaging, at least), the keep-device-vs-install question stays open empirically. Happy to re-run the moment Civ II boots on master again — and the tolerance-aware variant offer stands if it then shows a regression. `SYSTEMLESS_PREFER_POWERPC=1` reaches a different state (white screen at 40M instructions, no error dialog) for completeness. — Claude Fable 5

benletchford (approved):
The explicit-tolerant path now installs the supplied palette colors as documented, while retaining exact-device behavior and focused regression coverage. The cancelled Linux jobs were dependency-install timeouts; other platforms passed and the combined integration suite is green.

PR #748, opened 2026-08-22
fix: run Blue's ABC Time Activities

### LINKED ISSUE #747: Blue's ABC Time Activities exits after a false color-depth warning
## Observed behavior

Launching the Blue's ABC Time Activities demo reaches a warning that the display should be set to 256 colors. Accepting the warning exits the application before the CD-ROM speed test or intro. The configured main display is already 800×600 at 8-bit indexed color.

## Expected behavior

The demo should accept the existing 256-color display, run its CD-ROM speed test, and continue into the Nickelodeon intro and playable demo.

## Reproduction

1. Launch `Blue'sABCTimeDemo` from the demo archive with a 16 MiB application partition and an 800×600, 8-bit display.
2. Allow startup to run for about 60 ticks.
3. Observe the 256-color warning and subsequent exit.

## Oracle evidence

A System 7.5.3 BasiliskII run with the same archive, memory allocation, display dimensions, and color depth does not show the warning. It opens the auxiliary resource file, runs the CD-ROM speed test, and reaches the Nickelodeon intro. The Systemless pre-dialog trap sequence contains no Graphics Device depth query; its final capability probes are Component Manager and Code Fragment Manager Gestalt selectors, so startup capability handling remains the leading divergence under investigation.

benletchford:
## Summary

- report Component Manager 3.0 through `Gestalt('cpnt')`, matching the implemented dispatch surface
- provide the standard System 7 window color table and resolve negative-index volume queries by full path, volume reference, or working-directory reference
- initialize standard zoom-window state so the game expands its window to the intended 800×600 display
- position the complete window structure for System 7 alert-position constants
- preserve the affected `DialogPtr` from `DialogSelect` update and activate events while retaining its `FALSE` Boolean result
- cover every compatibility correction with focused unit tests

## Generic root causes

The startup color warning was caused by an understated Component Manager capability response. After that gate, the application required the standard `wctb` resource, Finder-compatible volume lookup semantics, and initialized standard zoom-window state to locate its companion data and establish the intended display.

The pause loop exposed two Dialog Manager defects. Alert-position constants were centering only the content rectangle rather than the complete WDEF structure, and `DialogSelect` discarded the affected dialog pointer on handled update/activate events. System 7 preserves that pointer while returning `FALSE`; the application retains it during its movable-modal loop. Both behaviors are implemented generically following *Macintosh Toolbox Essentials* (1992), pp. 4-125–4-126 and 6-139–6-143.

## 68K compatibility result

The checksum-pinned 68K `CODE 0` executable now completes its CD-ROM check and intro, enters the demo’s only playable activity (Bathtime Rhymes), accepts a rhyme-bubble selection, produces non-silent response audio, pauses and resumes, exposes and applies Game-menu music and volume options, plays the Quit outro, and exits normally through `ExitToShell`. The demo offers no save/load, restart, scenario-selection, or ending route. The archive’s incomplete native PowerPC path is outside this explicitly 68K issue.

Fresh Systemless and System 7.5.3 BasiliskII runs used the same archive and deterministic inputs. Canonical before/resume/continued gameplay frames scored 97.37%, 97.97%, and 96.70% strict pixel match (97.35% overall); the menu-options route scored 97.89%; and the Quit route scored 95.70% followed by a pixel-exact outro frame. Two clean Systemless runs of every final route produced byte-identical artifacts. The meaningful bubble action produced 1,516,490 non-silent samples with no callback or runtime error.

## Validation

- `cargo fmt --all -- --check` — compatibility diff clean; current `master` has one pre-existing format-only diff in untouched `src/trap/menu.rs`
- `cargo test --lib` — 3,831 passed, 3 ignored
- `cargo test --lib --features test-support` — 3,837 passed, 3 ignored
- `cargo check --target wasm32-unknown-unknown --no-default-features`
- `cargo package` verification
- focused Component Manager, window-color-table, volume-selection, zoom-window, alert-position, and `DialogSelect` tests
- deterministic startup/intro, activity, bubble/audio, pause/resume, Game-menu settings, Quit/outro, and normal-exit routes under both 68K runtimes

Closes #747

### DIFF
--- a/src/trap/dialog.rs
+++ b/src/trap/dialog.rs
@@ -3999,15 +3999,6 @@ impl super::TrapDispatcher {
         // Macintosh Toolbox Essentials (1992), pp. 4-29, 4-31, 4-55,
         // and 4-124 to 4-126.
         let main_screen = (menu_bar_height, 0, screen_h, screen_w);
-        let place_at_alert_position = |target: (i16, i16, i16, i16)| -> (i16, i16, i16, i16) {
-            let dialog_w = bounds.3 - bounds.1;
-            let dialog_h = bounds.2 - bounds.0;
-            let target_w = target.3 - target.1;
-            let target_h = target.2 - target.0;
-            let new_left = target.1 + (target_w - dialog_w) / 2;
-            let new_top = target.0 + (target_h - dialog_h) / 5;
-            (new_top, new_left, new_top + dialog_h, new_left + dialog_w)
-        };
         // Parent-relative placement positions the complete dialog structure,
         // then converts the result back to the content bounds used by NewWindow.
         // Macintosh Toolbox Essentials 1992, pp. 4-125 to 4-126 and 6-62 to 6-63
@@ -4050,7 +4041,7 @@ impl super::TrapDispatcher {
             // alertPositionMainScreen / ParentWindowScreen
             // Macintosh Toolbox Essentials 1992, pp. 4-125 to 4-126
             0x300A | 0x700A => {
-                bounds = place_at_alert_position(main_screen);
+                bounds = place_structure_at_alert_position(main_screen);
             }
             // alertPositionParentWindow uses the content rectangle of the
             // window in which the user was last working.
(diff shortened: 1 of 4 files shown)


benletchford:
A fresh native PowerPC reference run of the exact public archive changes the diagnosis: Mac OS 8 also stops at startup with the message that it cannot open the auxiliary resource file. The native bootstrap requests Blue'sABCTimeDemo (u), but the published archive contains only the (0), (2), (4), (a), and (r) companions. The archive SHA-256 is 7062f23be117c8fd44f902ba1e6131a13391c4a2dc357d617abbf5a820f09a73. The earlier BasiliskII result exercised the fat application's 68K path and therefore could not validate the native PPC path. Because the selected public archive cannot reach gameplay in the native PPC reference either, this draft cannot meet the issue's compatibility objective and should not land as a partial fix.

PR #838, opened 2026-08-26
fix: complete classic manager compatibility paths

### LINKED ISSUE #770: fix: complete classic manager compatibility paths
## Problem

A classic application runtime reaches several generic Toolbox paths that Systemless currently models only partially:

- Notification Manager accepts `NMInstall` but never executes `nmResp`, leaving applications waiting forever.
- application-defined controls do not receive native `testCntl` / `drawCntl` callbacks during `FindControl` and `TrackControl`;
- popup tracking does not publish the Menu Manager choice consumed by `MenuChoice`;
- high-bit CDEF IDs are sign-extended instead of decoded as a 16-bit bit field;
- variable-sized MENU resources can be truncated;
- full HFS directory pathnames are not resolved by `PBGetCatInfo`;
- System-file pattern resources and ROM-shaped post-instruction exception vectors are missing;
- repeated `SelectWindow` and application-drawn dialog composition can cause destructive redraws;
- active control labels can resolve against a stale offscreen palette.

Together these prevent a representative 68K application from progressing from startup through configuration, real gameplay, its completed state, and its detailed report.

All fixes must remain generic Toolbox/runtime behavior with no title-specific logic.

## Acceptance criteria

- Validate, queue, invoke, automatically complete, and explicitly remove Notification Manager requests using documented conventions.
- Dispatch application CDEF hit testing and tracking callbacks, including high-bit CDEF resource IDs.
- Preserve popup `MenuChoice` state and complete MENU resource records.
- Resolve full HFS directory pathnames in `PBGetCatInfo`.
- Supply standard PAT# ID 0 and ROM-shaped exception-vector handlers.
- Preserve application-drawn dialog compositions and make already-front `SelectWindow` idempotent.
- Resolve direct-framebuffer control labels through the live device palette.
- Add focused regression coverage for every corrected path.
- Keep the native test suite and WebAssembly library check green.

benletchford:
## Summary

- execute Notification Manager response procedures with documented queue and removal behavior
- dispatch application-defined control hit testing and tracking callbacks, including popup MenuChoice state and high-bit CDEF IDs
- preserve complete menu resources, application-drawn dialogs, and already-front window selection
- synthesize standard System patterns and ROM-shaped exception handlers
- resolve full HFS catalog paths and direct-framebuffer control labels against the live palette
- add focused regression tests for every corrected compatibility path

## Validation

- `cargo test`
- `cargo check --target wasm32-unknown-unknown --lib`
- two clean deterministic 68K completion runs with identical checkpoint hashes
- the same 122-action route under BasiliskII, covering setup, gameplay, accelerated completion, final state, box-score options, and the detailed report

Closes #770

### DIFF
--- a/src/memory/bus.rs
+++ b/src/memory/bus.rs
@@ -13,12 +13,18 @@ use super::globals::LowMemGlobals;
 const LEGACY_SOUND_BUFFER_WORDS: u32 = 370;
 const LEGACY_SOUND_BUFFER_BYTES: u32 = LEGACY_SOUND_BUFFER_WORDS * 2;
 const SYNTHETIC_RESERVE_BYTES: u32 = 64 * 1024;
+// Keep the boot-ROM shadow deliberately narrow: bytes outside these witnessed
+// or synthetic HLE words retain the bus's existing unmapped-zero behavior.
 // System 7.5.3 on the Quadra 650 leaves exception vector 0 pointing to
-// $40810000. A BasiliskII oracle capture of that ROM establishes the word at
-// offset 6 as $0372. Keep the shadow deliberately narrow: bytes outside this
-// witnessed range retain the bus's existing unmapped-zero behavior.
-const BOOT_ROM_SHADOW_BASE: u32 = 0x4081_0006;
-const BOOT_ROM_SHADOW: [u8; 2] = 0x0372u16.to_be_bytes();
+// $40810000; a BasiliskII oracle capture establishes the word at offset 6 as
+// $0372. The two RTE words let the runner retain ROM-shaped exception-vector
+// values while safely resuming after the post-instruction exceptions it
+// handles (Zero Divide, CHK, and TRAPV).
+const BOOT_ROM_WORD_SHADOWS: &[(u32, u16)] = &[
+    (0x4080_26F8, 0x4E73),
+    (0x4080_26FA, 0x4E73),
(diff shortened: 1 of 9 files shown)


benletchford:
This is not merge-ready against current master. The branch now conflicts in the control path, and master has since landed substantial classic window/menu behavior plus shared cross-architecture runtime and memory changes that overlap the nine-subsystem scope here.

Please rebase onto current master and reduce the PR to the compatibility work that is still missing after those releases. If the remaining delta is still broad, split it into focused issue/PR units so each documented Toolbox behavior can be reviewed and released independently. Then rerun the focused trap tests, full suite, wasm check, deterministic route, and BasiliskII comparison. I am leaving #770 open and not merging the stale conflict resolution.

benletchford:
I attempted a rebase onto current master in an isolated checkout. This is not a mechanical realignment: it conflicts in globals, runner, controls, dispatcher and menus. In particular, the incoming runner code restores ActiveInterruptCallback/active_interrupt_callback-era scheduling while current master uses the shared guest-call machinery and callback_suspends_guest_clock; the MENU changes also overlap current complete-resource and nested-MDEF handling. Taking either side wholesale would lose behavior. Please port only the still-missing notification/control behaviors to the current task model and remove superseded menu changes, with focused regressions and a fresh #770 route/oracle comparison. I left the public branch unchanged and the PR open.

PR #853, opened 2026-08-26
perf(render): row-bulk the 8bpp draw primitives

rlanday:
From Claude Fable 5, working with @rlanday.

## What

Five 8-bit framebuffer primitives made one bus call per pixel; they now make one bulk call per row through the slice fast paths the bus already has (`read_bytes_into`, `write_bytes`, `fill_bytes` — one address translation and one `memcpy`/`memset` per call, the same paths `BlockMove` and the same-depth window blit already use):

- `save_screen_rect_pixels` (8bpp): one `read_bytes_into` per row instead of a `read_byte` per pixel. The packed 2/4bpp path keeps its per-pixel unpacking.
- `restore_screen_rect_pixels` (8bpp): one `write_bytes` per on-screen row; off-screen columns are clipped and a truncated pixel buffer stops exactly where the per-pixel loop stopped.
- `fb_hline` (8bpp): one `fill_bytes`.
- `blit_window_to_screen`: the 8bpp palette-translation loop and the generalised packed-depth loop (any 1/2/4/8-bit port to any 1/2/4/8-bit screen, from 17fb5af) each read the source row's span once — and, for a packed destination, the destination span the pixels merge into — repack against host buffers, and write the row once.

Every bus call pays address translation, the read-only-code check and the idle-probe journal check, so a 560-pixel title-bar border was 560 full-cost calls to store one constant. `fb_fill_rect`, `fb_fill_pattern_rect` and `restore_dialog_pixels` were already row-bulk; this brings the rest of the 8bpp family in line. No caller changes, no behavioural change: the conversions write the same guest bytes in the same order per row.

## Why it matters

`redraw_chrome` runs after every host frame. SimCity 2000 plays with the menu bar visible (MBarHeight stays 20, its main window sits at left=40, so the fullscreen gates never fire), which means the menu-bar repaint and the per-window save/paint/restore pass run every frame, all through these primitives. In a steady-state profile of the game (below) `save_screen_rect_pixels` and `redraw_chrome` (with `restore_screen_rect_pixels` inlined into it) were 2.9% and 3.3% of the main thread's self time, and `MacMemoryBus::write_byte` — mostly their per-pixel stores — 8.7%. Guest `CopyBits` into an offscreen window port, dialog restores and drag outlines go through the same primitives.

## Evidence

[...]

PR #857, opened 2026-08-26
perf(memory): reuse the write-probe journal and hash it by address

rlanday:
From Claude Fable 5, working with @rlanday. Independent of #853 (measured on top of it).

## What

The idle-cycle write probe journals the original byte of every address written while it is armed — the journal is what proves a candidate wait cycle left guest RAM unchanged. It was a `HashMap<u32, u8>` with the default SipHash, created fresh for every probe.

Two changes, no semantic change:

- **Hash by address.** A `Hasher` for the journal's `u32` keys: a multiplicative mix folded so the high address bits reach the low hash bits (hashbrown takes its tag from the top bits and its bucket from the low ones). SipHash's keyed, DoS-resistant mixing buys nothing for a map keyed by guest addresses that never leaves the process.
- **Keep one map.** The journal is parked when a probe is cancelled, finished or voided, and cleared when the next probe arms, so its table capacity survives instead of regrowing from empty (`reserve_rehash`) every time.

Same distinct-address cap (`WRITE_PROBE_MAX_ENTRIES`), same overflow void, same final byte-for-byte comparison; a new test checks that a reused journal starts empty after a finished and after a cancelled probe.

## Why it matters

On the SimCity 2000 headless run the probe arms about **1.3 million times per 2 billion instructions**, and while it is armed every guest store goes byte-wise through `write_byte` — `write_long` → `write_word` → `write_byte`, each byte a journal insert. In a steady-state profile (below) the journal's `hash_one`, SipHash `write` and `reserve_rehash` were about 4 % of the main thread's samples, all reached from `write_byte`.

## Evidence

**Instrument.** Headless SimCity 2000 (`--headless --max-instructions N --input-script`, the #824 input replay into a running city), identical guest work per arm — checked by the guest tick count printed at exit, identical in every run below. Metric: host **instructions retired** for the whole process from `/usr/bin/time -l` (load-insensitive, unlike wall or CPU time). Six pairs per run length, arms interleaved with alternating order; the figure is the geometric mean of the within-pair ratios, with the pair range and pairs won.

**Arms.** Base = upstream master 0.20.5 (1013a0e) + d2357f7 (a local runner fix carried in *both* arms: wait-identity proofs compare architectural CPU state only) + #853's row-bulk primitives; candidate = the same plus this commit.

The PR is rebased onto master 17fb5af; upstream's changes since the measured base do not touch the probe or the journal (the only `bus.rs` change is in `configure_screen_depth`).

| SimCity 2000 | base | this PR | Δ | pairs won |
|---|---|---|---|---|
| 400M instructions | 114.97 G | 110.08 G | **−4.25 %** (ratios 0.957–0.958) | 6/6 |
| 2B instructions | 504.27 G | 483.63 G | **−4.09 %** (0.958–0.960) | 6/6 |

Guest ticks 128,460 / 283,023 in all 24 runs. Identity: the 261 screenshots of a 130M-instruction scripted run (one per 500K instructions) are byte-identical between the arms — the journal's verdicts, and therefore which cycles the runner parks, are unchanged.

**Profile** (`/usr/bin/sample` 1 ms × 5 s at 45 s of a 2B run, main thread self time, 3,431 samples, base arm): `hash_one` 2.1 %, `reserve_rehash` 1.0 %, SipHash `Hasher::write` 0.9 % — call-graph attribution puts 96 of the 128 `reserve_rehash` samples and all of the SipHash samples under `MacMemoryBus::write_byte`, i.e. the probe journal; the remaining allocator time under `reserve_rehash` (`nanov2_free`, 11 samples) is the per-probe table growth this removes.

## Tests

- `cargo test --lib --features test-support` on this branch (master 17fb5af): 4,210 passed, 0 failed, 3 ignored (the existing `write_probe_*` tests cover accept/reject/overflow; the new `write_probe_journal_is_reused_without_stale_entries` covers reuse after finish and after cancel).
- `cargo fmt --check` clean on the changed lines (the file's one pre-existing diff, in `configure_screen_depth`, is untouched). `cargo clippy --all-targets --all-features`: no finding inside the changed lines (the crate's pre-existing findings are unchanged).

### DIFF
--- a/src/memory/bus.rs
+++ b/src/memory/bus.rs
@@ -4,6 +4,7 @@
 
 use std::cell::{RefCell, UnsafeCell};
 use std::collections::HashMap;
+use std::hash::{BuildHasherDefault, Hasher};
 use std::rc::Rc;
 use std::sync::atomic::{AtomicBool, AtomicU32, Ordering};
 #[cfg(not(target_arch = "wasm32"))]
@@ -493,7 +494,11 @@ pub struct MacMemoryBus {
     /// journal with final RAM proves whether a candidate execution cycle left
     /// guest memory unchanged, while still allowing temporary stack writes
     /// that are restored before the cycle closes.
-    write_probe_original: Option<HashMap<u32, u8>>,
+    write_probe_original: Option<WriteProbeJournal>,
+    /// The journal's allocation between probes. Probes start more than a
+    /// million times in a long SimCity 2000 session; reusing one map keeps
+    /// the table's capacity instead of regrowing it from empty each time.
+    write_probe_spare: WriteProbeJournal,
     /// An out-of-range write makes the probe unverifiable even though the
     /// normal bus retains its legacy warn-and-ignore behavior.
     write_probe_invalid: bool,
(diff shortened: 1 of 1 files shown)


benletchford:
Maintainer update: I removed my earlier workflow-blocker comment because it incorrectly applied the project-authored issue/dev-branch convention to an outside contribution. I rebased this through the shared-RAM conflict while preserving both the current address-space implementation and the journal optimization, then verified the full library suite and green cross-platform CI before squash-merging.

PR #860, opened 2026-08-26
perf(render): row-blit 8bpp text and column-fill vertical borders

rlanday:
From Claude Fable 5, working with @rlanday. **Stacked on #853** (this branch includes #853's commit until it merges; measured on top of it).

## What

#853 made the row-shaped framebuffer primitives one bus call per row. What was left in the chrome painters was not row-shaped, and it was slow for a different reason: **`fb_set_pixel(…, black)` re-resolves the colour on every pixel** — `active_gdevice_ctab` (TheGDevice → GDevice → PixMap → colour table, five dependent bus reads) plus the luma lookup — and in the profile those two were three quarters of `fb_set_pixel`'s time; the write itself was the minority.

- **Text.** A new 8-bit glyph painter, `fb_blit_glyph_rows_8bpp`, paints a glyph one row at a time: one bulk read of the covered destination span, the glyph's set pixels (coverage ≥ 128) applied against the host buffer, one bulk write; rows with nothing in view never touch the bus. Three paths take it: the plain painter (`fb_draw_glyph_bitmap_with_slant`, for unslanted and unstretched glyphs), the clipped painter used for window titles (which now resolves the colour once per string instead of once per pixel), and the styled painter when no per-glyph style bit is set — bold, italic, outline, shadow, condense or extend — where the pixel set it would have built is exactly the glyph bitmap.
- **Vertical lines.** `fb_vline` resolves the colour once and writes the column with one strided bus fill. `MemoryBus::fill_bytes_strided` takes one address translation and bounds check for the whole span, then a stride loop over the RAM slice; it falls back to byte writes — which skip exactly the protected bytes and journal each write — when the span touches read-only code, an idle-probe write journal is armed, or a tracer is active. `draw_window_chrome` (title-bar sides, close-box edges, content borders, shadow column), `draw_thick_rect_border`, `draw_grow_icon` and the dialog `draw_rect_border` use it.

Other depths, slanted or stretched glyphs, and styled text keep their existing per-pixel paths unchanged.

## Why it matters

`redraw_chrome` runs after every host frame; with the menu bar visible it repaints the menu-bar titles and every window's title bar and borders each frame. After #853, in a steady-state profile of SimCity 2000, the remaining chrome pass was 9.7 % of the main thread's samples: window title text 2.2 %, the vertical borders/shadow/close box 1.8 %, menu-bar title text 2.9 %, with `fb_set_pixel`'s colour resolution the bulk of each. Dialog text and borders, menus and controls go through the same painters.

## Evidence

**Instrument.** Headless SimCity 2000 (`--headless --max-instructions N --input-script`, the #824 input replay into a running city); identical guest work per arm, checked by the guest tick count at exit. Metric: host instructions retired for the whole process (`/usr/bin/time -l`; load-insensitive). Six interleaved, order-alternated pairs per run length; geometric mean of the within-pair ratios, with range and pairs won.

**Arms.** Base = upstream master 0.20.5 (1013a0e) + d2357f7 (a local runner fix carried in both arms) + #853; candidate = the same plus this commit. The PR is rebased onto master 17fb5af on top of #853's rebased commit; it applied without conflicts (17fb5af's `blit_window_to_screen` rewrite does not touch the text painters or the border loops).

| SimCity 2000 | base (#853) | this PR | Δ | pairs won |
|---|---|---|---|---|
| 400M instructions | 114.92 G | 108.09 G | **−5.94 %** (ratios 0.939–0.942) | 6/6 |
| 2B instructions | 504.11 G | 470.70 G | **−6.63 %** (0.933–0.935) | 6/6 |

Guest ticks 128,460 / 283,023 in all 24 runs.

**Identity.** The 261 screenshots of a 130M-instruction scripted run (one per 500K instructions) are byte-identical between the arms (ticks 102,146 in both): the row painter, the column fill and the once-per-string colour resolution write exactly the bytes the per-pixel painters wrote.

**Profile.** `/usr/bin/sample` at 1 ms for 5 s, attached 45 s into a 2B run, main thread, share of self-time samples (the headless process is never idle). Base arm (#853), 3,431 samples: `redraw_chrome` 9.7 % inclusive (window pass 4.8 %, menu bar 4.1 %), `fb_set_pixel` 4.6 % inclusive with `active_gdevice_ctab` + `ctab_value_luma` — the per-pixel colour resolution — 76 % of that. This PR, 3,122 samples: `redraw_chrome` **3.5 %** inclusive (menu bar 2.0 %, window pass 0.6 %, save/restore 0.5 %); `fb_set_pixel`, `ctab_value_luma` and `active_gdevice_ctab` no longer appear in the top-of-stack list; the new row painter is 0.5 % and `memmove` under `save_screen_rect_pixels`/the glyph rows 0.5 %.

## Tests

Seven new tests: the row painter equals the per-pixel painter byte for byte, plain and clipped, including glyphs partly off every screen edge and a clip cutting through the glyph boxes; style 0 through the styled painter equals the plain painter; `fb_vline` equals the per-pixel loop with clipping on every side; the strided fill writes only its column, skips read-only bytes exactly like byte writes, and is journaled by a write probe. The existing chrome tests assert specific pixels of the frames these painters draw and are unchanged.

- `cargo test --lib --features test-support` on this branch (master 17fb5af + #853): 4,216 passed, 0 failed, 3 ignored (4,145 on the measured base).
- `cargo fmt --check` clean on the changed lines (`bus.rs` keeps its one pre-existing diff in `configure_screen_depth`). `cargo clippy --all-targets --all-features`: no finding inside the changed lines.

### DIFF
--- a/src/memory/bus.rs
+++ b/src/memory/bus.rs
@@ -449,6 +449,16 @@ pub trait MemoryBus {
             self.write_byte(address.wrapping_add(i), value);
         }
     }
+
+    /// Write `value` to `count` bytes spaced `stride` apart starting at
+    /// `address` — one column of an 8-bit framebuffer. Default impl is a
+    /// byte-by-byte loop; the [`MacMemoryBus`] override takes one bounds
+    /// check for the whole span.
+    fn fill_bytes_strided(&mut self, address: u32, stride: u32, count: u32, value: u8) {
+        for i in 0..count {
+            self.write_byte(address.wrapping_add(i.wrapping_mul(stride)), value);
+        }
+    }
 }
 
 /// Flat guest RAM with low-memory globals, allocation state, and diagnostics.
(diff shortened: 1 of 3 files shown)


benletchford:
Maintainer update: I removed my earlier workflow-blocker comment because it incorrectly applied the project-authored issue/dev-branch convention to an outside contribution. After #853 landed, I restacked this to its single child commit and rebased it through the later memory changes, preserving both probe and strided-fill coverage. The fully combined suite and cross-platform CI passed before squash merge.

PR #861, opened 2026-08-26
perf(memory): journal write probes by 32-bit unit and keep bulk stores on their fast paths

rlanday:
From Claude Fable 5, working with @rlanday. **Stacked on #857** (this branch includes #857's commit until it merges; measured on top of it).

## What

While an idle-cycle write probe is armed, the journal records the original value of everything the guest writes so the proof can check that RAM was restored. It was keyed by byte: one hash insert per byte written, and every multi-byte or bulk store fell back to a byte loop to feed it.

Two changes:

- **Journal by aligned 32-bit word.** The journal maps a word address to the word's original contents; a write journals the words it covers — one insert per word instead of one per byte. The bytes of a word the write did not touch cannot change while a probe is armed (every writer goes through the bus and fastmem is withdrawn), so comparing them at the end is harmless; if anything ever did change one, the proof fails, which is the safe direction. The 4,096-entry cap now counts words, so a probe overflows a little later than before — the cap bounds the journal's cost, and a word journal of 4,096 entries costs what a byte journal of 4,096 entries did.
- **Armed-probe fast paths.** When the probe is the only thing keeping a store off its fast path (no framebuffer-write tracer, no watchpoint), `write_word`/`write_long`, `write_bytes`, `fill_bytes`/`fill_zeros`, `block_move`, `copy_ram_bytes` and `copy_mapped_ram_bytes` journal their range and take the same in-bounds bulk write they take when no probe is armed. Read-only overlap, tracers and watchpoints keep the byte path, which skips exactly the protected bytes and lets the diagnostics see every byte.

## Why it matters

On the SimCity 2000 headless run the probe is armed for most of the wait loop and the m68k core's MOVEM, LINK and other complex ops store through the bus. After #857 removed the hashing and allocation cost, the remaining per-byte insert chain under `write_long` was still about 4 % of the main thread's samples; a first attempt that kept byte-granular journaling but removed the redundant bus passes measured nothing, which is how the insert itself was identified as the cost.

## Evidence

| SimCity 2000 | base (#853 + #857) | this PR | Δ | pairs won |
|---|---|---|---|---|
| 400M instructions | 110.05 G | 109.52 G | **−0.48 %** (ratios 0.995–0.996) | 6/6 |
| 2B instructions | 483.75 G | 481.65 G | **−0.44 %** (0.995–0.996) | 6/6 |

Guest ticks 128,460 / 283,023 in all 24 runs — the cap now counting words changed no probe outcome on this workload. Small but consistent; the remaining cost of an armed probe is the withdrawn fastmem window (the interpreter's reads go through the bus accessors) and the probes cancelled by a trap before they can prove (57 % of the 228k starts per 400M run), not the journal inserts.

**Identity.** The 261 screenshots of a 130M-instruction scripted run (one per 500K instructions) are byte-identical between the arms (ticks 102,146 in both).

## Tests

- `cargo test --lib --features test-support`: 4,141 passed, 0 failed, 3 ignored on the measured base; 4,212 passed, 0 failed on the PR branch (master 17fb5af + #857 + this commit). New tests: every write size and bulk path is journaled while a probe is armed (a changed byte is seen, a restored one is accepted); the armed-probe fast paths leave RAM byte-identical to the byte-at-a-time path, including overlapping `block_move`s in both directions and a long that crosses a read-only byte; the cap and reuse tests are updated for word entries. `cargo clippy --all-targets --all-features`: no finding inside the changed lines.

### DIFF
--- a/src/memory/bus.rs
+++ b/src/memory/bus.rs
@@ -508,9 +508,9 @@ pub struct MacMemoryBus {
     write_probe_overflowed: bool,
 }
 
-/// An exact-state probe journals the bytes an idle cycle writes: a spilled
+/// An exact-state probe journals the words an idle cycle writes: a spilled
 /// register or two, an event record, a saved keymap -- tens of bytes, a few
-/// hundred at most. A journal past this many distinct addresses is not
+/// hundred at most. A journal past this many distinct words is not
 /// watching a wait; it is watching real work, and while it stays open every
 /// guest store pays a hash insert and fastmem is suspended for the whole
 /// core. Past the cap the probe is voided on the spot so normal fast paths
@@ -654,8 +654,12 @@ impl Hasher for AddressHasher {
     }
 }
 
-/// Original bytes of the addresses written during an exact-state probe.
-type WriteProbeJournal = HashMap<u32, u8, BuildHasherDefault<AddressHasher>>;
+/// Original contents of the aligned 32-bit words a probe's writes touched,
+/// keyed by the word's address. One insert per word instead of one per
+/// byte; the bytes of a word a write did not touch cannot change while a
+/// probe is armed (every writer goes through the bus and fastmem is
+/// withdrawn), so comparing them at the end is harmless.
+type WriteProbeJournal = HashMap<u32, u32, BuildHasherDefault<AddressHasher>>;
 
 /// RAM storage - either a stable owned allocation or borrowed slice.
 enum RamStorage {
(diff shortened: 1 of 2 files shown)


rlanday:
The macOS leg failed with `The hosted runner lost communication with the server` — an infrastructure drop, not a test failure (ubuntu, windows, headless & package, licenses and fmt/clippy all passed, and the suite is green locally on macOS: 4,212 passed, 0 failed). Reopening to re-trigger the checks.

benletchford:
Maintainer update: I removed my earlier workflow-blocker comment because it incorrectly applied the project-authored issue/dev-branch convention to an outside contribution. After #857 landed, I restacked this to its single child commit on current master, verified the probe/bulk-path suite and green cross-platform CI, and squash-merged it.

PR #962, opened 2026-08-29
perf(runner): admit journal-complete Window Manager queries and TEIdle to idle-cycle proofs

### LINKED ISSUE #1052: Idle-park exact-memory gate must not be blinded by journal-suspended chrome repaints
Focused issue for the correctness boundary raised in #962 review.

#962 keeps host chrome repaints out of an armed idle-proof journal (`redraw_chrome_outside_idle_journal`): the per-frame repaint would otherwise flood and overflow the bounded journal at 8 bpp, cancelling every proof. Suspension is byte-safe on every frame except one case: the **first repaint after a park** where guest code overwrote a chrome pixel before entering the idle loop. The park's proof captured guest RAM with the guest's pixel value; the suspended repaint then restores the chrome value — a guest-visible framebuffer mutation the exact-memory resume gate never sees, after which the parked cycle is replayed against memory that differs from its proven precondition.

Fix direction (per review): make a changed repaint revoke/re-prove the park — e.g. a single-flag byte-change check on the bus write path, armed only while a park is live (no journal-capacity concerns), with a regression test for the scribbled-chrome-then-park sequence and a control test that a byte-identical repaint keeps the park. Repaints during the probing phase (pre-park) need no check: proofs re-execute the full cycle, so a changed repaint simply fails to prove.

https://claude.ai/code/session_01LnuSFJRzTWFhxrFUMw622y

rlanday:
## Summary

SimCity 2000's wait loops call `LocalToGlobal`, `FrontWindow`, `FindWindow`, `GetWRefCon` and `TEIdle` on every iteration. None of them was admitted to the exact idle-cycle prover, so a loop that had nothing to do still executed a full tick's instruction budget to earn each tick. This PR admits those five traps, with the argument for each written out below, and fixes a frame-boundary problem the admissions expose in the GUI. Two commits:

1. `fix(runner): keep host chrome repaints out of an armed idle-proof journal`
2. `perf(runner): admit journal-complete Window Manager queries and TEIdle to idle-cycle proofs`

Headless SimCity 2000 on master 0.23.8 (m68k 0.11.6), 2 billion instructions, three runs per arm:

| | master 298bf2e | + Window Manager queries | + TEIdle (this PR) |
|---|---|---|---|
| host instructions retired | 394.76 G | 358.62 G | 356.82 G |
| guest ticks reached | 283,023 | 304,960 | **355,838 (+25.7 %)** |
| **host instructions per guest tick** | 1,394.8 k | 1,176.0 k (−15.7 %) | **1,002.7 k (−28.1 %)** |

GUI results are in the *Evidence* section.

### How the prover decides a loop is idle

Two independent gates; a bad admission would have to beat both.

**Gate 1 — the proof** (`try_exact_idle_cycle_fastfwd`). Two arrivals at the same trap site in the same guest tick must show an identical `CpuArchitecturalSnapshot` *and* a write journal in which every guest-RAM byte written during the cycle still holds its original value. A trap is *journal-complete* if every consequence of calling it lands in CPU registers or guest RAM, because then any real difference between the two iterations shows up in one of those two comparisons.

**Gate 2 — the park** (`try_resume_proven_idle_cycle`). A proven cycle is not replayed blindly. `park_proven_idle_cycle` arms a second write journal while parked, and the cycle may be reused only while guest memory, the CPU snapshot, the guest tick and `IdleCycleHostSnapshot` are unchanged and the Event Manager has nothing to deliver. A park advances only to the next guest tick, the GUI wall-clock cap, or an interrupt — never further — so every tick still runs the two real iterations a proof needs.

The dangerous shape is therefore narrow: **a trap that reads host-side state which can change while the guest is stopped, is not covered by the host snapshot or the event stream, and whose changed answer the guest would have acted on.** The rustdoc on `idle_cycle_trap_is_journal_complete` now states this as the admission checklist.

### The five traps and where their inputs and outputs live

| trap | reads | writes | why it is journal-complete |
|---|---|---|---|
| `LocalToGlobal` $A870 | A7, the `Point` (RAM), A5 → QuickDraw globals → `thePort` → port bounds (RAM) | the `Point` in place, A7 | Guest RAM and registers only; the mirror image of the already-admitted `GlobalToLocal`. |
| `GetWRefCon` $A917 | A7, `window+152` (RAM) | the result long on the stack, A7 | Guest RAM only. Any writer of `refCon`, guest or HLE, writes through the bus, so the journal sees it. |
| `FrontWindow` $A924 | `dispatcher.window_list` (host mirror), `GhostWindow`, per-window visibility bytes (RAM) | the result on the stack, A7 | The mirror is rewritten into the guest window chain (journaled) by every mutation; see below. |
| `FindWindow` $A92C | A7, the `Point` and VAR pointer (RAM), `MBarHeight` (RAM), `dispatcher.pending_native_menu_selection` (host), `window_list` (host), window records (RAM) | the window through the VAR parameter, the part code as the result, A7 | Both host inputs are covered: see below. |
| `TEIdle` $A9DA | A7, the TERec (RAM), the current tick | either nothing, or `caretState`/`caretTime` (RAM) and then the caret/text pixels (RAM) | Writes nothing, or stamps guest RAM *before* any draw; see below. |

### Why each admission is sound

**LocalToGlobal / GetWRefCon.** Pure functions of guest RAM and registers. Nothing to argue beyond the table.

**FrontWindow / FindWindow.** These read a host-side mirror of the guest window chain, so the honest statement is that their soundness is *transitive*: every mutation of `window_list` in the tree is under `src/trap/` (reachable only from guest trap dispatch — no runner or frontend path touches it), no guest code runs while parked, and every such mutation ends in `sync_window_list_links`, which rewrites the guest chain through the bus, so a proof's journal sees it. `FindWindow` additionally answers `inMenuBar` while a native menu selection is staged; that staging always sets `pending_native_menu_event` in the same branch, which `peek_toolbox_event` merges into the stream, so the park gate revokes the park. Because both of these are invariants rather than local facts, this PR (a) states them at the two field declarations in `dispatch.rs`, and (b) adds `window_list` and `pending_native_menu_selection` to `IdleCycleHostSnapshot`, so a parked cycle cannot resume across a change to either even if a future mutator forgets the chain rewrite or the paired event.

**TEIdle.** `textedit_idle` reads the TERec and the dispatcher tick, and does one of two things: returns having written nothing (record inactive, selection not an insertion point, or the 32-tick blink interval not yet elapsed), or toggles `caretState`, stamps `caretTime` with the current tick, and *then* calls `draw_te_contents`. The draw path mutates host-cached QuickDraw state the journal cannot see (`set_current_port_state`, `last_screen_frame_rect`), so the admission rests on the ordering: the two guest-RAM stamps precede any draw, and a blink that lands inside a proof therefore fails the proof on memory rather than escaping it. That ordering is now called out at the stamp site in `dialog.rs`, and `teidle_inside_a_proof_parks_when_idle_and_fails_on_memory_when_it_blinks` exercises both halves through the real handler under a real probe.

*Blink fidelity.* A park is at most one tick and every tick runs the two real iterations a proof needs, with the real tick count. While the interval has not elapsed TEIdle returns having written nothing — a genuine wait, parked exactly as the unparked loop would have spun; when it has, the stamps fail the proof and the guest runs that iteration for real. A due blink therefore fires in the same tick it fires on the real machine, and no tick is ever skipped, so caret parity is preserved. Two toggles inside one probe window — the only way to leave the final bytes equal after a blink — would need 64 ticks inside a single tick.

### Attacks considered, and what defeats them

1. *The window list changes while a loop is parked, so FrontWindow should have returned a different pointer.* No guest code runs while parked, only trap dispatch mutates the list, and the answer is never cached across a park: on resume the loop runs again and calls the trap afresh.
2. *The user picks something from the native menu bar while parked.* The strongest attack, since `FindWindow` would answer `inMenuBar`. Defeated by the paired `pending_native_menu_event` (the park gate sees a deliverable event) and now also by the host snapshot.
3. *The guest polls for a brief mouse-down inside a parked tick and misses it.* Pre-existing machinery: `mouse_button` and `mouse_pos` are in the host snapshot, so any click or movement revokes the park.
4. *A trap's answer changes but the guest discards it, so state repeats and a loop "making progress" is parked.* Not a soundness problem: if the guest discarded the value no guest-visible state depends on it, and the park ends at the next tick, so the loop is re-run with fresh answers; nothing is skipped, only re-derived.
5. *`refCon` is written between iterations.* Every writer goes through the bus; the journal sees it.
6. *TEIdle skips or desynchronises blinks.* See *Blink fidelity* above — a park never spans more than one tick.
7. *TEIdle's host-side draw effects escape the journal.* They can only occur after the two stamps, which fail the proof; the comment at `textedit_idle` marks that ordering as load-bearing.
8. *`tick_count` (dispatcher) and `Ticks` (guest RAM) diverge across a proof.* The resume gate checks both agree before extending a park, and TEIdle only ever runs in unparked execution.
9. *Diagnostics.* `FindWindow` and `TEIdle` print under trace env vars; parked ticks emit none, as for every admitted trap. Not a correctness matter, but worth knowing before diffing two trace logs.

### The fix this exposes (commit 1)

The runner repaints the host-owned chrome (menu bar, window frames, cursor) every frame after the guest slice returns. When a proven cycle is parked, its journal stays armed across that repaint, so the chrome's writes were recorded against the proof — and at 8 bpp the menu bar alone is about 4,000 journal words against `WRITE_PROBE_MAX_ENTRIES = 4096`. On any 8-bit screen the journal overflowed, the park was cancelled and the cycle re-proved from scratch on every frame. Nothing parked at 8 bpp before these admissions, which is how it went unnoticed; the headless harness never parks at all (it fast-forwards within a slice), so no headless number could have shown it.

The bus can now detach an armed journal (`suspend_write_probe`) and re-arm the same one (`resume_write_probe`), and the runner paints the chrome between the two at all four call sites. This takes nothing from the proof: the chrome's pixels change only when dispatcher state does, which happens through guest traps (never admitted into a proof) or host input (gated before a park resumes). `SYSTEMLESS_WAIT_STATS=1` now also reports `resumed` and `resume_fail_{mem,host,other}`, so park survival is observable in the GUI.

### Residual points, and what is deliberately not admitted

- The transitive invariants behind FrontWindow/FindWindow are now written down where they are enforced and mirrored in the host snapshot; a future mutator that breaks either still cannot make a park resume, only cost a re-proof.
- `find_window_at_point` reads window-record fields and `MBarHeight` — all bus reads, so covered.
- Under-inclusion, not risk: other pure window queries (`GetWTitle`, `GetWVariant`, …) are equally journal-complete by this argument and are *not* admitted; this change is measured on one workload and stops at what that workload needs.
- The neighbours that mutate host-mirrored state — `SetWRefCon`, `SelectWindow`, `ShowWindow`, `HideWindow`, `TEActivate`, `TEDeactivate`, `TEKey` — still cancel, and a test pins that, because the admission is a list of specific traps, not a range.

### Evidence

**Instrument (headless).** `--headless --max-instructions 2000000000 --input-script`, the #824 replay into a running city; host instructions retired from `/usr/bin/time -l`, guest ticks from the run log; three runs per arm, interleaved. Because the change alters how much guest time an instruction budget buys, host instructions *per guest tick* is the figure that matters. Tick counts repeat bit-for-bit across runs on every arm (spreads ≤ 0.16 % on instructions).

**Correctness (headless).** Over the full replay the arms' window/dialog/menu/resource milestone sequences are prefix-identical to master's and only *append* three lines — the window the parking arms get far enough to open. Comparing screenshots at equal guest ticks is *not* a valid check for an arm that changes instructions-per-tick, because `--input-script` schedules events by instruction count, so the arms have seen different input at the matched tick; the milestone-sequence diff over the complete replay is the sound check. `SYSTEMLESS_WAIT_STATS=1` on the arm shows the byte-exact journal still rejecting proofs (`fail_mem` > 0), i.e. gate 1 doing work rather than rubber-stamping. An admit-every-trap probe puts the ceiling near −35 % per tick; it is not what this admits.

**GUI** (hand-driven sessions: boot → new city → map, then idle; `SYSTEMLESS_WAIT_STATS=1`): on master a 232 s session used 106.6 s of CPU with `cancel_trap = 5.56 M` — 2.79 M FrontWindow and 2.78 M LocalToGlobal cancels, i.e. the dialog and menu phases spin. On this branch a **937 s** session used **88.1 s** of CPU with `cancel_trap = 221` and **51,627 parks** (master: 7,157). `probe_overflows = 0` confirms commit 1: the chrome repaint no longer overflows a parked journal. `resumed = 0 / resume_fail_mem = 49,693` over ~56 k frames means parks are re-proved about once per frame, and this is correct behaviour, not overhead to remove: `finish_audio_frame` schedules the guest’s SndDoubleBackProc callbacks (SimCity 2000 streams double-buffered sound), that guest code writes guest RAM, and the journal is right to fail the resume. The re-proof after each callback is bounded (one loop iteration plus two snapshots) and the new counters make it observable. An Instruments Time Profiler capture of a 75.7 s boot-to-map session on this branch plus the CopyBits change shows the main thread using 17.9 s (~24 %), with ~58 % of that in the m68k core and no anomaly attributable to this PR.

## Validation

- New tests: a suspended journal ignores a host-sized burst and re-arms intact (`memory::bus`); the runner's chrome repaint leaves an armed journal armed and unchanged; the host snapshot tracks both new fields; a due `TEIdle` blink inside a proof fails it on memory while an idle `TEIdle` lets the cycle park, through the real handler under a real probe; the window and TextEdit mutators next to the admitted traps still cancel; the journal-complete list is exercised in both the plain and auto-pop encodings. The existing membership test is renamed to say what it checks and its negative case now names the real disqualifier (MoveTo mirrors `pnLoc` into dispatcher state).
- `cargo test --lib` (4,262 passed, 0 failed), `cargo build --release`, `cargo check --no-default-features`, `cargo package`.
- `cargo fmt --check` clean on the touched files; `cargo clippy --all-targets --all-features` adds no diagnostic relative to master.

https://claude.ai/code/session_01NoS2nQc8CBYWtb9LZgdGVh

Closes #1052

### DIFF
--- a/src/memory/bus.rs
+++ b/src/memory/bus.rs
@@ -517,6 +517,10 @@ pub struct MacMemoryBus {
     /// probed cycle was doing work, not waiting. Sticky until the runner
     /// takes it, so the verdict survives the journal's disappearance.
     write_probe_overflowed: bool,
+    /// Armed journal has no entry cap (parked-repaint byte-identity check:
+    /// no guest code runs, so unbounded growth is impossible and overflow
+    /// must not void the check).
+    write_probe_uncapped: bool,
     /// Temporary sparse-memory view used only while a parked native process
     /// runs its emulated 68k context. The runner installs and removes this
     /// view around each serialized execution interval.
@@ -1137,6 +1141,7 @@ impl MacMemoryBus {
             write_probe_spare: WriteProbeJournal::default(),
             write_probe_invalid: false,
             write_probe_overflowed: false,
+            write_probe_uncapped: false,
             foreign_address_space: None,
         };
         bus.write_word(super::globals::addr::ROM85, 0x7FFF);
@@ -1233,6 +1238,7 @@ impl MacMemoryBus {
             write_probe_spare: WriteProbeJournal::default(),
             write_probe_invalid: false,
             write_probe_overflowed: false,
+            write_probe_uncapped: false,
             foreign_address_space: None,
         }
     }
@@ -1341,6 +1347,17 @@ impl MacMemoryBus {
         self.write_probe_original = Some(journal);
         self.write_probe_invalid = false;
         self.write_probe_overflowed = false;
+        self.write_probe_uncapped = false;
+    }
+
+    /// Begin a write probe with no entry cap. Only for host-owned drawing
+    /// performed while the guest is parked (no guest code runs, so the
+    /// journal is bounded by what the drawing touches): the byte-identity
+    /// answer from [`Self::finish_write_probe_unchanged`] must never be
+    /// voided by the overflow guard sized for guest wait cycles.
+    pub(crate) fn begin_uncapped_write_probe(&mut self) {
+        self.begin_write_probe();
+        self.write_probe_uncapped = true;
     }
 
     /// Discard an incomplete write probe and restore normal fast-memory use.
@@ -1348,6 +1365,7 @@ impl MacMemoryBus {
         self.park_write_probe_journal();
         self.write_probe_invalid = false;
         self.write_probe_overflowed = false;
+        self.write_probe_uncapped = false;
     }
 
     /// Detach the armed journal so writes made meanwhile -- host-owned
@@ -1382,6 +1400,7 @@ impl MacMemoryBus {
         self.write_probe_spare = original;
         self.write_probe_invalid = false;
         self.write_probe_overflowed = false;
+        self.write_probe_uncapped = false;
         unchanged
     }
 
@@ -1431,7 +1450,7 @@ impl MacMemoryBus {
                 .as_mut()
                 .expect("write probe checked above");
             journal.entry(word).or_insert(original);
-            if journal.len() > WRITE_PROBE_MAX_ENTRIES {
+            if !self.write_probe_uncapped && journal.len() > WRITE_PROBE_MAX_ENTRIES {
                 // Too much written for a wait cycle: void the probe now so
                 // the fast paths (and fastmem) come back for the work in
                 // progress.
--- a/src/memory/bus.rs
+++ b/src/memory/bus.rs
@@ -676,6 +676,11 @@ impl Hasher for AddressHasher {
 /// withdrawn), so comparing them at the end is harmless.
 type WriteProbeJournal = HashMap<u32, u32, BuildHasherDefault<AddressHasher>>;
 
+/// An armed write journal temporarily detached from the bus by
+/// [`MacMemoryBus::suspend_write_probe`]; hand it back with
+/// [`MacMemoryBus::resume_write_probe`].
+pub(crate) struct SuspendedWriteProbe(WriteProbeJournal);
+
 /// RAM storage - either a stable owned allocation or borrowed slice.
 enum RamStorage {
     Owned(Vec<u8>),
@@ -1343,6 +1348,19 @@ impl MacMemoryBus {
         self.park_write_probe_journal();
         self.write_probe_invalid = false;
         self.write_probe_overflowed = false;
+    }
+
+    /// Detach the armed journal so writes made meanwhile -- host-owned
+    /// drawing the guest never performed -- are neither recorded nor able
+    /// to overflow it. The bus's fast paths come back while it is detached;
+    /// `resume_write_probe` re-arms the very same journal. `None` when no
+    /// journal is armed.
+    pub(crate) fn suspend_write_probe(&mut self) -> Option<SuspendedWriteProbe> {
+        self.write_probe_original.take().map(SuspendedWriteProbe)
+    }
+
+    pub(crate) fn resume_write_probe(&mut self, suspended: SuspendedWriteProbe) {
+        self.write_probe_original = Some(suspended.0);
     }
 
     /// Report -- and clear -- whether the most recent probe's journal
@@ -2811,6 +2829,39 @@ mod tests {
         bus.begin_write_probe();
         bus.fill_bytes_strided(0x105, 16, 4, 0xEE);
         assert!(bus.finish_write_probe_unchanged(), "same bytes: unchanged");
+    }
+
+    #[test]
+    fn suspended_write_probe_ignores_writes_and_rearms_intact() {
+        // A host-sized burst (more units than WRITE_PROBE_MAX_ENTRIES)
+        // overflows an armed journal; made while the journal is suspended it
+        // is neither recorded nor able to overflow, and the re-armed journal
+        // still catches the next write.
+        let mut bus = MacMemoryBus::new(1024 * 1024);
+        let base = 0x0008_0000u32;
+        let burst = (WRITE_PROBE_MAX_ENTRIES as u32 + 64) * 4;
+
+        bus.begin_write_probe();
+        bus.fill_bytes(base, burst, 0xAA);
+        assert!(bus.take_write_probe_overflow());
+        bus.cancel_write_probe();
+        bus.fill_bytes(base, burst, 0x00);
+
+        bus.begin_write_probe();
+        let suspended = bus.suspend_write_probe().expect("journal armed");
+        assert!(bus.suspend_write_probe().is_none());
+        bus.fill_bytes(base, burst, 0xAA);
+        bus.resume_write_probe(suspended);
+        assert!(!bus.take_write_probe_overflow());
+        bus.write_byte(base + 8, 0x01);
+        assert!(!bus.finish_write_probe_unchanged());
+
+        bus.begin_write_probe();
+        let suspended = bus.suspend_write_probe().expect("journal armed");
+        bus.fill_bytes(base, burst, 0x55);
+        bus.resume_write_probe(suspended);
+        assert!(bus.finish_write_probe_unchanged());
+        assert!(bus.suspend_write_probe().is_none());
     }
 
     #[test]
--- a/src/menu_manager.rs
+++ b/src/menu_manager.rs
@@ -817,7 +817,7 @@ impl SharedNativeMenuSelection {
         self.0.borrow_mut().take()
     }
 
-    #[cfg(test)]
+    /// Current value without consuming it (idle-park host snapshot).
     pub(crate) fn snapshot(&self) -> Option<(i16, i16)> {
         *self.0.borrow()
     }
--- a/src/runner.rs
+++ b/src/runner.rs
@@ -935,6 +935,10 @@ static WS_CANCEL_TRAP: AtomicU64 = AtomicU64::new(0);
 static WS_PARKED: AtomicU64 = AtomicU64::new(0);
 static WS_PERIOD_STEPS: AtomicU64 = AtomicU64::new(0);
 static WS_PROBE_OVERFLOWS: AtomicU64 = AtomicU64::new(0);
+static WS_RESUMED: AtomicU64 = AtomicU64::new(0);
+static WS_RESUME_FAIL_MEM: AtomicU64 = AtomicU64::new(0);
+static WS_RESUME_FAIL_HOST: AtomicU64 = AtomicU64::new(0);
+static WS_RESUME_FAIL_OTHER: AtomicU64 = AtomicU64::new(0);
 static WS_CANCEL_TRAP_WORDS: std::sync::Mutex<Option<std::collections::BTreeMap<u16, u64>>> =
     std::sync::Mutex::new(None);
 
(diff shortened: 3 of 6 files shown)


rlanday:
Force-pushed one amendment: the two new bus methods had been inserted between `take_write_probe_overflow`'s doc comment and the function, so rustdoc attached that comment to `suspend_write_probe`. The doc is reattached to its function; no code change.

benletchford (changes requested):
The trap-admission argument and tests are careful, but this branch is not mergeable against current master: it conflicts in `runner.rs` and `trap/dispatch.rs`, which have since moved process-owned runtime state. Please rebase and resolve against the current ownership model, then rerun the full/native/no-default/package checks and the measured GUI route.

The rebase should also pin the correctness boundary around `redraw_chrome_outside_idle_journal`. Suspending the journal is safe only if the repaint cannot change guest RAM relative to the proven state. Add a regression for the first repaint after a park where guest code has overwritten a chrome pixel before entering the idle loop; either prove the repaint is byte-identical at that point or make a changed repaint revoke/re-prove the park. That keeps the optimization from hiding a guest-visible framebuffer mutation from the exact-memory gate. Please link a focused issue as well.

rlanday:
Rebased onto master (0.25.0): kept the `ProcessContext`-plumbed `redraw_chrome()` and re-routed the journal-suspending wrapper through it at all four call sites; the park's host snapshot now reads the pending native menu selection through the shared handle (`SharedNativeMenuSelection::snapshot`, promoted from `cfg(test)`).

The correctness boundary is pinned per review — filed as #1052. While **parked**, a chrome repaint now runs under an uncapped write probe (`begin_uncapped_write_probe`: the journal entry cap is sized for guest wait cycles, and a parked repaint is host-bounded, so overflow must not void the answer); a repaint that is not byte-identical revokes the park through the ordinary cancel path and the site re-proves from scratch. Regression pair added: a guest-scribbled chrome pixel before the park makes the first parked repaint revoke it, and the byte-identical control keeps both the park and its armed, unchanged journal. Probing-phase repaints stay suspended with no check — proofs re-execute the whole cycle, so a changed repaint simply fails to prove.

Full suite, no-default-features check rerun locally; the measured GUI route rerun is next on the rebased stack and I'll post numbers here.

rlanday:
Correction on the previous push: it went out with the new revocation regression **failing** — two bugs on my side. My push script's test stage was piped through grep, so the chain gated on grep's exit status instead of cargo's; and the regression test's fixture never painted any chrome in the first place (`redraw_chrome` paints the menu bar only when `!menus.is_empty() && !fullscreen_locked && !menu_bar_hidden`, and the fixture had an empty menu list — so the "repaint" was a no-op and could neither restore the scribbled pixel nor trip the byte-identity check).

The gate itself is unchanged. The fixtures now install a menu the way `trap/framebuffer`'s menu-bar tests do (`trap::menu` is now `pub(crate)` so the runner-level test can construct one) and assert against a 0xAA sentinel rather than an exact palette index. With a painted bar: the scribbled-pixel repaint repaints and revokes the park as designed, and the byte-identical control keeps both the park and its armed, unchanged journal. Amended and force-pushed; the full suite is green on the amended commit.

rlanday:
Measured route, rebased branch (da65a8a) vs current master (12b6fe7 / 0.25.0), both release builds on upstream m68k 0.11.6.

**Instrument:** deterministic headless SimCity 2000 boot→gameplay input-script replay, fixed 400M-instruction budget, periodic screenshots disabled; metric is host `instructions retired` (`/usr/bin/time -l`), robust to background load; arms interleaved A,B,A,B,A,B on the same machine, saves directory reset before every run.

| pair | master | #962 | B/A |
|---|---|---|---|
| 1 | 308,638,589,869 | 297,895,057,202 | 0.9652 |
| 2 | 308,590,365,257 | 297,927,095,876 | 0.9654 |
| 3 | 308,515,437,454 | 297,915,521,941 | 0.9656 |

**−3.46% host instructions on the fixed budget — and the budget buys more game.** Master ends every run at exactly tick 128,758; the branch at exactly tick 184,781 (+43.5%), because the admitted traps let idle waits prove and fast-forward. Combined: **host work per guest tick drops 32.7%** (2.397M → 1.612M instructions/tick). The tick difference between arms is the intended effect, not nondeterminism — each arm's final tick count is identical across its three runs.

Wait-stats on a branch run: probe_starts=157,792, exact_repeats=153,844, journal overflows=0, and only 114 proof-cancelling traps remain in the whole replay (top: $AA9A ×42, $AB1D ×34, $A01C ×32). parked=resumed=0 headless — parking (and therefore the new repaint-revocation gate) engages at GUI frontend boundaries; the headless gain comes from the admission alone.

benletchford (approved):
The rebased implementation resolves the blocking review: parked chrome repaints are checked for byte identity and revoke on change, the new host-side inputs are included in the resume snapshot, mutation neighbors remain excluded, targeted regressions cover both safe and revocation paths, and the measured deterministic replay shows the intended performance improvement. I also removed one duplicated documentation block via the enabled maintainer-edit path.

PR #963, opened 2026-08-29
perf(quickdraw): copy CopyBits rows span by span instead of pixel by pixel

### LINKED ISSUE #1040: perf(quickdraw): accelerate common 8-bit CopyBits spans
## Observed behavior

The common unscaled, non-overlapping 8-bit `srcCopy` path performs region membership, bus access, and mode evaluation for every pixel. In a representative SimCity 2000 deterministic replay, this shape accounts for nearly all CopyBits pixels and is a leading runtime cost.

## Expected behavior

Eligible transfers should process decoded region spans row-by-row while preserving byte-exact `srcCopy` results, palette translation, bus write guards, screen bookkeeping, and fallback behavior for scaling, overlap, masks, other modes, unsupported regions, or diagnostics.

## Reproduction

Run the deterministic SimCity 2000 replay used by the runner performance harness and profile the per-pixel CopyBits paths. Compare guest ticks, milestone sequences, and destination pixels against the current implementation, including complex vis/clip regions.

rlanday:
## Summary

For the commonest CopyBits shape — 8-bit to 8-bit `srcCopy`, unscaled, non-overlapping bitmaps — both CopyBits arms (the inline `_CopyBits` handler, and `copy_bits_common` behind `StdBits`, which is where a port with `grafProcs` sends every transfer) did a region membership test, a bus read, a mode evaluation and a bus write for every pixel.

Both arms now share one helper, `copy_bits_src_copy_rows_8bpp`. For `srcCopy` every written pixel is `table[source pixel]`, so each arm derives a 256-entry table from the very per-pixel mode function it calls in its loop, with the arguments it passes there (the destination byte does not take part — a test pins that), and each row moves as one bus read, one table pass and one bus write per covered span. A region that needs a per-pixel test is honoured through its decoded row spans — the same `RegionMembershipCache` rows `region_contains_point_cached` consults — via a single merge walk over the two sorted sequences. The bus's slice primitives keep `write_byte`'s guards (read-only code, write tracing, the idle-proof write journal), so nothing observable changes.

The helper refuses, *before writing anything*, when a tested region's decoded band does not cover the clip or the clip lies outside the destination bounds, and the arm's pixel loop then runs exactly as before; an area floor of 64 pixels keeps transfers smaller than the table from paying for it. The arms fall through to their existing screen bookkeeping (dialog-snapshot refresh, kiosk letterbox) rather than returning early. Scaling, masks, other modes, overlapping bitmaps, depth conversions and every diagnostic probe still take the pixel loops unchanged.

### Why it matters

A pixel-weighted census of the headless SimCity 2000 replay (an env-gated diagnostic, kept out of this PR): **90.2 %** of all CopyBits pixels are exactly the shape above but behind a complex visRgn, 8.9 % the same shape unclipped with a palette translation, 0.9 % both; nothing else registers. In the steady-state profile the per-pixel CopyBits work was the largest remaining Systemless-side item.

### Evidence

**Instrument.** Headless SimCity 2000 (`--headless --max-instructions 2000000000 --input-script`, the #824 input replay into a running city), `/usr/bin/time -l` for host instructions retired, cycles and wall clock, guest ticks from the run log. Three runs per arm, interleaved. Instructions and ticks are load-insensitive; cycles and wall are not, and two of the three pairs here ran while macOS's `mediaanalysisd` was consuming ~190 % CPU, so those two figures are quoted from the quiet pair only.

Base = current master (0.23.8, m68k 0.11.6):

| SimCity 2000, 2B | master 298bf2e | this PR | Δ | runs won |
|---|---|---|---|---|
| host instructions retired | 394.85 G | 375.58 G | **−4.88 %** (−4.7 / −4.9 / −5.0) | 3/3 |
| CPU cycles (quiet pair) | 163.56 G | 156.63 G | **−4.23 %** | — |
| wall clock (quiet pair) | 46.22 s | 43.21 s | **−6.51 %** | — |
| guest ticks | 283,023 | 283,023 | identical | — |

Post-change steady-state profile (20 s window, 14,711 samples): `region_contains_point`, the per-byte bus accessors under CopyBits, and `copy_bits_common` itself no longer register a sample; the only CopyBits-attributable hit left is `write_bytes` at 0.27 %.

**Correctness.** Guest tick count identical in every run, and the window/dialog/menu/resource milestone sequences of the two arms are line-for-line identical over the full replay (72 lines, zero diff).

## Validation

Six new tests: table destination-independence for the null-CLUT, translated and shared-mask colour-space cases; the span intersection against `endpoints_contain_point` over a dense sweep (including a doubled edge); the helper against `region_contains_point` through a complex region, plus its three refusal paths leaving memory untouched; `StdBits` and `_CopyBits` 16×16 transfers through complex vis/clip regions checked pixel-for-pixel against region membership; and a `StdBits` screen transfer at the area floor that must leave a visible dialog's retained snapshot updated. Full lib suite: 4,263 passed, 0 failed. `cargo fmt --check` clean on the touched file; clippy adds no diagnostic relative to master (465 → 461, none new).

Checks run on the branch: `cargo build --release`, `cargo test --lib` (and `--features test-support`), `cargo check --no-default-features`, `cargo package`; `cargo fmt --check` on the touched files; `cargo clippy --all-targets --all-features` adds no diagnostic relative to master.

https://claude.ai/code/session_01NoS2nQc8CBYWtb9LZgdGVh

Closes #1040.

### DIFF
--- a/src/trap/quickdraw.rs
+++ b/src/trap/quickdraw.rs
@@ -50,6 +50,10 @@ pub(super) struct RawPixPat {
     pub(super) clut: [[u16; 3]; 256],
 }
 
+/// Smallest clip area (in pixels) for which the whole-row CopyBits path is
+/// taken: below it the 256-entry table costs more than the pixels it saves.
+const COPY_BITS_ROW_PATH_MIN_PIXELS: i32 = 64;
+
 pub(super) struct RegionMembershipCache {
     pub(super) top: i16,
     pub(super) rows: Vec<Vec<i16>>,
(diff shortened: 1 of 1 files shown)


benletchford (approved):
Reviewed the row-span gates, region intersection, palette table derivation, overlap fallback, bus-write semantics, and screen bookkeeping. The focused release-mode equivalence tests pass locally, all CI checks are green, and the implementation remains narrowly fail-closed. This is a sound performance direction.

PR #992, opened 2026-08-30
feat(desktop): draw the guest cursor as the host pointer on macOS

### LINKED ISSUE #1049: Hardware cursor must scale with the presentation viewport, not window width over guest width
Follow-up to the review of #992. The hardware cursor currently derives its scale as `window inner width ÷ guest screen width`, while rendering and input use the active presentation `ContentRect` with `scale = min(drawable_width / content.width, drawable_height / content.height)` (see `physical_to_mac_in_viewport` and `presentation_content_rect`).

Observed with the #992 implementation:
- A height-constrained window (drawable height is the limiting axis) shows an **oversized** pointer relative to guest pixels.
- A learned/cached gameplay crop (e.g. a 640-wide content rect inside an 800-wide guest screen) shows an **undersized** pointer.
- Transient dialog expansion changes the presentation scale without changing `guest_width`, so the cursor scale silently drifts from the viewport scale.

Expected: the cursor's guest-pixel scale always equals the presentation viewport scale — the same `ContentRect` + min-axis scale used by rendering and input — including under transient expansion, and the cursor rebuilds when that scale changes.

https://claude.ai/code/session_01LnuSFJRzTWFhxrFUMw622y

rlanday:
## Summary

- draw the guest cursor through the macOS hardware pointer instead of compositing it into each guest frame
- preserve QuickDraw cursor masks, color, hotspot, visibility, and crisp integer scaling
- derive cursor scale from the active presentation viewport so height-constrained windows, learned crops, and transient dialog expansion stay aligned with rendering and input
- retain `SYSTEMLESS_SOFTWARE_CURSOR=1` as an explicit fallback

This removes the guest cursor from the content presentation pipeline, so pointer motion no longer requires a frame submission. Guest content that follows the mouse can still trail the hardware cursor because it remains subject to the normal presentation pipeline.

## Validation

- `cargo build --release`
- `cargo test --lib`
- `cargo test --bin systemless host_cursor`
- `cargo package`
- cross-platform CI and macOS build/test jobs pass
- manual macOS checks cover resize, guest cursor visibility, the software-cursor fallback, and windowed/fullscreen presentation

Closes #1049

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,7 +60,7 @@ test-support = []
 
 [target.'cfg(target_os = "macos")'.dependencies]
 objc2 = { version = "0.5.2", optional = true }
-objc2-app-kit = { version = "0.2.2", optional = true, features = ["NSApplication", "NSBitmapImageRep", "NSCell", "NSGraphics", "NSImage", "NSImageRep", "NSMenu", "NSMenuItem"] }
+objc2-app-kit = { version = "0.2.2", optional = true, features = ["NSApplication", "NSBitmapImageRep", "NSCursor", "NSCell", "NSGraphics", "NSImage", "NSImageRep", "NSMenu", "NSMenuItem"] }
 objc2-foundation = { version = "0.2.2", optional = true, features = ["NSGeometry", "NSObject", "NSProcessInfo", "NSString"] }
 objc2-metal = { version = "0.2.2", optional = true, features = [
     "MTLBuffer",
--- a/README.md
+++ b/README.md
@@ -284,6 +284,7 @@ name.
 | `SYSTEMLESS_TRACE_LOAD` | Logs archive, VFS, resource, and startup loading diagnostics. |
 | `SYSTEMLESS_TRACE_LOADSEG` | Logs Segment Loader jump-table patching. |
 | `SYSTEMLESS_TRACE_TRAP_COUNTS` | Prints trap dispatch frequency summaries. |
+| `SYSTEMLESS_SOFTWARE_CURSOR` | Restores the composited guest cursor overlay instead of the hardware pointer (macOS). |
 
 ## References & Documentation Conventions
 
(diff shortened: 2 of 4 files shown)


benletchford (changes requested):
The hardware-cursor direction is good, but its scale is wrong for existing macOS presentation paths. `HostCursor::sync` uses `inner.width / guest_width`; rendering and input use the active `ContentRect` and `min(drawable_width / content.width, drawable_height / content.height)`. A height-constrained window therefore gets an oversized pointer, and a learned/cached gameplay crop gets an undersized pointer (for example a 640-wide crop inside an 800-wide guest). Transient dialog expansion can change that scale too without changing `guest_width`.

Please pass the actual presentation content rectangle/viewport scale into the cursor sync, include it in the rebuild key, and add coverage for a height-constrained window and a cropped presentation. Please also link an issue that records the observed/expected behavior and document `SYSTEMLESS_SOFTWARE_CURSOR` alongside the other runner environment variables before this default-on behavior lands.

rlanday:
Rebased onto master (0.25.0, clean) and reworked the scale per review: the cursor now derives its guest-pixel scale from the presentation viewport — the same `presentation_content_rect` (including transient dialog expansion) and `min(drawable/content)` axis rule that rendering and input use — via a pure `presentation_scale` helper, and the scale is part of the cursor rebuild key so viewport changes rebuild it. Added unit coverage for the height-constrained window and the cropped-presentation (640-in-800) cases, documented `SYSTEMLESS_SOFTWARE_CURSOR` in the README environment table, and filed #1049 recording observed/expected. Full local set rerun: release build, lib+bin tests, no-default-features check, package.

rlanday:
Force-pushed a hygiene fix: my local rustfmt had reflowed seven untouched upstream files (runner.rs, loader/ppc.rs, trap/dispatch.rs, trap/toolbox.rs, trap/dialog.rs, memory/address_space.rs, process_context.rs) and those slipped into the previous push. Reverted them to master byte-for-byte — the PR now only touches host_cursor.rs, systemless.rs, README.md, and Cargo.toml (the NSCursor feature). No functional change.

benletchford (approved):
The revised implementation resolves the blocking scale issue: it uses the active presentation viewport and binding axis, rebuilds on scale changes, covers both height-constrained and cropped presentations, documents the software fallback, and links the tracking issue. CI is green.

PR #1005, opened 2026-08-30
fix(window): use system wctb resource for default window color table

### LINKED ISSUE #1004: fix(window): use system wctb resource for default window color table
### Description
When creating a window without an explicit `'wctb'` resource, `ensure_window_aux_record` and `default_window_color_table_handle` fell back to `gdPMap->pmTable` (the GDevice's indexed 256-color pixel ColorTable) rather than the standard Window Manager `'wctb'` resource ID 0 (`WinCTab`).

Because applications with custom palettes (such as *Spin Doctor*) install custom color tables into the `GDevice`, entry 0 of `gdPMap->pmTable` contains whatever RGB value the palette mapped to index 0. Treating this device ColorTable as a semantic `WinCTab` caused `window_content_color` to extract entry 0 as `wContentColor`, inadvertently coloring plain windows and modal dialogs with the game's palette index 0 color rather than the standard white/light gray background.

Additionally, `window_content_color` did not restrict part-code lookups to semantic `WinCTab`/`DCTab` tables with `table_size <= 12`, allowing device ColorTables with 256 entries to be mistakenly treated as window color tables.

### References
- *Inside Macintosh Volume V* (1986), pp. V-201..V-203, V-278
- *Inside Macintosh: Macintosh Toolbox Essentials* (1992), pp. 4-71..4-72, 4-118, 4-127, 6-120..6-121

benletchford:
## Summary

- use the standard system `wctb` ID 0 when a window has no explicit color table
- reject indexed device CLUTs when interpreting semantic Window Manager color roles
- add regressions for the white default window content color and 256-entry CLUT rejection

Applications can install custom palettes in the main graphics device. Treating that device ColorTable as a semantic `WinCTab` made palette entry 0 leak into otherwise standard window and dialog backgrounds. The fallback now uses the synthesized system `wctb` resource instead.

## References

- *Inside Macintosh Volume V* (1986), pp. V-201 to V-203 and V-278
- *Inside Macintosh: Macintosh Toolbox Essentials* (1992), pp. 4-71 to 4-72, 4-118, and 4-127

## Validation

- `cargo test --lib default_window_color_table --locked`
- `cargo test --lib window_content_color --locked`
- `git diff --check`

Closes #1004

### DIFF
--- a/src/trap/window.rs
+++ b/src/trap/window.rs
@@ -352,7 +352,11 @@ impl super::TrapDispatcher {
     ) -> Option<u32> {
         let resource_ptr = self
             .find_or_load_resource_any(bus, resource_type, table_id)
-            .map(|(_, ptr)| ptr)?;
+            .map(|(_, ptr)| ptr)
+            .or_else(|| match (&resource_type, table_id) {
+                (b"wctb", 0) => self.synthesize_system_wctb(bus, 0),
+                _ => None,
+            })?;
 
         if bus.get_alloc_size(resource_ptr).unwrap_or(0) < 8 {
             return None;
@@ -405,7 +409,15 @@ impl super::TrapDispatcher {
             return None;
         }
 
-        let entry_count = u32::from(bus.read_word(ctab_ptr + 6)) + 1;
+        // WCTab/DCTab resources contain the semantic Window Manager
+        // roles. A device ColorTable instead contains indexed palette entries
+        // and must not be interpreted as window-part identifiers.
+        let table_size = bus.read_word(ctab_ptr + 6);
+        if table_size == 0xFFFF || table_size > 12 {
+            return None;
+        }
+
+        let entry_count = u32::from(table_size) + 1;
         for index in 0..entry_count {
             let entry = ctab_ptr + 8 + index * 8;
             if bus.read_word(entry) == 0 {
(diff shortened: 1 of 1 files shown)


benletchford:
The window fallback part is the right fix: a default AuxWin must use the system `wctb` ID 0, and a 256-entry device CLUT must not be parsed as semantic window-part colors. I cannot merge this revision because the added synthetic `dctb` ID 0 and explicit `0x8282` gray content color are not supported by the cited source. Macintosh Toolbox Essentials 6-75 shows a `dctb` with `ctSize = -1` to request the default window colors; it does not define a standard `dctb` 0 or a gray default. The existing synthetic system `wctb` 0 is the source-backed default and has white `wContentColor`.

Please reduce this to the `wctb` fallback plus the semantic-table guard (the documented standard-WDEF maximum is `ctSize <= 12`), or add direct source/oracle evidence for any separate default dialog table. Also clean the stray blank lines: `git diff --check master...HEAD` currently reports new blank lines at EOF in `dialog.rs` and `window.rs`.

benletchford:
I applied the requested revision directly: removed the unsupported synthetic `dctb` path and gray default, retained only the source-backed system `wctb` fallback and semantic-table guard, cleaned the whitespace issues, rebased onto current master, and reran the two focused regressions plus `git diff --check`. Fresh CI is now running on the amended branch.

PR #1082, opened 2026-08-31
perf(runner): admit journal-complete poll traps and bound doomed idle probes per site

rlanday:
## Problem

During live SimCity 2000 gameplay the app pins ~100% of a core, while the guest
retires only a fraction of its realtime instruction budget. The cause is not guest
work but the idle-cycle prover's failure mode: the game's play-mode poll loop calls
`_PtInRect` ($A8AD) and `_QDExtensions` ($AB1D — a GetGWorld/SetGWorld save/restore
pair around its hit-testing) on every pass. Neither trap was on the admission list,
so every probe died to a "foreign" trap mid-cycle. While a probe's write journal is
armed, the bus withdraws its fast paths — execution is de-optimized — so each doomed
probe taxes a full partial cycle and proves nothing.

Two accounting problems made the tax unbounded rather than budget-capped:

1. **The per-site probe budget lived in a single slot.** The poll loop interleaves
   probes from about six distinct anchor sites, and every probe from site B reset
   site A's spent-budget record — so the 2-probes-per-site-per-tick cap never bound,
   and the loop re-armed ~1,700 doomed journals per second, forever.
2. **A trap-cancelled site had no memory.** Nothing distinguished "this site's cycle
   funnels through a non-admitted trap every single pass" from a one-off cancel, so
   the same doomed sites were re-probed at full rate on every tick.

## Changes

**Commit 1 — admissions.** `_PtInRect` and the two QDExtensions selectors are
admitted to idle-cycle proofs, vetted per the `idle_cycle_trap_is_journal_complete`
checklist:

- **PtInRect ($A8AD):** reads the point and rect from stack/RAM, writes one Boolean
  at `sp+8` (journaled), pops A7 (CPU state the proof compares). No dispatcher state
  touched. (An existing test used PtInRect as its example of a trap with "host-side
  HLE semantics" — reading the handler shows that was never true; the test now uses
  a genuinely non-admitted QDExtensions selector.)
- **QDExtensions ($AB1D) is selector-multiplexed**, so the admission is per-selector:
  live D0 is now passed into the admission check, and only GetGWorld ($00080005) and
  SetGWorld ($00080006) qualify. GetGWorld writes its two VAR results through the
  bus (journaled) and pops A7. SetGWorld's guest-visible writes (the `thePort` low
  global and the A5 world's mirror) are journaled; the dispatcher port/draw-state
  mirrors it rewrites are pure functions of its stack arguments and guest RAM, so
  across two proven-identical cycles they sit at a fixed point, and parking
  preserves them. Its one impure path — a cold-start `ensure_main_gdevice`
  allocation — writes fresh heap bytes that the byte-identity proof rejects on its
  own, so no special-casing is needed. Every other selector (NewGWorld, LockPixels,
  UpdateGWorld, …) allocates, locks or frees host-mirrored state and still cancels.
  A non-admitted selector cancels **pre-dispatch**, so the post-dispatch
  classification only ever sees the admitted selectors, whose handlers leave D0
  intact.

  The wait-stats cancel diagnostics gain a per-selector breakdown for $AB1D (live D0
  recorded at the pre-dispatch cancel site) — a cancel count against the bare trap
  word cannot name the selector that needs vetting, and this is how the pair above
  was identified.

**Commit 2 — per-site accounting + backoff.** The single accounting slot becomes a
small fixed table (8 records; the busiest loop measured interleaves ~6 sites), so
the per-tick budget binds per site regardless of interleaving. On top of that, a
site whose probes die to a non-admitted trap **twice in a row** is backed off
exponentially (2^streak ticks, capped at 120 ≈ 2 s); any probe that later closes on
its origin site — whatever the proof's verdict — resets the streak. This bounds the
worst-case armed-journal tax for *any* workload that polls through a trap the
admission list does not yet cover. The backoff is scheduling-only and cannot affect
soundness: it decides when a probe is worth arming, never whether a proof is valid;
parking still requires the full CPU + byte-identity proof. Slot eviction (if an app
ever exceeded 8 concurrent sites) degrades to bounded extra probing, never to a
wrong park.

## Measurements

Live GUI gameplay (the workload that exposed the problem; headless replay does not
reproduce it — no live mouse-over, so the hit-testing poll path stays cold).
Protocol: ~1-minute open-gameplay sessions played by hand on the same machine and
build configuration, app CPU sampled by `ps` every 5 s against the process's own
child PID, wait-stats dumped on window close. Sessions are human-driven, so arms are
not input-identical; the effect size dwarfs that variance.

| | before (admissions absent) | after (this PR) |
|---|---|---|
| app CPU, active play | ~90–100% of a core | 24–43% |
| app CPU, hands off | ~90–100% | 12–18% |
| probe cancels by foreign trap | 155,882 / 91 s (all $AB1D) | 48 / 61 s |
| probes armed | 164,421 | 5,251 |
| proofs parked | 1,024 | 2,271 |
| backoffs engaged | n/a | 1 |

In the "after" profile the main thread spends ~half its samples blocked in
`mach_msg` between frames — the frame loop finishes its work and sleeps, where
before it burned to its wall-clock deadline executing de-optimized. The residual
active-play CPU is genuine guest execution (interpreter), not prover overhead.
With the admissions in place the backoff engaged exactly once — it is the safety
net for future unvetted poll traps, not the mechanism carrying this win.

Headless steady-state: **neutral**. SimCity 2000 headless replay
(`--input-script`, periodic screenshots off, `/usr/bin/time -l` host
instructions retired, arms interleaved in alternating order on the same box):
6 pairs at 400M guest instructions → mean **+0.04%** host instructions (range
−0.04%..+0.13%; noise floor on this harness ≈ ±0.1%); 3 pairs at 2B → mean
**+0.004%** (range −0.01%..+0.02%). Guest ticks are identical in every run of
both arms (184,781 at 400M; 358,121 at 2B) — no headless-observable behavior
change. Wait-stats: neither arm parks on this workload (no live pointer, so
the hit-testing poll path stays cold); the backoff engaged twice on the
residual cancellers at no measurable cost.

## Tests

- `journal_complete_traps_do_not_cancel_an_idle_probe` now covers PtInRect in both
  encodings and both admitted QDExtensions selectors (plain and auto-pop), and
  still proves MoveTo cancels.
- `exact_idle_cycle_rejects_changed_memory_and_non_quiescent_traps` uses a
  non-admitted QDExtensions selector (LockPixels) as its canceller.
- New: `repeated_trap_cancels_back_a_site_off_and_a_closed_proof_resets_it` —
  the backoff lifecycle: two cancels back the site off, probing resumes when the
  backoff expires, an origin-closed proof resets the streak.

https://claude.ai/code/session_01JbmJEAB3FNp3wZPkwA1XnK

### DIFF
--- a/src/runner/mod.rs
+++ b/src/runner/mod.rs
@@ -1146,6 +1146,8 @@ static WS_RESUMED: AtomicU64 = AtomicU64::new(0);
 static WS_RESUME_FAIL_MEM: AtomicU64 = AtomicU64::new(0);
 static WS_RESUME_FAIL_HOST: AtomicU64 = AtomicU64::new(0);
 static WS_RESUME_FAIL_OTHER: AtomicU64 = AtomicU64::new(0);
+static WS_CANCEL_BACKOFFS: AtomicU64 = AtomicU64::new(0);
+static WS_BACKOFF_SKIPS: AtomicU64 = AtomicU64::new(0);
 static WS_CANCEL_TRAP_WORDS: std::sync::Mutex<Option<std::collections::BTreeMap<u16, u64>>> =
     std::sync::Mutex::new(None);
 static WS_CANCEL_AB1D_SELECTORS: std::sync::Mutex<Option<std::collections::BTreeMap<u32, u64>>> =
@@ -1203,6 +1205,11 @@ pub fn dump_wait_stats() {
         WS_RESUME_FAIL_MEM.load(AtomicOrdering::Relaxed),
         WS_RESUME_FAIL_HOST.load(AtomicOrdering::Relaxed),
         WS_RESUME_FAIL_OTHER.load(AtomicOrdering::Relaxed),
+    );
+    eprintln!(
+        "[WAIT-STATS] cancel_backoffs={} backoff_skips={}",
+        WS_CANCEL_BACKOFFS.load(AtomicOrdering::Relaxed),
+        WS_BACKOFF_SKIPS.load(AtomicOrdering::Relaxed),
     );
     if let Ok(guard) = WS_CANCEL_TRAP_WORDS.lock() {
         if let Some(map) = guard.as_ref() {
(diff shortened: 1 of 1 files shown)


benletchford:
Completed the rebase onto current master, resolving the runner module split and updating the added tests to use the guest clock rather than the removed dispatcher field. Also corrected the backoff deadline comparison to expire across a 32-bit TickCount wrap and added a regression. All six idle-cycle tests pass; the rebased branch is pushed and fresh CI is running.

benletchford (approved):
Completed the rebase onto the split runner layout and reviewed the journal-complete poll admissions and bounded site backoff. Fixed deadline handling across the wrapping guest TickCount and added a regression; all six idle-cycle tests and the journal-complete classifier test pass locally. Fresh CI is green, including the library suite, 68k/PowerPC/theme showcase integrations, and headless/package checks.

PR #1210, opened 2026-09-01
fix(quickdraw): resolve screen colors against logical palette

rlanday:
## Summary

Fix indexed-screen RGB color resolution after an application replaces the main screen's logical ColorTable.

This fixes the SimCity 2000 regression in which dialog borders became green and buttons became blue. It preserves the hardware-palette animation behavior fixed by #1152, while making the inverse lookup follow the GDevice's current *logical* palette instead of assuming that the logical palette is always the standard System 7.5.3 palette.

The same semantics are applied to the native 68K and PPC QuickDraw paths. The 68K Palette Manager path also now restores the System palette when `ActivatePalette` is called for a window with neither a window-specific palette nor an application-default palette, matching the existing PPC behavior.

Before:
sc2k-color-regression-bad-94c5f8f

After:
sc2k-color-regression-good-parent-39e4ca1

## Root cause

#1152 correctly separated two pieces of indexed-color state that classic QuickDraw treats differently:

- The GDevice ColorTable and its inverse table determine which pixel index QuickDraw writes for a requested RGB color.
- The live hardware palette determines how an existing pixel index appears on the display. A game may animate or fade this palette without changing RGB-to-index mapping.

The regression came from implementing the first item as one permanently cached inverse table derived from the standard System 7.5.3 8-bit palette. That works while the screen's logical ColorTable is still canonical, and it prevents a hardware-only fade from feeding back into drawing. It is not sufficient when an application installs a genuinely different logical ColorTable: the inverse table belongs to that ColorTable and must change with it.

SimCity 2000 installs a custom logical screen palette. Its dialog grays exist at different indices than they do in the standard System palette. Looking them up in the standard inverse table therefore selected indices whose entries in SC2K's live palette were unrelated colors:

| Requested RGB | Index from stale standard ITable | SC2K color at that index | Exact index in SC2K logical table |
| --- | ---: | --- | ---: |
| `9F9F,9F9F,9F9F` | 86 | `0000,0000,9B9B` (blue) | 144 |
| `E3E3,E3E3,E3E3` | 245 | `FCFC,F3F3,0505` (yellow) | 139 |
| `5757,5757,5757` | 251 | `1F1F,B7B7,1414` (green) | 149 |
| `8383,8383,8383` | 249 | `9090,7171,3A3A` (brown) | 146 |
| `BBBB,BBBB,BBBB` | 247 | `8080,8080,8080` (wrong gray) | 142 |

That is why controls that should have used neutral grays appeared blue or green.

This follows the classic model described by Inside Macintosh: the inverse table is associated with a GDevice ColorTable, and `ctSeed`/`iTabSeed` track whether that relationship is current. Hardware-only palette changes do not require a new inverse table, but replacing the logical ColorTable does.

- [Inside Macintosh: Imaging With QuickDraw](https://developer.apple.com/library/archive/documentation/mac/pdf/Imaging_With_QuickDraw/Imaging_TOC.pdf), particularly `MakeITable` and inverse-table behavior
- [Inside Macintosh: QuickDraw GX / GDevice color-table management](https://dev.os9.ca/techpubs/mac/QuickDraw/QuickDraw-198.html)
- [Inside Macintosh: Advanced Color Imaging — inverse tables](https://dev.os9.ca/techpubs/mac/ACI/ACI-103.html)

## Implementation

- Resolve main-screen RGB colors against `color_manager_clut`, the logical GDevice ColorTable, in both native 68K and PPC `RGBForeColor`/`RGBBackColor`/`Color2Index` paths.
- Continue using `device_clut` as the displayed hardware palette, so `SetEntries`-style fades and palette animation do not alter the indices QuickDraw writes.
- Use the exact precomputed System 7.5.3 inverse-table oracle only when all 256 logical ColorTable entries are canonical.
- For a custom ColorTable, return an exact entry first; otherwise compute the nearest entry for the requested 4-bit inverse-table cell using the same cell-center rule as `build_inverse_table_bytes`.
- Restore the System palette on native 68K `ActivatePalette(window)` when the window has no associated palette and there is no application-default palette. Protected and reserved cells remain intact, consistent with the existing PPC path and Palette Manager ownership rules.

The standard CLUT is cached once. Canonical lookup remains constant-time after the 256-entry identity check. A custom lookup examines at most 256 entries for the one requested cell, avoiding construction of a complete 4,096-cell inverse table for infrequent RGB color traps. This does not change guest instruction execution or the runner's performance paths.

## Validation

- Rebased onto upstream `master` at `1637baa`.
- Added native 68K and PPC regressions in which the logical and hardware palettes deliberately disagree; both resolve `9F9F` gray to logical index 144 rather than the stale/hardware index.
- Added a regression for restoring System colors after a window and application-default palette are both absent.
- Preserved the #1152 hardware-fade regression: standard logical inverse lookup remains unchanged when only the hardware CLUT is altered.
- `cargo test --all --quiet`: 4,753 library tests passed (3 ignored), 79 binary tests passed, 4 `pack-web` tests passed, and the toolbox showcase passed.
- All 13 toolbox-showcase visual reference frames are unchanged.
- Deterministic SC2K headless replay with a fresh save directory and identical scripted input for 10,000,000 guest instructions reached 17,574 ticks. Its final framebuffer was byte-identical to the known-good parent of #1152:
  - fixed/current: `6e1120f807b105be2f59088098382cd444926949b56b5c06dbc35d53d1e91e6c`
  - known-good parent: `6e1120f807b105be2f59088098382cd444926949b56b5c06dbc35d53d1e91e6c`
  - regressed #1152/master image: `fb36884d00bd112a8cea369bf9b5bf6d61f4f3c10b33b7b7c98f188f7bc7ac8e`

### DIFF
--- a/src/loader/ppc/mod.rs
+++ b/src/loader/ppc/mod.rs
@@ -17147,7 +17147,12 @@ fn dispatch_supported_import(
                 if main_8bpp {
                     quickdraw_fore_indices.insert(
                         *current_gworld,
-                        ppc_color_to_index(memory, *current_gdevice, screen_clut, color) as u8,
+                        ppc_color_to_index(
+                            memory,
+                            *current_gdevice,
+                            color_manager_clut,
+                            color,
+                        ) as u8,
                     );
                 } else {
                     quickdraw_fore_indices.remove(current_gworld);
@@ -17180,7 +17185,12 @@ fn dispatch_supported_import(
                 if main_8bpp {
                     toolbox_startup.quickdraw_back_indices.insert(
                         *current_gworld,
-                        ppc_color_to_index(memory, *current_gdevice, screen_clut, color) as u8,
+                        ppc_color_to_index(
+                            memory,
+                            *current_gdevice,
+                            color_manager_clut,
+                            color,
+                        ) as u8,
                     );
                 } else {
                     toolbox_startup
@@ -17306,7 +17316,9 @@ fn dispatch_supported_import(
         }
         PpcImportDispatcherTarget::Color2Index => {
             let pixel = ppc_read_rgb_color(memory, cpu.gpr[3])
-                .map(|color| ppc_color_to_index(memory, *current_gdevice, screen_clut, color))
+                .map(|color| {
+                    ppc_color_to_index(memory, *current_gdevice, color_manager_clut, color)
+                })
                 .unwrap_or(0);
             Some(PpcImportAction::Return(pixel))
         }
@@ -63169,7 +63181,7 @@ fn ppc_index_to_color(
 fn ppc_color_to_index(
     memory: &mut PpcSectionMem,
     current_gdevice: u32,
-    screen_clut: &[[u16; 3]; 256],
+    logical_screen_clut: &[[u16; 3]; 256],
     color: PpcRgbColor,
 ) -> u32 {
     let pixmap = ppc_current_gdevice_record(memory, current_gdevice)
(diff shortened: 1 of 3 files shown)


benletchford (approved):
Reviewed the logical-versus-hardware palette separation and both ABI call paths. The custom-table lookup retains the existing standard inverse-table oracle, while default-palette restoration agrees with the Palette Manager model in Inside Macintosh Volume V, V-155 and V-161–V-162. Tested the patch together with the merged memory/trap optimizations on current master: 4,990 library tests passed, three ignored. Existing cross-platform CI is green.

Issue #1363, opened 2026-09-04
refactor: unify the Macintosh runtime across 68K and PowerPC


benletchford:
## Observed architecture

Systemless currently shares substantial Macintosh state between its 68K and PowerPC execution paths, but the sharing is implemented through adapter attachments, foreign memory mappings, synchronized state mirrors, process-wide guest-call stacks, and manager-specific callback trampolines. This gives strong mixed-execution behavior while leaving CPU adapters and the runner responsible for architecture-independent Toolbox state and scheduling.

Representative symptoms include:

- the runner owns CPU contexts, callbacks, scheduling, host presentation, and process services;
- PowerPC application state retains QuickDraw, windows, events, sound, files, callbacks, and other Toolbox managers;
- 68K and PowerPC memory views share bytes through bridges rather than one address-routing authority;
- Mixed Mode continuations are process-wide and separate from Thread Manager tasks;
- many operations have separate 68K trap and PowerPC import implementations;
- callback-capable services use bespoke trampolines instead of a resumable execution-kernel contract.

## Expected architecture

Both instruction sets should execute inside one Macintosh world:

- machine, process, task, CPU-engine, and host state have explicit lifetimes;
- every guest-visible fact has one semantic authority;
- both CPU memory interfaces are views over one authoritative address map;
- tasks own suspended execution and Mixed Mode continuations;
- a universal procedure resolver and ProcInfo codec call either ISA;
- callback-capable services yield typed effects to the execution kernel;
- 68K trap routing and PowerPC CFM routing remain guest-visible before converging on one semantic implementation;
- host presentation derives snapshots without selecting an active ISA.

## Implementation sequence

- [ ] inventory runtime fields and operations by scope, authority, and destination service
- [ ] add architectural ratchets against new adapter-owned manager state and callback machinery
- [ ] prove vertical slices for TickCount, NewHandle, NGet/NSetTrapAddress, CopyBits, and MenuSelect with a cross-ISA MDEF
- [ ] introduce task-owned continuations and one universal guest-call path
- [ ] consolidate address routing while retaining efficient storage backings
- [ ] split machine, process, task, engine, and host ownership
- [ ] migrate services vertically and delete superseded implementations
- [ ] replace adapter-selected presentation with derived world snapshots
- [ ] remove compatibility attachments, state mirrors, and manager-specific trampolines

## Validation

[...]

PR #1362, opened 2026-09-04
feat(theme): experiment with Systemless guest chrome

benletchford:
## Summary

- make `systemless-default` the normal 68K guest chrome in both the library runner and desktop command
- keep `classic-system7` available through `--ui-theme classic-system7` for compatibility and original reference snapshots
- use the Systemless logo palette (`#15130f`, `#c9b183`, `#2794c4`, `#dff6ff`) while preserving classic Macintosh window/control geometry and metrics
- preserve theme colors when architecture-neutral bitmaps are blitted into indexed 68K framebuffers
- cover desktop dithering, window title chrome, controls, scrollbars, menus, dialogs, window stacking, resizing, and repaint behavior
- simplify modal and alert frames to one dark outline so blue is reserved for focus and default-action emphasis
- color every title-bar pinstripe consistently, including short runs beside window controls
- give default buttons a three-pixel blue emphasis ring with the classic one-pixel separator
- fully fill tracked menu rows in blue, reverse their content for contrast, and preserve complete pane borders
- replace ambiguous scrollbar diamonds with directional arrows and use logo blue for resting thumbs
- add a deterministic, losslessly optimized 43-frame Toolbox Showcase theme gallery

The themed presentation currently targets the 68K HLE chrome. Native PowerPC guest chrome remains classic until the shared provider is threaded through that adapter.

## Visual review

[Browse all 43 themed Toolbox Showcase frames](https://github.com/benletchford/systemless/tree/dev/systemless-theme-experiment/tests/toolbox-showcase/reference/systemless-theme-68k)

| Windows and z-order | Controls and scrollbar | Dialog | Popup tracking |
| --- | --- | --- | --- |
| Themed overlapping windows | Themed controls | Themed modal dialog | Themed popup tracking |

## Verification

- `cargo check --lib --bin systemless`
- `cargo test --lib --features test-support` — 4,915 passed, 3 ignored
- `cargo test --bin systemless cli_` — 8 passed
- `cargo test --locked --test toolbox_showcase` — classic 68K 43-frame baseline
- `SYSTEMLESS_PREFER_POWERPC=1 cargo test --locked --test toolbox_showcase` — classic PowerPC gallery
- `SYSTEMLESS_TOOLBOX_THEME=systemless-default cargo test --locked --test toolbox_showcase` — Systemless 43-frame gallery

Closes #1361

PR #1364, opened 2026-09-04
feat: establish unified Macintosh runtime foundations

benletchford:
## Summary

- document the refined one-process architecture, ownership laws, audited gap matrix, migration order, and deletion gates
- add a single address-routing authority for both CPU memory contracts, including atomic mixed and read-only scalar and bulk writes
- add CPU-free task-owned continuation identity and lifecycle validation, plus transactional cross-ISA completion and callback timing fixes
- converge TickCount, ordinary NewHandle, and Trap Manager walking slices on shared process authorities
- require explicit system-memory provenance for privileged permanent trap chains on both ABI paths
- add a CI ratchet that prevents adapter state, trampoline machinery, and direct native-call construction from growing

## Audit status

This is the foundation increment for the tracking issue, not the end of the migration. The migration contract records the remaining concrete work: kernel-owned context banks, resumable MenuSelect and MDEF effects, one CopyBits transfer engine, manager dependency waves, and the final MacintoshProcess composition root. The tracking issue remains open until those deletion gates are complete.

## Verification

- cargo check --lib --no-default-features
- cargo check --all-features
- cargo test --lib --no-default-features --features test-support -- --quiet: 4,932 passed, 3 ignored, 0 failed after rebasing onto public master
- scripts/check-unified-runtime-guardrails.sh
- shell syntax and diff whitespace checks

Part of #1363.


rlanday:
Heads-up: while re-baselining perf work against current `master` we found two host-work regressions that trace to this PR (`git log -S`), and we will send fixes.

1. `TrapDispatcher::trap_table_address` now calls `bus.protected_code_ownership()`, which clones `readonly_code_ranges` into a `ProtectedCodeOwnership` on every trap dispatch. On a 3 in Three headless replay (200 M guest instructions, `sample` at 1 ms), `trap_table_address` is 22% of self time and the `memmove` under it another 6%, plus the allocator traffic (`madvise`, `bzero`). Before this PR the lookup was two bus reads. The fix is a borrowed check (`bus.protected_code_contains(address)`) for the immutable-bus callers; the snapshot stays for `install_trap_address`, which needs the mutable bus.
2. `MacMemoryBus::route` is now on every `read_word`/`read_long` and shows as 6.5% of self time out of line; we are looking at a fast path for the flat-RAM case.

Scale, with #1220 rebased on `master` so its single-step effect is excluded: `master` executes about 3× the host instructions of the pre-#1364 tree on 3 in Three (312 B vs 103 B for the same 200 M guest instructions, identical ticks and screenshots) and 3.8× on SimCity 2000 (192 B vs 51 B per 400 M). A bisect over 3c63ca4..c40588f is running to confirm the split between this PR and #1362; the profile attribution above is from the symbol level.

rlanday:
Follow-up with the fixes and the confirmed attribution.

Bisect (3 in Three headless replay, 200 M guest instructions, host instructions retired, single runs): 5d87791 = 105.9 B; eef89a3 (this PR) = 268.0 B; 305027d (#1362) = 308.6 B; 83af901 = 308.5 B.

Fixes, all against `master` with #1220 rebased on it so the single-step mode is excluded:
- #1419: `trap_table_address` no longer clones the protected-code ranges per dispatch, and the range list is kept sorted and merged so `is_protected_head`, `readonly_code_contains` and `readonly_code_overlaps` are binary searches instead of linear scans over ~1,000 entries.
- #1420: `MacMemoryBus::route` answers the flat-RAM case inline when no foreign address space is attached.

Together with #1421 (the #1362 side), machine idle, 5 pairs each, identical ticks and framebuffer hashes: 3 in Three −60.7% host instructions and −53.0% CPU time (35.4 s → 16.6 s); SimCity 2000 −39.1% and −30.8% (18.6 s → 12.9 s). The pre-#1364 tree is 105.9 B on 3 in Three; the fixed stack is 122.7 B.

PR #1419, opened 2026-09-05
perf(trap): look up trap table addresses without copying or scanning the protected-code ranges

rlanday:
## Summary

Two fixes to the per-dispatch trap-table lookup that #1364 introduced, keeping its provenance check intact:

1. `trap_table_address` snapshotted the bus's protected-code ownership on every trap dispatch, cloning the `readonly_code_ranges` `Vec` so the check closure could outlive an immutable borrow it did not need (the whole lookup holds the bus immutably). It now answers the check from the live ranges through a borrowed accessor; the snapshot stays for `install_trap_address`, whose memory closure needs the mutable bus.
2. The protected-code range list is queried by `is_protected_head` on every hop of every trap-table chain walk, by `readonly_code_contains` on protected writes, and by `readonly_code_overlaps` on every guest write. Every materialized trap slot registers a range (trampoline, gateway, come-from head), so the list holds on the order of a thousand entries and each query was a linear scan. The list is now kept sorted by start with overlapping and touching entries merged on insert, and all three queries are one binary search. Answers are unchanged: coverage of a contiguous interval by the union of ranges is coverage by the single merged range containing its start, and the last range starting before an interval's end has the largest stop. A unit test checks merged insertion and compares both queries against a brute-force union over every interval in a synthetic layout.

## Why

Re-baselining perf work against current `master` showed it executing several times the host work of a tree based on 3c63ca4. Bisect (3 in Three headless replay, 200 M guest instructions, single runs): 5d87791 = 105.9 B host instructions, eef89a3 (#1364) = 268.0 B, 305027d (#1362) = 308.6 B, 83af901 = 308.5 B. A `sample` profile of `master` (with #1220 rebased on it, so the single-step effect is excluded) put `trap_table_address` at 22% of self time with another 6% in the `memmove` under it.

## Measurements

Paired hardware counters on the headless replay (`--max-instructions`, scripted input where the game needs it), both arms interleaved, identical guest instructions, ticks and framebuffer hash on every pair; the medians of the pair ratios:

| step | 3 in Three, 200 M: instructions / CPU time | SimCity 2000, 400 M: instructions / CPU time |
|---|---|---|
| no clone (fix 1) vs rebased #1220 | −3.9% (5/5) / per-step CPU time to follow in a comment | −1.8% (5/5) / to follow |
| sorted ranges (fix 2 here) on top of the chrome and routing fixes | **−47.7%** (5/5) / to follow | **−20.3%** (5/5) / to follow |
| all four fixes together vs rebased #1220 | **−60.7%** instructions, **−53.0%** CPU time (35.4 s → 16.6 s), 5/5 | **−39.1%** instructions, **−30.8%** CPU time (18.6 s → 12.9 s), 5/5 |

Absolute, 3 in Three: 246.7 B → 128.9 B host instructions for the second step. Cycles and CPU time moved the same direction but the machine was loaded during these runs, so the instruction counts are the numbers to trust.

Together with the routing fast path and the themed-chrome cache (separate PRs), 3 in Three goes from 308.5 B on `master` to 128.9 B; the pre-#1364 tree is 105.9 B.

Measured on builds based on c40588f, the `master` of the day; the branch is rebased onto d30234c, whose seven commits do not touch these paths.

The per-step instruction counts above were measured while the machine was loaded (cycles and CPU time were noise); the combined row was re-measured with the machine idle, and per-step CPU times from the same idle rerun follow in a comment.

Related fixes for the same regression: #1419 (trap-table lookup), #1420 (routing fast path), #1421 (themed chrome cache); #1220 removes the single-step mode they were measured on top of.

## Tests

`cargo test`: the new unit test plus the existing suite; one failure, `step_frame_forces_render_after_same_tick_foreground_progress`, fails identically on untouched `master` (c40588f) here.

https://claude.ai/code/session_01UqCuD9vsGeeij5DdYt3vcK

### DIFF
--- a/src/memory/bus.rs
+++ b/src/memory/bus.rs
@@ -543,26 +543,47 @@ fn protected_code_covers_long(addressing_32_bit: bool, ranges: &[(u32, u32)], ad
     } else {
         address & 0x00FF_FFFF
     };
-    let end = (u64::from(translated) + 4).min(1u64 << 32);
-    if end != u64::from(translated) + 4 {
-        return false;
+    let end = u64::from(translated) + 4;
+    end <= (1u64 << 32) && protected_ranges_cover(ranges, u64::from(translated), end)
+}
+
+/// Insert `[start, end)` into a list kept sorted by start with no overlapping
+/// or touching entries, merging as needed. Every materialized trap slot
+/// registers a range (trampoline, gateway, come-from head), so the list
+/// grows to about a thousand entries and every trap dispatch queries it.
+fn insert_protected_range(ranges: &mut Vec<(u32, u32)>, start: u32, end: u32) {
+    let mut lo = ranges.partition_point(|&(s, _)| s < start);
+    if lo > 0 && ranges[lo - 1].1 >= start {
+        lo -= 1;
     }
-    let mut cursor = u64::from(translated);
-    while cursor < end {
-        let mut covered_end = cursor;
-        for &(start, stop) in ranges {
-            let start = u64::from(start);
-            let stop = u64::from(stop);
-            if start <= cursor && cursor < stop {
-                covered_end = covered_end.max(stop.min(end));
-            }
-        }
-        if covered_end == cursor {
-            return false;
-        }
-        cursor = covered_end;
+    let mut merged = (start, end);
+    let mut hi = lo;
+    while hi < ranges.len() && ranges[hi].0 <= merged.1 {
+        merged.0 = merged.0.min(ranges[hi].0);
+        merged.1 = merged.1.max(ranges[hi].1);
+        hi += 1;
     }
-    true
+    ranges.splice(lo..hi, [merged]);
+}
+
+/// Whether `[address, end)` lies entirely inside the protected ranges. With
+/// sorted, merged ranges that is true exactly when the last range starting
+/// at or before `address` reaches `end`.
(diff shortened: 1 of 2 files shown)


rlanday:

|---|---|---|

benletchford (approved):
Reviewed the range insertion and binary-search invariants, all range writers, 24-bit translation, and protected longword boundaries. The immutable borrow removes the dispatch-time copy without changing ownership checks. Tested this together with #1420 and #1220 on current master: 4,987 library tests passed, three ignored; CI is green.

rlanday:
This PR's two commits on their own (on rebased #1220, against rebased #1220), 5 pairs, identical ticks and framebuffer hash on every pair:

| workload | host instructions | cycles | CPU time |
|---|---|---|---|
| 3 in Three, 200 M | **−41.6%** (5/5) | −34.8% (5/5) | **−35.0%** (5/5, 29.7 s → 20.1 s) |
| SimCity 2000, 400 M | **−18.8%** (5/5) | −15.6% (5/5) | **−15.7%** (5/5, 19.9 s → 16.8 s) |

The 3 in Three pairs ran on an idle machine; during the SimCity 2000 pairs macOS PDF indexing was using about 1.6 cores, so treat that CPU-time figure as slightly pessimistic.

Issue #174, opened 2026-07-30
Encode valid indexed screen snapshots in ClosePicture


benletchford:
`ClosePicture` currently serializes its retained screen image as a 1-bit bitmap regardless of the active indexed display depth. At 2-bit and 4-bit depths this truncates packed rows and can produce malformed picture data that `DrawPicture` cannot replay reliably.

Teach the picture recorder to emit a valid indexed PICT snapshot with the active color table and cover a 4-bit `ClosePicture`/`DrawPicture` round trip.

PR #1421, opened 2026-09-05
perf(chrome): replay themed dialog frames, the menu bar and menu titles from a pixel cache

rlanday:
## Summary

Cache resolved theme artwork for dialog frames, menu-bar backgrounds and
menu-title backgrounds instead of rebuilding their `ThemeBitmap`s every
chrome redraw. The cache holds up to 24 entries with least-recently-used
replacement. It is used by the non-classic themes.

**September 8 refresh:** this branch incorporates upstream `b5f4f083` (0.39.1)
through the refreshed #1423 branch, plus a new retained-text equivalence test.
It remains stacked on #1423; the diff includes that mirror's commits until it
merges. Fresh CPU acceptance on this upstream is still required.

|---|---|---|

## What is cached, and what is not

Keys cover artwork kind, rectangles/state, theme, screen geometry and depth.
Each cached rendering also remembers the colours it used and their resolved
device indices. Replay is allowed only if those colours still resolve to the
same indices through #1423's current device-table mirror.

This is **not `ctSeed`-only validation**. Same-seed table rewrites, depth changes
and the host CLUT fallback are checked. A palette change unrelated to the used
colours need not force a new bitmap rendering.

**Title glyphs are not part of this cache.** `draw_theme_menu_title_chrome`
draws the title's themed background; actual font text is painted separately.
Both fresh artwork and cached pixels go through `fb_set_pixel_index`, preserving
the bus's write observers and retained-text erases, including same-value writes.
Caching device indices does not authorize bypassing those observers.

`fill_theme_rect` also resolves its colour through the mirror and requests
row fills. With current retained presentation enabled, those requests still
fall back to observed per-byte writes; this PR alone does not remove that
observer cost. A separate bulk-observer prototype is being evaluated locally.

## Rejected approaches and motivation

The original seed-keyed validation missed same-seed palette changes. A later
whole-environment digest cache key was correct but caused frequent re-rendering
under palette animation. The current used-colour validation avoids that
unnecessary invalidation while checking every colour the artwork depends on.

Historical profiling during the #1362 re-baselining found dialog-frame bitmap
blits at 7.5% self time and menu-bar/title-background blits at another 4.1% in
a 3 in Three replay. These are old workload-specific samples, not current
profile shares or predictions of the remaining opportunity.

## Historical measurements — not current-head acceptance

Five interleaved pairs against `5f21681` (then master, already carrying #1220,
#1419 and #1420), **including #1423**, using the older instruction-quota
headless runner. These are not an isolated measurement of this cache or the
corrected fixed-time runner in #1596, and not GUI CPU/FPS measurements.

| workload | host instructions | cycles | CPU time |
|---|---|---|---|
| SimCity 2000, 400 M | −23.0% (5/5) | −20.4% (5/5) | −20.8% (5/5, 12.0 s → 9.5 s) |
| 3 in Three, 200 M | −19.5% (5/5) | −20.2% (5/5) | −19.0% (5/5, 16.0 s → 13.0 s) |

The old comparison against the preceding digest-keyed version reported SC2K
−18.7% instructions / −17.1% CPU and a flat result on 3 in Three, consistent
with palette animation affecting invalidation. It does not establish the
size of that effect after upstream's later presentation/scheduling changes.
Equal ticks and logical framebuffer hashes were reported for those pairs;
checkpoint equality is not proof of all transient or retained high-resolution
output. The original logs and rejected versions remain retained locally.

## Tests and review status

Existing tests compare cached and fresh screen bytes, unused same-seed table
rewrites, active/inactive cache entries, colour-to-grayscale changes with an
unchanged seed, and the host-table fallback with `MainDevice` nil. An unused
entry rewrite should retain the rendering, not force a fresh one.

The new regression paints retained outline text across dialog-frame, menu-bar
and menu-title artwork, then compares cached replay with a forced fresh render:
guest bytes **and saved subpixel detail**, resolved RGB and subsequent title
painting must match. It covers the distinction between cached backgrounds and
separately painted glyphs. [CI on `3a8fcc16`](https://github.com/benletchford/systemless/actions/runs/34200352443)
passed Linux build/tests, classic/PowerPC showcase assertions, headless/package,
licenses and the nonblocking formatting/clippy job. The test-only follow-up
`3cdd2547` strengthens the fixture precondition: its saved snapshot must contain
actual retained coverage, not merely grey logical pixels. [CI on that final
head](https://github.com/benletchford/systemless/actions/runs/34201318954) also
passes all four jobs, including the strengthened regression and classic/PowerPC
showcase assertions.

Earlier review requested changes; the fixes have not yet received approval.
No new performance percentage is
claimed until the final implementation is measured on a current, equal-time
baseline.

Original implementation/review history: [Claude session](https://claude.ai/code/session_01UqCuD9vsGeeij5DdYt3vcK).

### DIFF
--- a/src/trap/dispatch.rs
+++ b/src/trap/dispatch.rs
@@ -1964,6 +1964,9 @@ pub struct TrapDispatcher {
     pub(crate) color_mirror: std::cell::RefCell<super::framebuffer::ColorTableMirror>,
     /// Set for the duration of a chrome pass, whose lookups reuse the mirror.
     pub(crate) color_mirror_fresh: std::cell::Cell<bool>,
+    /// Themed chrome rendered earlier, replayed while its inputs hold; see
+    /// `ThemeChromeCacheEntry`.
+    pub(crate) theme_chrome_cache: std::cell::RefCell<Vec<super::framebuffer::ThemeChromeCacheEntry>>,
     /// Color tables produced from palettes whose entries are all pmExplicit.
     /// Their pixel values are literal device indices, so indexed CopyBits
     /// must preserve those values instead of color-matching duplicate RGBs.
@@ -3519,6 +3522,7 @@ impl TrapDispatcher {
             menu_mark_indices: std::cell::Cell::new(None),
             color_mirror: std::cell::RefCell::new(Default::default()),
             color_mirror_fresh: std::cell::Cell::new(false),
+            theme_chrome_cache: std::cell::RefCell::new(Vec::new()),
             explicit_palette_ctabs: HashSet::new(),
             icon_transform_override: 0,
             printing_error: 0,
--- a/src/trap/framebuffer.rs
+++ b/src/trap/framebuffer.rs
@@ -108,6 +108,58 @@ pub(crate) struct MenuMarkIndexCache {
     pub(crate) indices: [u8; crate::ui_art::RETRO_COMPUTER_MENU_MARK_PALETTE.len()],
 }
 
+/// Which piece of themed chrome a cached rendering is, with the inputs its
+/// artwork depends on.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub(crate) enum ThemeChromeKind {
+    DialogFrame {
+        content: (i16, i16, i16, i16),
+        frame: (i16, i16, i16, i16),
+        proc_id: i16,
+        active: bool,
+        fill_content: bool,
+    },
+    MenuBar {
+        width: i16,
+        height: i16,
+    },
+    MenuTitle {
+        rect: (i16, i16, i16, i16),
+        enabled: bool,
+        highlighted: bool,
+    },
+}
+
+/// Themed chrome rendered earlier, as device pixel indices, keyed by
+/// everything the rendering depends on. The chrome redraw runs every host
+/// frame and its pieces rarely change between frames, so replaying resolved
+/// pixels replaces rendering a bitmap, resolving its colours through the
+/// device table and scanning every pixel again.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub(crate) struct ThemeChromeCacheKey {
+    pub(crate) kind: ThemeChromeKind,
+    pub(crate) theme: crate::ui_theme::UiThemeId,
+    /// Screen geometry and depth: the artwork is laid out against them.
+    pub(crate) screen: (u32, u32, i16, i16, u16),
+}
+
+#[derive(Clone, Debug)]
+pub(crate) struct ThemeChromeCacheEntry {
+    pub(crate) key: ThemeChromeCacheKey,
+    /// The artwork's colours and the device indices they resolved to when
+    /// it was rendered. The entry is replayed only while every one of them
+    /// still resolves to the same index, so a palette change that touches
+    /// none of them keeps the rendering, and one that does re-renders it.
+    pub(crate) resolved: Vec<(Rgb8, u8)>,
+    /// Pixels as (x, y, device index), in blit order; only the opaque ones
+    /// for masked artwork.
+    pub(crate) pixels: Vec<(i16, i16, u8)>,
+}
+
+/// Enough entries for a menu bar, its titles and a dialog frame; the
+/// least recently used entry is evicted beyond that.
+const THEME_CHROME_CACHE_ENTRIES: usize = 24;
+
 impl super::TrapDispatcher {
     /// Read screen parameters from the dispatcher's screen_mode.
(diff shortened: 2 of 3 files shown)


rlanday:

|---|---|---|

benletchford (changes requested):
The rendered-pixel cache has a reproducible invalidation bug. `install_standard_depth_clut` rewrites the main colour table in place and sets its seed to the depth, so a colour-to-grayscale SetDepth change can keep the screen tuple, table pointer, and seed unchanged while replacing all RGB entries.

I tested this on this PR head: select colour 8 bpp, draw the themed menu bar to warm the cache, select grayscale 8 bpp (SetDepth: whichFlags=1, flags=0), draw again, and compare with a fresh rendering after clearing `theme_chrome_cache`. The cached and fresh framebuffers differ. This is a supported mode change, not a guest violating the colour-table protocol.

The key also omits the fallback `device_clut`: when no main-device table is available, `theme_pixel_index` reads that host palette, but the cache records only `ctab: None`.

Please cover these paths with reliable invalidation (or decline to cache when the dependency cannot be tracked), and add cached-versus-fresh regressions for both cases. The performance improvement is worthwhile, but the current implementation can keep stale chrome colours, so I am leaving this open with changes requested.

rlanday:
Thanks, both findings were real and your repro was exact. Fixed and rebased:

- The branch is now stacked on #1423 (the diff includes its two commits until it merges) because both caches share the fix: `color_environment_digest`, a digest of everything colour resolution reads (screen depth, the main device table's pointer and every entry via one bulk read, and the host-side `device_clut` used when no device table exists). The chrome cache key carries that digest instead of the table pointer and `ctSeed`, so a same-seed rewrite such as `install_standard_depth_clut` and a change on the fallback path both miss the cache.
- Regression tests follow your steps: draw the themed menu bar at colour 8 bpp, `install_standard_depth_clut(8, grayscale)` with an assertion that the seed is unchanged, draw again through the cache, then compare against a fresh rendering after clearing `theme_chrome_cache`; a second test does the same with `MainDevice` nil and a `device_clut` change. The original test now rewrites a table entry in place with the seed untouched.
- The two original commits are squashed into one (dialog frame, menu bar and titles), with the fix as a second commit for review.

Rebased onto current `master` (5f21681). Suite result in the next comment.

Suite on the pushed head: 5,069 pass. One failure, `test_toolbox_showcase` at `27-sprites-animated.png` (601 pixels), fails identically on untouched `master` (5f21681) on this machine right now, while the same branch passed it an hour earlier; the themed chrome code is not reached in that test (the showcase runs the classic theme), so it looks environmental here rather than related, but flagging it in case you see it too.

rlanday:
Follow-up, replacing the digest key after measuring it on SimCity 2000: that game animates palette entries many times a second, so a whole-environment key re-rendered the chrome nearly every frame (13.5% of the run in chrome work, worse than before the fix). Pushed a rebuilt branch on top of #1423's mirror: commit 1 is the chrome cache, commit 2 records, per cached rendering, the colours it used and the indices they resolved to, and replays only while each still resolves to the same index through the mirror. Your two repros pass unchanged, and a new test checks that a palette change touching only unused entries keeps the rendering while one touching a used colour re-renders identically to a fresh draw. Commit 2 also stops `fill_theme_rect` (themed window chrome, every frame) from scanning the table per call and writing one routed byte per pixel.

Numbers against current `master` with #1423 included, 5 pairs, identical framebuffer hashes: SimCity 2000 −23.0% host instructions / −20.8% CPU time; 3 in Three −19.5% / −19.0%. Against the previous head of this PR: SimCity 2000 −18.7% instructions / −17.1% CPU time; 3 in Three flat.

Suite on the pushed head: 5,067 pass; the one failure is the `test_toolbox_showcase` frame that fails identically on untouched `master` on this machine (the showcase runs the classic theme, which never reaches this code).

benletchford:
Resolved the current-master conflict by retaining both required memory imports. #1423 has now landed, so I rebased this branch to drop that prerequisite and retain only the theme-cache delta. The branch is now one conventional commit; new-head CI is running. The resulting source tree is identical to the locally tested combined palette/theme tree. Historical benchmark percentages remain historical, not a current-master performance rerun.

benletchford:
Fresh combined validation with the current palette mirror, sound-finalization, simulated-time headless, and theme-cache changes: 5,437 library tests passed (3 ignored), 98 macOS desktop tests passed (1 ignored), including the Metal minification test, and wasm32-unknown-unknown library check passed. The complete source diff is whitespace-clean. Waiting for the refreshed PR CI showcase checks before merging; no new performance percentage is claimed.

PR #1423, opened 2026-09-05
perf(menu): resolve menu-mark colors through a device-table mirror

rlanday:
## Summary

Resolve the retro computer menu mark's four colours against a host-side mirror
of the main device colour table, preserving the original bus scan's answers.

**September 8 refresh:** `d991e12c` incorporates upstream `b5f4f083` (0.39.1),
including the retained outline presentation and foreground/nested callback work.
No new CPU benefit is claimed for this refreshed implementation yet.

|---|---|---|

## Current implementation and review fixes

The dispatcher refreshes the mirror at most once per chrome pass. Its digest
covers screen depth, table pointer, header and every entry, plus the host
`device_clut` fallback. A changed entry is detected even when `ctSeed` stays the
same. Lookups preserve black/white shortcuts, the first exact match in table
order, and otherwise the nearest entry with the lowest index; device-table and
ordinal-index modes remain distinct.

**The four menu-mark indices are resolved against the mirror on every call.**
The `menu_mark_indices` cell stores the latest result for diagnostics/tests;
production does not serve subsequent calls from that cell. This is a table
mirror, not the seed-keyed four-result cache described by the old PR title.

Rejected approaches, retained in commit history:

- A `ctSeed`-only result cache was incorrect: Systemless's depth switch reuses
  the seed, and table entries can be rewritten in place.
- A full environment digest per lookup fixes invalidation but repeats work
  during palette animation. The current mirror bounds digest refresh to once
  per chrome pass and serves colour resolution from host memory.

## Historical measurements — not current-head acceptance

The following measurements are preserved from the older headless experiments
on the `3c63ca4`-based stack carrying #1220. They predate the current mirror and
the corrected fixed-simulated-time frontend scheduling in #1596. They must not
be read as measurements of `d991e12c`, current GUI CPU usage, or the isolated
benefit of the final post-review implementation.

| workload | host instructions (5–6 pairs) | CPU time, quiet rerun |
|---|---|---|
| SimCity 2000, 400 M | −1.75% (6/6) | −0.9% (4/5); instructions −1.8% (5/5) in the same rerun |
| Lemmings, 1 B | −2.25% (6/6) | −21.7% (5/5, 15.7 s → 11.1 s); instructions −2.3% (5/5), cycles −15.0% (5/5) |
| 3 in Three, 200 M | −0.44% (5/5) | — |
| System's Twilight, 100 M | −0.09% (6/6) | — |
| Escape Velocity, 100 M | +0.06% (2/6) | — |

The older report recorded equal ticks/framebuffer checkpoints except Escape
Velocity's nondeterministic replay. Those checks do not establish retained
high-resolution text equivalence. The unusually large Lemmings CPU delta is a
historical observation, not a demonstrated mechanism or prediction for this
mirror. Fresh equal-time acceptance is still needed.

## Tests and review status

Current regressions compare mirror answers against the bus scan with duplicate
colours, out-of-order indices and both table modes, and exercise same-seed
entry rewrites, colour/grayscale depth changes, and the host-table fallback.
The old description of poisoning a result cache and bumping its seed no longer
describes the current test or implementation. #1421 builds on this mirror and
adds cached-versus-fresh artwork tests.

[Refreshed-head CI on `d991e12c`](https://github.com/benletchford/systemless/actions/runs/34200063368)
passes Linux build/tests, headless/package, licenses and the nonblocking
formatting/clippy job. Earlier review requested changes; the subsequent fixes
have not yet received approval. Green correctness checks are not a new CPU
measurement.

Original implementation/review history: [Claude session](https://claude.ai/code/session_01UqCuD9vsGeeij5DdYt3vcK).


benletchford (changes requested):
The cache has a reproducible invalidation bug on the existing SetDepth path. `install_standard_depth_clut` rewrites the main table in place and stamps `ctSeed` with the depth, so switching from colour to grayscale at 8 bpp preserves both parts of this cache key while changing RGB entries.

I reproduced this on this PR head: warm `menu_mark_palette_indices`, dispatch SetDepth with depth 8, whichFlags 1 and flags 0, then compare the cached result with the result after clearing `menu_mark_indices`. The cached indices are `[253, 44, 157, 36]`; fresh resolution returns `[221, 79, 125, 11]`.

Please use an invalidation mechanism that covers same-depth palette/personality changes and add this regression. The current test only increments the seed itself, so it cannot catch writers that reuse a seed. The optimization is useful, but it should not merge with stale menu colours after a supported display-mode change.

rlanday:
Thanks, that is a real bug and your repro was exact. Fixed in e4bf537 on top of the rebase:

- The cache no longer trusts `ctSeed`. It is keyed on a digest of everything the lookup reads: the screen depth, the main device table's pointer and every entry (one `read_bytes_into` of the table), and the host-side `device_clut` that the lookup falls back to when there is no device table. A same-seed rewrite such as `install_standard_depth_clut` changes the digest; so does a change on the fallback path. One bulk read per redraw is still far cheaper than the four 256-entry scans the cache removes.
- Regression tests follow your steps: warm at colour 8 bpp, `install_standard_depth_clut(8, grayscale)` with an assertion that the seed is unchanged, then compare the cached path against a fresh resolution after clearing the cache; a second test does the same with `MainDevice` nil and a `device_clut` change. The original test now rewrites a table entry in place with the seed untouched instead of bumping the seed itself.

Rebased onto current `master` (5f21681). Suite: 5,065 pass; the one failure is `step_frame_forces_render_after_same_tick_foreground_progress`, which fails identically on untouched master here (wall-clock frame setup under load).

rlanday:
Follow-up, after measuring the digest fix on SimCity 2000: that game animates palette entries many times a second, so a whole-environment key re-resolved almost every frame and paid for the digest per lookup. Pushed 05e1555 on top: a host-side mirror of the main device's colour table, refreshed when the environment digest changes (at most once per chrome pass), with the mark's colours resolved against it using exactly the bus scan's semantics (a unit test compares the two on a table with duplicates, out-of-order values and both device-table modes). Your regression tests pass unchanged. #1421 builds on the mirror for the chrome cache; its numbers are in that PR.

Suite on the pushed head: 5,067 pass; the one failure is the `test_toolbox_showcase` frame that fails identically on untouched `master` on this machine.

PR #1433, opened 2026-09-05
fix(desktop): size game windows for the display automatically

### LINKED ISSUE #1432: fix(desktop): choose comfortable display-aware window sizes
Systemless opens a very small game window on high-DPI Macs. For example, launching a 640×480 game without display options creates a 640×480 physical-pixel client area, which is only 320×240 logical points on a 2× Retina display. Low-resolution games are also small on other displays.

The default --display-scale=1 currently forces a physical 1:1 mapping. Screen-mode changes reapply that default, and learned viewport crops round the window down to an integer physical scale, so simply enlarging the initial window is insufficient.

Expected: without any flags, launch at a comfortable display-aware size, constrained to the monitor, preserving aspect ratio. Keep explicit --display-scale as the exact physical-pixel override. Handle guest resolution/crop changes without reverting to tiny windows, and avoid resizing fullscreen or maximized windows for guest layout changes. Preserve user sizing and DPI transitions.

benletchford:
A 640×480 game previously opened in 640×480 physical pixels: only 320×240 points on a Retina Mac. The default now fits the game's aspect ratio into a 960×720 logical-point box, capped at 80% of the display's available bounds. macOS uses the actual display's visible frame to account for the menu bar and Dock; the window is sized and centered before it is shown.

Automatic mode keeps the host window through guest resolution, learned-crop, and dialog changes, fitting the guest presentation into it instead of shrinking the window or undoing manual resizing. Explicit `--display-scale 1..8` retains physical-pixel sizing. Guest layout changes do not resize fullscreen or maximized windows. DPI transitions retain winit's native logical-size behavior.

Validation: all 83 desktop tests pass on ARM64 macOS, including new cases for 1×/1.25×/1.5×/2×/3× DPI, multiple guest resolutions, small/portrait monitors, unavailable monitor information, invalid DPI, explicit scale, and fullscreen/maximized resize policy. Built the desktop binary and visually verified Marathon Infinity's default launch with its cached viewport on a Retina Mac. CI now runs desktop tests as well as building them; fresh Linux checks are pending.

Closes #1432

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,6 +62,9 @@ jobs:
       - name: Run library tests
         run: cargo test --lib --features test-support
 
+      - name: Run desktop tests
+        run: cargo test --locked --bin systemless
+
       - name: Run toolbox showcase integration test (68k)
         run: cargo test --test toolbox_showcase
 
--- a/README.md
+++ b/README.md
@@ -221,9 +221,11 @@ systemless --ui-theme classic-system7 path/to/game.sit
 systemless --fullscreen path/to/game.sit
 ```
 
-Desktop windows default to a physical 1:1 guest-to-host pixel ratio. Use
-`--display-scale` with an integer from 1 through 8 for explicit pixel-perfect
-enlargement. `--fullscreen` starts the guest in a borderless fullscreen space.
+Desktop windows automatically open at a comfortable size for the display,
+including Retina and other high-DPI screens. Games keep their aspect ratio,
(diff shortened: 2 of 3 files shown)


benletchford:
The library and desktop tests passed on Linux, alongside 83 passing desktop tests and a visual Marathon launch on ARM64 macOS. The remaining CI failure is the exact 601-pixel animated-sprite reference mismatch already present on base commit cc7063cd (run 33943876506), tracked separately in #1436.

PR #1593, opened 2026-09-07
fix: package current bundled font licences in release archives

### LINKED ISSUE #1592: Release archives fail to package bundled font licences
The v0.39.0 release workflow builds binaries successfully on all six targets, then fails while copying the removed root OFL.txt. Release archives should include the current URW and Noto licence texts and attribution files. Reproduction: https://github.com/benletchford/systemless/actions/runs/34125344529

benletchford:
Release v0.39.0 builds successfully on all six targets but packaging fails because it copies the removed root OFL.txt. Include the current URW and Noto licence texts and attribution documents under distinct archive filenames, and verify they survive extraction. ZIP and tar round-trip checks preserve all seven documentation files byte-for-byte. Closes #1592.

### DIFF
--- a/.github/workflows/release-binaries.yml
+++ b/.github/workflows/release-binaries.yml
@@ -108,8 +108,17 @@ jobs:
               package.mkdir()
               for binary in binaries:
                   shutil.copy2(pathlib.Path("target") / target / "release" / binary, package)
-              for filename in ("LICENSE", "OFL.txt", "README.md"):
-                  shutil.copy2(filename, package)
+              documents = {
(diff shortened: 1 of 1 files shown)


benletchford:
Verified the repaired release workflow on all six Linux, macOS and Windows targets, including x86_64 and aarch64. Every build, archive extraction and executable smoke test passed. All six v0.39.0 archives and SHA-256 files are now attached to the existing release. Successful run: https://github.com/benletchford/systemless/actions/runs/34126982032

PR #1524, opened 2026-09-06
feat(fonts)!: replace hardcoded fonts with bundled TrueType defaults

benletchford:
Replace the hardcoded font catalogue with bundled URW Core 35 TrueType fonts and Noto Sans Symbols 2. Both 68K and PowerPC retain sharp text through shared Rust rendering, snapshots and supported pixel transfers. Guest fonts and explicit overrides retain precedence; font assets, licences and provenance are included.

### Rendering and performance

4× is retained glyph detail, not a mandatory output frame size. Desktop output selects the smallest integer scale covering the actual drawable, resolves directly to that size, and caches the result until drawing or the palette changes. Browser RGBA and desktop ARGB share the resolver; Metal and software output preserve coverage during fractional reduction. Ordinary frames without visible retained detail keep the existing bitmap path.

Modal filters that draw nothing reuse their completed dialog snapshot. Snapshots share immutable detail until modified, and repeated restores skip unchanged contents. An unchanged palette check does not invalidate these caches. Actual drawing, palette changes, same-byte erases over text, copies and snapshot edits still invalidate or update retained coverage.

The EV Override registration bug was a separate layout error: its condensed heading contaminated measurement of later plain text. Styled TextEdit now measures each run with its own font, size and face. Standard push button corners and default rings use shared coverage on indexed and direct-color surfaces, preserving guest pixels, hit rectangles and clipping. Text embedded in application PICT artwork remains bitmap content.

Local Apple M1 measurements, optimized `ci-test` profile, 1,000 warm EV Override slices:

| Phase | Current master, logical output | This revision, 2× output |
| --- | --- | --- |
| Execution mean / p95 | 0.30 / 0.41 ms | 0.62 / 0.90 ms |
| Additional composition mean | 0.006 ms | 0.0005 ms |
| Presentation mean / p95 | 0.21 / 0.24 ms | 0.82 / 0.89 ms |

[...]

PR #1596, opened 2026-09-07
fix(headless): replay simulated time with GUI wait scheduling

rlanday:
# Drive headless replays by simulated frontend time and GUI wait scheduling

## Summary

Add `--max-ticks` and `--tick-input-script` for windowless, deterministic-input
replays at fixed simulated frontend time. Reuse the desktop's retained-wait,
timer, sound-callback and execution-rate policies. Preserve the existing
instruction-quota mode as an explicitly labeled diagnostic mode.

This corrects a performance-measurement artifact; it does **not** claim a
windowed CPU improvement, change the guest CPU timing model, or tune the JIT.
Originally based on master `b4bbe723`, including the merged font/presentation work (#1524). Normal merge commit `9f984ae6` now incorporates upstream `b5f4f083` / 0.39.1 and #1595's foreground callback/nested-dialog corrections.
No private macOS performance-counter instrumentation or unrelated performance
candidate is included.

**Refreshed-head checks:** [CI on `9f984ae6`](https://github.com/benletchford/systemless/actions/runs/34197977034) passes Linux build/tests, headless/package checks, licenses, and the nonblocking formatting/clippy job. The earlier MPW-download timeout is not a failure on this head. Historical local test counts and replay checkpoints below are not relabeled as new 0.39.1 measurements.

|---|---|---|

## Why instruction-bounded headless runs were misleading

Some Toolbox calls retain a native wait instead of returning immediately to
guest code. ModalDialog is one example: the native handler can rewind the
guest PC and wait for another frontend scheduling opportunity. Desktop
execution already recognizes that condition, services the appropriate time
and audio work, and yields.

The old headless `run_steps` path does not use those GUI yield boundaries.
Given a large instruction quota, it can repeatedly re-enter the retained
wait until that quota is spent. Making each such entry cheaper produces a
real saving per call, but at an artificially high call frequency. That is
not evidence of the same saving in a window or during gameplay. One EV
Override replay intended to reach gameplay actually ended in a retained
ship-name dialog; its previous CPU figures are excluded from gameplay claims.

Instruction-indexed input has a second problem: eliminating guest idle
instructions changes when the same numbered instruction is reached in
simulated time. Equal instruction quotas are useful interpreter diagnostics,
but can penalize legitimate idle skipping or compare different game progress.

The replacement holds frontend time and scripted input constant. Acceptance
still requires checking actual game progress, guest ticks, pictures and audio;
equal frontend ticks alone cannot establish equivalent workloads.

## Scheduling and compatibility

- A tick is 1/60.15 simulated seconds. There is no wall-clock sleep, window,
  compositor, or physical audio device. With no legacy options, headless
  defaults to 600 ticks (about ten simulated seconds).
- Initialize the loaded architecture with the same execution rate as the
  GUI: 415,628 instructions/tick for 68K, 1,995,012 for PPC. Calling GUI slice
  methods alone does not replace the fixture's much smaller default rate.
  The initialization is shared with the GUI so these values cannot diverge.
- Each virtual frame requests the next guest tick through bounded foreground
  CPU batches. A short retained-tracking slice ends foreground work for that
  frame. A nontracking short slice with progress can be an ordinary task/engine
  handoff, so it is allowed to continue within the safety bound. Zero progress
  yields rather than spinning.
- Menu tracking may freeze guest TickCount. The frontend clock, input and
  audio still advance. The menu presentation clock advances once per frame.
- Audio samples retain their fractional remainder between frames. Mixing is
  interleaved with guest work; bounded guest sound-callback servicing and the
  desktop's remaining-audio chunk size are reused. Playback/callbacks still
  run without a host audio device.
- Follow the newly merged deferred-presentation path: `run_gui_cpu_slice`
  does not repaint chrome per batch; audio mixing is separate; text
  presentation preparation and composition occur at the end of the frame.
- Use the existing mouse-release latch so a scripted press/release in one
  tick still gives the guest an opportunity to observe button-down.
- Fix the startup Mac clock/initial seed by default; allow an explicit
  `--headless-start-time` override. Save files still need independent pristine
  copies. This is not a guarantee against every source of game nondeterminism.
- Retain `--max-instructions` / `--input-script` behavior and emit a warning
  describing the retained-wait limitation. Tick and instruction modes cannot
  be mixed or silently reinterpret the other's timestamps. Reject a script
  event at or beyond the chosen endpoint rather than silently dropping it.

The instruction safety bound remains an existing machine scheduling policy,
not a realistic cycle cost for every 68K instruction. A halted run exits with
failure. Reports expose frontend/guest clocks, retired instruction work,
delivered inputs, same-guest-tick frames, budget exhaustion and captured mono
audio sample count/digest. Exhaustion is explicitly warned about; it must be
investigated before using a run for a CPU comparison.

## Usage

```sh
systemless --headless --max-ticks 600 game.sit
systemless --headless --max-ticks 4800 --tick-input-script inputs.txt game.sit
```

Input coordinates are vertical, horizontal. For example:

```text
120 mousedown 317 491
122 mouseup 317 491
```

These are elapsed **frontend ticks**, not guest TickCount or host wall time.
Use game-specific, visually verified inputs and fresh copies of the same
archive/saves on both sides of a comparison.

## Verification

**September 8 refresh:** the test counts and gameplay checkpoints below are historical validation on `b4bbe723` / PR head `26e4d7c2`. The refreshed head `9f984ae6` needs combined validation with upstream #1595; no new test or performance result is implied by the clean merge. The prior Linux build/test job passed, but its headless/package job stopped in the external pinned MPW download before those checks ran. That infrastructure failure is not counted as a passing headless/package result.

On b4bbe723 with published m68k 0.13.0, the full library suite passed:
**5,364 passed, 0 failed, 3 ignored**. All nine headless-specific desktop
tests passed. Full desktop suite: **97 passed, 1 failed, 1 ignored**.

The desktop failure is the unchanged upstream
`metal_present::tests::minification_retains_thin_strokes_between_sample_centers`.
It first could not access a Metal device under the local sandbox; with
offscreen GPU access it fails the CPU/GPU pixel-tolerance assertion. It also
fails alone in a fresh process, without running the headless tests. The
Metal source, shader, and software display source match upstream blob hashes
exactly. This is reported rather than hidden by skipping the test. No window
was opened. Whole-tree rustfmt also reports the pre-existing formatting
differences already acknowledged as non-blocking in upstream CI; the new
headless module passes its targeted rustfmt check and `git diff --check`.

Test command (Intel macOS, rustc 1.96.1):

```sh
CARGO_INCREMENTAL=0 cargo test --profile ci-test --locked --offline \
  --features test-support --config profile.ci-test.codegen-units=4 \
  --lib --bin systemless -j1 -- --test-threads=2
```

Production-release checkpoints were revalidated on the pre-refresh driver
with the same published dependencies (private phase counters were present
only in the local measurement build). These were **correctness-only runs**
overlapping test compilation; their CPU totals are excluded from acceptance.

| Replay | Frontend phase | Guest phase ticks | Guest phase instructions | Checkpoints inspected |
|---|---:|---:|---:|---|
| EV Override | 3300–6300 | 3000 | 202,572,307 | Actual flight near Earth, NPCs/asteroids, full shields at entry; not a retained name dialog |
| SimCity 2000 | 1800–4800 | 3000 | 151,529,853 | Empty new city advancing January to October 1900, no budget dialog |

Both deliver every scripted input (16 EV, 8 SC2K), remain running, and report
zero same-tick frames and zero exhausted frame budgets. SC2K's phase guest
instruction count and captured mono audio also match the earlier driver;
fonts and other upstream changes mean old pictures are not substituted for
current checkpoints. These checks verify the replay milestones, not visual
perfection or a CPU saving. No windowed CPU/FPS claim is made.

Focused regressions cover CLI mode separation and invalid endpoints, shared
GUI-rate initialization, one-tick progress, retained-modal GUI/headless PC/SP
and clock agreement, one refire per frontend tick, mouse-release latching,
sound callbacks before subsequent mixing, and deferred menu painting. The
last test has guest code repeatedly read a menu-bar pixel across foreground
batches: it must see the original byte throughout guest execution, and the
final presentation must still paint the menu.

The library regression contrasts the same retained wait under legacy and GUI
scheduling. It illustrates removal of synthetic replay work, not a new game
optimization. Font/presentation changes require fresh gameplay checkpoints;
old replay screenshots are not accepted as current-master validation.

## Limits and lessons

Headless does not reproduce the GUI's real wall-clock cutoff/catch-up behavior,
host rendering, drawable waits or device output. Its CPU seconds are not an
Activity Monitor percentage and its virtual ticks are not measured display
FPS. Captured audio checks cover mono mixed samples, not full stereo/device
equivalence. Checkpoints cannot prove every transient frame.

For a native-style optimization, first remove work that an event-driven app
would not perform. A benchmark that forces an already-parked GUI wait to
re-enter obscures that distinction. Correcting this driver makes subsequent
trap/drawing optimizations testable at a representative call frequency; those
patches and their measured effects belong in separate PRs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,33 @@ The terminal reports CPU, compositing, outline rendering and Metal drawable-wait
 phases that take at least 50 ms. During normal gameplay, drawable waits on the
 presentation worker do not block the guest CPU or input handling.
 
+### Headless replays
+
+`systemless --headless --max-ticks 600 game.sit` runs 600 simulated frontend
+ticks (about ten seconds) without opening a window or sleeping in host time.
+It uses the GUI runner's retained-wait/callback scheduling, advances audio,
+uses the same architecture-specific instruction rate as the GUI, and
+composites once per frontend tick. This is also the default headless
+mode when neither legacy instruction option is supplied.
+
+Use `--tick-input-script inputs.txt` with `--max-ticks` to replay inputs;
+each line is `<elapsed-tick> <action> [args]`, e.g. `120 mousedown 317 491`
+and `122 mouseup 317 491` (coordinates are vertical, horizontal). The
+frontend clock continues while menu tracking freezes guest TickCount.
+Reports include both clocks and actual instruction work. Startup Mac time
+is fixed for repeatability; `--headless-start-time SECONDS` overrides it.
+Inputs at or beyond the endpoint are rejected. The summary also reports
+same-tick frames and frames that exhausted the instruction safety budget;
+those counters help detect stalled or unmatched workloads. Use a fresh copy
+of the same save files for each comparison (saves live beside the archive).
+
+`--max-instructions` and `--input-script` retain the old instruction-clock
+diagnostic mode. Retained modal waits can re-fire repeatedly in that mode,
+so its CPU totals are **not a proxy for GUI or gameplay CPU usage**. Tick
+scripts and instruction scripts cannot be combined. Time-based headless
+results still exclude the host window, compositor, and physical audio device;
+compare equal game progress and outputs, and verify windowed CPU separately.
+
 ## Try it in your browser
 
 | [Marathon](https://systemless.org/marathon) | [Escape Velocity](https://systemless.org/escape-velocity) |
--- a/src/bin/desktop/headless_time.rs
+++ b/src/bin/desktop/headless_time.rs
@@ -1,0 +1,429 @@
+//! Simulated frontend time, without a window or wall-clock sleeps.
+//!
+//! An instruction-bounded run is useful for interpreter diagnostics, but a
+//! retained Toolbox wait is not useful guest computation. Drive the existing
+//! GUI scheduling path with one-tick deadlines instead of re-firing that wait
+//! until millions of synthetic A-line instructions have retired.
+
+use super::{
+    configure_realtime_execution_rate, foreground_cpu_batch_instructions, game, save_screenshot,
+    service_pending_sound_work_budgeted, DesktopSaveStore, FixtureRunner, HostMouseReleaseLatch,
+    InputAction, ScriptedInput, UiThemeId, AUDIO_CALLBACK_CHUNK_SAMPLES,
+};
+
+fn deliver(runner: &mut FixtureRunner, mouse: &mut HostMouseReleaseLatch, action: InputAction) {
+    match action {
+        InputAction::MouseMove { v, h } => runner.set_mouse_position(v, h),
+        InputAction::MouseDown { v, h } => {
+            mouse.press();
+            runner.push_mouse_down(v, h);
+        }
+        InputAction::MouseUp { v, h } => {
+            if let Some((v, h)) = mouse.release((v, h)) {
+                runner.push_mouse_up(v, h);
+            }
+        }
+        InputAction::KeyDown { key, ch } => runner.push_key_down(key, ch),
+        InputAction::KeyUp { key, ch } => runner.push_key_up(key, ch),
+    }
+}
+
+#[derive(Debug, Default)]
+pub(super) struct FrameWork {
+    instructions: usize,
+    foreground: usize,
+    budget_exhausted: bool,
+}
+
+fn service_sound(runner: &mut FixtureRunner, total: &mut usize, reserve_used: &mut usize) {
+    if let Some(steps) = service_pending_sound_work_budgeted(
+        runner,
+        game::MAX_INSTRUCTIONS_PER_FRAME,
+        *total,
+        reserve_used,
+    ) {
+        *total = total.saturating_add(steps);
+    }
+}
+
+/// One virtual frontend tick. The frontend clock keeps moving while a menu
+/// freezes guest TickCount, just as the real host clock and audio do. A short
+/// retained-tracking slice is a yield boundary, not an invitation to re-fire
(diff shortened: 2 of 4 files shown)


rlanday:
CI update: the Linux build/test job passed. The headless & package job stopped in the pre-existing reproducible-showcase Docker setup: curl could not connect to the pinned MPW archive on sonixwave.com after 133.8 seconds (exit 28). It never reached the headless library, docs, or package checks, so those are not reported as passing. The job-only retry was not available. No source or CI workaround has been added. Failure log: https://github.com/benletchford/systemless/actions/runs/34142339444/job/101806858255

benletchford:
Rebased onto current master and flattened to one conventional commit, retaining the legacy diagnostic mode and the separate simulated-time mode. New-head CI and focused desktop headless tests are running. I am also testing this together with the sound-finalization and theme-cache candidates; no new CPU-performance claim is implied.

benletchford:
Fresh combined validation with the current palette mirror, sound-finalization, simulated-time headless, and theme-cache changes: 5,437 library tests passed (3 ignored), 98 macOS desktop tests passed (1 ignored), including the Metal minification test, and wasm32-unknown-unknown library check passed. The complete source diff is whitespace-clean. Waiting for the refreshed PR CI showcase checks before merging; no new performance percentage is claimed.

PR #1649, opened 2026-09-10
fix: preserve explicit screen CopyBits output

### LINKED ISSUE #1644: Crystal Quest loses static frame and menu pixels after monitor-mode launch
## Summary

Crystal Quest Sample loses static artwork and text layers after its documented Option-launch monitor selection. The game remains active and dynamic sprites continue to render, but the current frame no longer preserves the complete native artwork.

## Reproduction

1. Launch the public Crystal Quest Sample archive: https://assets.systemless.org/games/crystal-quest.sit
2. Hold Option while launching, release it when the monitor options appear, and choose the documented temporary monitor selection.
3. Press Return to reach the high-score menu, then click to start the arena.
4. Capture the native-resolution frame after each transition.

Archive SHA-256: `88cf365c6985786e383acec7582ea6467b448f6f12c43cb9f46cd1fd1a552f0e`.

## Expected

The 800x600 RGB frame preserves the complete nested colored frame, score/menu text, and static HUD layers, matching the BasiliskII oracle for the same script and archive.

## Actual

Systemless captures remain 800x600 RGB and show the blue outer frame and moving sprites, but the nested colored frame and substantial static score/menu text layers are absent or replaced by flat palette-colored rectangles. After mouse movement, the HUD is also missing while dynamic objects remain visible.

Current capture hashes:

- `launch.png`: `6b3d6b6638dc26195dd945bb1c65b44ab81a1859b91d535ad50202e5736bb367`
- `menu.png`: `a89872bb35bcd1ab1cec66f47b6d1545ca85eec6185253707b0923d38215effd`
- `gameplay.png`: `2d3727e40a5469051eff5d6eb959ee320b2e15d44abae80a1e30dd696a76ef59`
- `after_mouse_move.png`: `09c46ccda30039b2d6eedd44179581c662d68f46ce8d16be5288fada8a7b9202`

Please diagnose and fix this generically in the rendering/framebuffer path; do not add Crystal Quest-specific runtime logic. Related dirty-redraw investigation: #613.

benletchford:
Closes #1644

## Root cause

The generic redraw bridge could blit an offscreen window backing port after an application had explicitly copied pixels into the physical screen. That second blit erased screen-owned artwork that was not present in the backing port, producing the missing static frame and menu-layer pixels reported in Crystal Quest.

## Fix

Treat explicit screen-targeted CopyBits output as authoritative for the physical framebuffer and suppress only the later fallback window blit. Ordinary offscreen windows still use the bridge before explicit screen output. The change is generic compositor behavior and contains no game-specific runtime logic.

## Proof and validation

- Added focused framebuffer regression coverage for both pre-screen fallback blitting and post-screen preservation.
- Focused framebuffer/compositor test suite: 77 passed.
- git diff --check clean.
- Commit: 7077ebfb15a5cf8de00fad70eaa9e72023939128.

### DIFF
--- a/src/trap/dispatch.rs
+++ b/src/trap/dispatch.rs
@@ -1819,8 +1819,19 @@ pub struct TrapDispatcher {
     /// ~20-30ns measurement overhead per trap when enabled. Dump via
     /// `print_trap_timing_histogram`.
     pub trap_time_ns: Box<[u64; 4096]>,
-    /// Number of copybits_screen events emitted (screen-affecting draws).
+    /// Cumulative number of effective CopyBits draws into the screen.
+    ///
+    /// This lifetime is intentional: desktop content-rect detection compares
+    /// successive totals. Compositor suppression uses
+    /// `last_screen_copybits_tick` instead of treating this counter as a
+    /// current-frame flag.
     pub copybits_screen_count: u64,
+    /// Guest tick in which the most recent effective screen CopyBits occurred.
+    ///
+    /// The window-backing bridge must defer only for the frame that contains
+    /// the explicit screen draw; the cumulative event count cannot express
+    /// that lifetime.
+    pub(crate) last_screen_copybits_tick: Option<u32>,
     /// Most recent sizeable CopyBits blit into the screen framebuffer.
     pub last_screen_copybits_rect: Option<ScreenCopyBitsRect>,
     /// Largest non-fullscreen FrameRect drawn into the screen framebuffer in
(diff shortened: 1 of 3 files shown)


benletchford:
Review of the current head: this needs a compositor-lifetime fix, not just a rebase. `screen_copybits_in_current_tick` treats guest TickCount as a presentation/ownership boundary. In the new regression, advancing to tick 101 deliberately copies the stale window backing over all previously preserved screen pixels. Static artwork drawn once will therefore still disappear on a later tick without a replacement screen CopyBits; conversely, a small screen CopyBits suppresses the entire window bridge for that tick. Please track the actual screen/backing ownership or dirty regions and add regressions for static artwork surviving later ticks, partial-screen CopyBits alongside independent backing updates, and a tick advancing before composition. Keep #1644 open until the actual Crystal Quest transition and mouse-movement checkpoints pass.

PR #1664, opened 2026-09-11
fix: animate palettes shared with the active window

### LINKED ISSUE #1662: Palette animation ignores shared active palettes through another window
## Observed behavior

Museum Madness reaches gameplay after correcting intrasegment relocation, but renders in grayscale. The game associates one animated palette with both its front window and another window, then calls AnimatePalette through the latter association. Systemless rejects the device update solely because the supplied window is not the front window, even though both refer to the active palette.

## Expected behavior

Animation of entries belonging to the active palette should update their device colors regardless of which window sharing that palette is supplied. Unrelated background palettes must not overwrite the active palette's colors.

Inside Macintosh Volume V, V-164 describes AnimateEntry/AnimatePalette in terms of the device indexes reserved by the palette.

## Reproduction

Run Museum Madness after resolving #1661. `SYSTEMLESS_TRACE_PALETTE=1` shows palette `0x0021FF44` activated for front window `0x0021FFA0`, then animated through window `0x003C9342`. The device remains on the grayscale initialization ramp. BasiliskII displays the same archive in color.

benletchford:
Closes #1662.

Permit palette animation and activation through another window when its exact palette association shares the front window's active palette. Keep unrelated background palettes isolated, and synchronize the guest color tables through the existing activation path.

Validation:
- Added a regression for both shared and unrelated background palettes.
- Full library suite: 5,414 passed, 3 ignored.
- Museum Madness now renders the main hall in color and responds to keyboard and mouse input; compared against a BasiliskII capture of the same archive.

Depends on #1663 for Museum Madness's startup correction.

### DIFF
--- a/src/trap/quickdraw.rs
+++ b/src/trap/quickdraw.rs
@@ -17490,13 +17490,13 @@ impl super::TrapDispatcher {
         // pmAnimated entries immediately. A pmExplicit entry still observes
         // its existing device cell during ordinary palette activation, so the
         // animation traps must update the changed cell before reactivation.
-        // Restrict the write to the front window's exact association: a
-        // background window may update its palette without changing the live
-        // screen until that window becomes active.
+        // Windows can share a palette. Animation through either association
+        // changes the same active device cells, while an unrelated background
+        // palette must not change the front window's color environment.
         // Inside Macintosh Volume V (1986), V-164; Volume VI (1991), 20-13.
         if usage & PM_ANIMATED == 0
             || window == 0
-            || window != self.front_window
+            || self.window_palette_handle(self.front_window) != palette_handle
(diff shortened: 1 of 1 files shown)


benletchford:
Additional actual-game verification after the fixes: Museum Madness opens its backpack and exhibit map, enters the Robotics exhibit, moves the player, and opens the moon-walker repair puzzle. Saving in the lobby creates a 550-byte SAVE/MMus file; loading it from the exhibit restores the lobby. Required CI checks and the local wasm32 check pass. This is a gameplay smoke test, not a complete playthrough of all 25 exhibits; interactive desktop file-picker, audio, and browser gameplay verification remain outstanding.

Issue #1677, opened 2026-09-11
Dialog background is not cleared when a dialog is brought to the front


wraitii:
When a dialog is created in the background and then brought to the front, its background is not cleared (to white, usually). This looks buggy in [Factory 1.3](https://classicreload.com/mac-factory-the-industrial-devolution.html) (at least).

PR #1678, opened 2026-09-11
fix(trap): redraw dialog background when brought to front

wraitii:
## Summary

See #1677, in Factory 1.3 the "about" dialog and the calendar dialog didn't properly clear the background as they're created in the back and then brought in focus.

(Contributing says to create an issue as well, so I did).

PR made by deepseek-v4.1-flash but it seems reasonable.

## Validation

Tested manually.


benletchford:
Hi @wraitii , happy to merge, plz review CLA.

PR #1680, opened 2026-09-11
feat(debugger): add opt-in inspection, execution control, and captures

wraitii:
With my apologies for the rather large PR. I think this is the minimum size where this makes some sense. FWIW, I've done several 5.6-sol & astra review iterations.

## Summary

This PR implements the first step toward an extensible debugger. It adds an opt-in API for inspecting guest execution and subsystem state, controlling 68K execution, and retaining captures.

The implementation includes:

- 68K register and RAM inspection, approximate disassembly, stack previews, stepping, and program-counter breakpoints.
- PowerPC register inspection with explicit unsupported-operation responses.
- Graphics, window, and event snapshots, including retained framebuffer artifacts.
- An optional Unix socket transport for GUI and both headless modes.
- Headless startup pause, frame-boundary captures, and inspection after terminal guest halts.

Live access stays on the runner thread. Read-only architecture adapters and subsystem providers provide extension points, while debugger state is disabled by default. This establishes the foundation for future tooling; a graphical inspector and broader execution control remain follow-up work.

Validation includes the debugger-enabled library suite, targeted pause and capture regressions, live headless socket checks, and desktop tests with debugging enabled and disabled.

**Architecture**

The debugger is an optional layer around `FixtureRunner`, which remains the owner of guest execution and emulator state. Live debugger requests run on that same thread, between execution slices, avoiding additional locking or concurrent access to guest state.

- **Shared request API.** Typed requests, replies, and errors serve both in-process tools and the Unix socket transport. The socket worker handles I/O; the runner applies requests at safe points.
- **Central coordination.** A runner-local coordinator tracks execution state, breakpoints, pending operations, captures, and notifications. Session and generation identities let clients detect stale references.
- **Architecture adapters.** Read-only adapters expose contexts, registers, address spaces, and inspection capabilities. Execution control stays in scheduler hooks, so additional architectures can advertise control support as their execution loops implement it.
- **Subsystem providers.** Graphics, windows, and events expose owned snapshots through a common provider interface. New subsystems can register providers without adding subsystem-specific logic to the request handler.
- **Explicit capture boundaries.** Captures copy state immediately or at a supported frame boundary. Retained snapshots and artifacts remain readable after execution resumes; future-boundary requests return operation IDs for tracking completion.
- **Opt-in integration.** The `debug` feature enables the foundation; `debug-server` adds the optional transport. Default builds retain inert hooks without carrying debugger state.

This leaves room for a graphical inspector, richer PowerPC support, and additional providers while keeping the current implementation bounded. Pausing takes effect at scheduler safe points and cannot interrupt synchronous host work already in progress.

## Validation

My agents have been able to use it to inspect some things.

## Known limitations:
- The disassembler is a bit broken and should be improved independently.
- no GUI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,12 +63,20 @@ jobs:
       - name: Build all targets
         run: cargo build --all-targets --all-features
 
-      # test-support enables the scripted_traces module and its tests.
+      # test-support enables the scripted_traces module and its tests. The
+      # debugger is an opt-in feature, so this run also covers its disabled
+      # path; the next step exercises the enabled one.
       - name: Run library tests
         run: cargo test --lib --features test-support
 
+      - name: Run debugger library tests
+        run: cargo test --lib --features test-support,debug
+
       - name: Run desktop tests
         run: cargo test --locked --bin systemless
+
+      - name: Run debugger desktop tests
+        run: cargo test --locked --bin systemless --features debug-server
 
       # Both Systemless runtime configurations use this exact optimized executable.
       # Keep ordinary library/desktop tests in the checked debug profile above.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,6 +70,11 @@ jit = ["m68k/jit"]
 # Exposes `scripted_traces`, deterministic trap-replay test scaffolding.
 # Off by default so it stays out of the published public API and docs.
 test-support = []
+# Opt-in debugger and inspection API. Off by default so ordinary builds do
+# not compile the control/inspection foundation or carry its runtime state.
+debug = []
+# Debugger socket transport (Unix-domain sockets). Implies `debug`.
+debug-server = ["debug"]
 
 [target.'cfg(target_os = "macos")'.dependencies]
 objc2 = { version = "0.5.2", optional = true }
--- a/src/bin/desktop/debug_server.rs
+++ b/src/bin/desktop/debug_server.rs
@@ -1,0 +1,277 @@
+//! Unix JSON-line transport for [`systemless::debug`]. The library module
+//! documents envelopes and client usage; request types document RPC parameters.
+//! This worker owns socket I/O only. The frontend must pump requests between
+//! runner calls, including while paused or halted.
+
+use std::io::{BufRead, BufReader, Write};
+use std::os::unix::fs::FileTypeExt;
+use std::os::unix::net::{UnixListener, UnixStream};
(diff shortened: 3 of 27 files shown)


benletchford:
Hi @wraitii , awesome contribution. Review CLA and will get this in, cheers.

wraitii:
@benletchford Done. Also pushed a few changed after letting claude use it in practice, I think it tested most currently implemented endpoint, things read correct, I added a basic symbolication pass where data is present on 68k.

FYI:
- currently limited to a single connected client on the socket
- can only read, not write
- PPC support is much more limited.

benletchford (changes requested):
The opt-in debugger is a sensible contribution, and keeping live requests on the runner thread is a good foundation. I am holding the merge for one reproducible crash in the current head (`0369c8b5a702a9da978c2d90ccba32cb55d001b5`).

**[P2] Bounds-check the symbol candidate at the end of a lookup window** — `src/debug/symbols.rs:73`

`resolve_containing` calls `symbol_at(data, offset + 2)` after finding a return instruction. When that instruction occupies the final two bytes of the window, the candidate offset equals `data.len()`. `symbol_at` then indexes `data[offset]` and panics. Because socket requests execute on the runner thread, an ordinary inspection request can terminate the emulator.

I reproduced this through the public request handler with a fresh `FixtureRunner`, writing `0x4e75` (RTS) at `0x20000`, then issuing:

```rust
handle_debug_request(&mut runner, DebugRequest::ResolveSymbol {
    context: ContextSelector::Active,
    address: DebugAddress::new(M68K_SPACE, 0x20000),
    window: Some(2),
});
```

The regression test fails with `src/debug/symbols.rs:73:30: index out of bounds: the len is 2 but the index is 2`. Please make the candidate read bounds-safe and add coverage for a window ending at a return instruction, including a read truncated at the end of mapped RAM. This should return an unnamed/not-found result rather than panic.

The public CI run for this head is also currently `action_required`, so there is no passing CI result yet. Please get the fix and the enabled/disabled debugger checks green before merging.

benletchford (approved):
Rechecked f7ca79710f5173a982d790a227bf55c9b454daba. The symbol lookup bounds check resolves the reported crash, and the added tests cover both a window ending at RTS and a read truncated at mapped RAM. My original public-API regression test now passes locally. CI passes the enabled/disabled debugger library and desktop suites, all-target build, and 68K/PowerPC showcase checks. The blocking finding is resolved; approving for squash merge.

Issue #1788, opened 2026-09-14
Investigate sharper small-text presentation while preserving classic font metrics


rlanday:
Investigate improving the sharpness of small rendered text, separately from the Geneva 9 spacing defect in #1787. The user requested this during native Windows SimCity 2000 testing.

“Power Plant Needed” in the upper-left status area provides a concrete example:

![Small status text in native Windows capture](https://github.com/user-attachments/assets/6b2a42e6-34e1-4a5c-a9ed-03ee832d89c5)

Findings from native Windows diagnostics using the actual QuickDraw and presentation code:

- The guest font raster uses monochrome-hinted, binary glyph masks. The host presentation separately renders bundled outlines with normal smooth hinting into a retained 4× surface, then averages coverage for a 1× output.
- A standalone `TextFont(3)`, `TextSize(9)`, `DrawString("Power Plant Needed")` reproduces the soft thin stems. It was tested with `srcCopy` and `srcOr`, raw guest output, and the supported 2×/4× retained surfaces at 1×/2×/4× output.
- On white, the raw lowercase `l` has a black one-pixel stem. The current 4×→1× presentation spreads it into adjacent columns with intensities 155 and 161 out of 255. In the game screenshot, the corresponding columns are 125 and 130 against a 205-gray background. This explains the low-contrast appearance; a new Windows capture is not needed to establish it.
- Reducing the retained surface from 4× to 2× did not fix that example: both produced the same 155/161 stem at 1×. At 2× output, the central samples differed only by one intensity level. Do not claim that lowering the oversampling factor alone improves sharpness.
- At this capture's 800×601 client size, the 800×600 guest frame fits at 1×. The Windows submission path can copy the 1× output without another scaling filter. The observed softness is reproducible in shared presentation code before Windows compositing; this is not evidence of a ClearType or host-font dependency.
- The existing retained-presentation API accepts 2×–4× surfaces, not 1×. Direct output-size hinting at 1× is a proposed experiment, not an implemented or verified improvement.

Suggested investigation:

1. Compare a crisp mode using the already available guest bitmap at 1× with output-size-aware hinted outline rendering. Keep the logical advances, bearings contract, hit testing, wrapping, and guest framebuffer semantics stable.
2. Evaluate thin-stem contrast, curves/diagonals, bold/italic/underline, and readability at small sizes across integral and fractional Windows scaling. A dark-pixel count alone is not a quality measure.
[...]

Issue #1787, opened 2026-09-14
Geneva 9 fallback produces conspicuous gaps in SimCity 2000 status text


rlanday:
SimCity 2000's small status text visibly reads like “Pow er Pl ant Needed” on the native Windows build. The user identified extra gaps after the `w` in “Power” and after the `l` in “Plant”.

The existing native GUI capture shows the affected status text at the upper left (the save-error dialog is a separate bug, #1785):

![SimCity 2000 status text and independent save error](https://github.com/user-attachments/assets/6b2a42e6-34e1-4a5c-a9ed-03ee832d89c5)

Reproduction: launch SimCity 2000 1.2, create a city without a power plant, and examine “Power Plant Needed” in the toolbar status area. Also reproduced independently by drawing that exact ASCII string through `InitGraf`, `TextFont(3)`, `TextSize(9)`, `MoveTo`, and `DrawString` in a native Windows diagnostic linked to the current Systemless library. Both `srcCopy` and `srcOr` exhibit the spacing.

Investigation:

- The fallback for Geneva/Application 9 uses Nimbus Sans Regular outlines and bearings, but substitutes the separately licensed classic-compatible ASCII advances.
- For `w`, the compatible advance is 8 logical pixels versus the underlying hinted font's 7; for `l`, it is 3 versus 2. The isolated reproducer therefore adds one logical pixel at each reported boundary while retaining the replacement glyph artwork. Helvetica 9, routed to the same outline without the Geneva compatibility table, supplies a useful diagnostic comparison, not a proposed replacement font.
- The mixing of outlines and compatibility advances was introduced by commit `7ef19f3ed6c3d2e8833d135751c6b763bf567821` (“fix(fonts): restore classic Geneva 9 advances”, 2026-09-09). Existing Apeiron layout tests require these classic advances. Do not simply remove the table or add ad hoc kerning for this one phrase.
- The screenshot also has faint, antialiased small-letter stems; sharpness is a separate investigation from the spacing/shape mismatch. The larger outline rendering makes the extra space readily visible too.
- This is reproduced in the shared font/rendering code; it has not been established as Windows-specific. The exact application draw-call state has not yet been traced, so the standalone reproduction does not establish whether the request used font ID 1 or 3 (both share this fallback).

[...]

Issue #1753, opened 2026-09-13
SC2K text fields still insert blanks for ordinary F2 and End; volume and caret fixes merged


rlanday:
On a native Windows host, pressing the laptop’s F2 key to lower sound volume while SimCity 2000’s first-run personalization form has focus inserts apparently blank characters and leaves short black marks along the bottom of the text field. The user reported this during interactive testing; a client-area capture confirms the marks in both Owner Name and Company or Organization fields.

Reproduction:

1. Launch `SimCity_2000_1.2.hqx` and reach the first-run form headed “To receive updates and customer service…”.
2. Focus Owner Name or Company or Organization.
3. Press the physical F2 key used for volume down several times.
4. Observe the insertion point advance and black tick marks accumulate at the lower edge of the field.

Expected: a host volume action should not insert text into the guest. If Windows delivers an ordinary F2 function-key event, preserve its Macintosh function-key semantics without manufacturing printable text. Trace the actual Windows event before deciding whether the defect is in host key translation, guest event encoding, or text-field handling/rendering. Fn/Fn-lock state and whether this keyboard reports both volume and F2 events have not yet been captured.

Environment: Windows 11 x64, native `x86_64-pc-windows-gnu` release binary built in WSL with Rust 1.98.1. Base commit `61d01c61ef774df9286a3849bb6a1bc6094858fc` plus uncommitted Windows cursor, screenshot-path, VISE external-file, QuickDraw color-state, and dialog/title fixes. PR #1598’s hashing changes are excluded. Tested executable SHA-256: `a5627ef922648dfea768c878c944bcf088614f09ac8198fe8ef900a55d7815cc`.

The archive’s SHA-256 is `55c912b0979a6363c2923ae9fdcb8fec0360f9c36adea1f461ac50ecab7aa334`.

Initial code observation, not yet a confirmed root cause: `keycode_to_mac` maps F2 to Macintosh virtual key `0x78`, while `keycode_to_mac_char` falls through to character byte zero for F2. Investigate the resulting EventRecord and the application’s field renderer; do not simply drop all function keys, since guest applications may use them.

Behavioral reference: [Inside Macintosh: Macintosh Toolbox Essentials, Event Manager — Handling Low-Level Events](https://dev.os9.ca/techpubs/mac/Toolbox/Toolbox-40.html) distinguishes virtual key codes, translated character codes, and application handling of special keys. Verify the appropriate keyboard translation and text-editing contracts against Inside Macintosh while fixing this.

[...]

Issue #1719, opened 2026-09-13
SimCity 2000 installer draws a corrupt disk icon from copied low-memory bytes


rlanday:
The 32×32 disk icon in Install Location is vertical stripes and fragments instead of a recognizable disk.

Tested native Windows x64 (`x86_64-pc-windows-gnu`, Rust 1.98.1), cross-built in WSL, on Windows 11. Source: `61d01c61ef774df9286a3849bb6a1bc6094858fc`. The reproduction executable differs from that commit only in changing screenshot output from `/tmp` to the platform temporary directory; it does not contain the cursor/hash or graphics fixes under development.

Input: `SimCity_2000_1.2.hqx`, SHA-256 `55c912b0979a6363c2923ae9fdcb8fec0360f9c36adea1f461ac50ecab7aa334`. It contains `SimCity 2000® Disk 1` and `Disk 2`, with a split Installer VISE payload.

Launch a fresh copy with:
```powershell
.\systemless.exe .\SimCity_2000_1.2.hqx --ui-theme classic-system7 --display-scale 1
```

Press Enter twice to reach the main installer (the screen with Read Me, Select Folder, Switch Disk, Quit, and Install). Inspect the icon beside Select Folder and Switch Disk. It remains corrupt after opening and closing Read Me. The corruption is present in the guest framebuffer, before Windows presentation.

At the drawing checkpoint, CopyBits reads two 128-byte 1-bit planes from `0x0020F040` (srcBic mask) and `0x0020EFC0` (srcOr image), targeting local rect `(top=214,left=82,bottom=246,right=114)`. The complete 256-byte source at `0x0020EFC0` is byte-identical to guest memory `0x00000000..0x00000100`, including low-memory/vector values. It is not icon data. The archive's `ICN#` 1000 resource itself contains an intact disk bitmap and mask; no assumption is made yet that this is the resource the driver path should select.

Trace the first invalid pointer/result or copy that populates these buffers, including the disk-driver icon query and fallback resource path. Correct that producer rather than changing CopyBits to hide bad input. Add focused coverage for the responsible API and verify the icon survives Read Me redraw.

Screenshot evidence captured:

<img width="800" height="600" alt="Image" src="https://github.com/user-attachments/assets/d100799e-a54e-4f90-be12-7533956c1f2d" />

<img width="800" height="600" alt="Image" src="https://github.com/user-attachments/assets/c9acb7ca-90a1-47b2-b997-d6d385c3c022" />

[...]

PR #1800, opened 2026-09-14
fix: preserve smooth SimCity 2000 button text through CPU recoloring

### LINKED ISSUE #1791: SimCity 2000 New City button labels permanently lose high-resolution text after selection
In SimCity 2000's New City dialog, button labels initially use the high-resolution text presentation, but clicking a button changes its label to a visibly lower-resolution rendering. The degraded label persists after selecting a different button. The user reports this for both groups:

- Starting years: 1900, 1950, 2000, 2050.
- Difficulty: Easy, Medium, Hard.

Reproduction:

1. Launch SimCity 2000 afresh and start a new city.
2. Before interacting, observe the text on the year and difficulty buttons.
3. Click a year button, then another year button.
4. Repeat with the difficulty buttons.
5. Compare previously clicked labels with untouched labels. Previously clicked labels remain at lower resolution after selection moves away.

Expected: selecting or deselecting a button should preserve the label's text rendering quality. The selected appearance may change, but the label should not permanently fall back to low-resolution text.

Reported during native Windows GUI testing with the SimCity_2000_1.2.hqx archive (SHA-256 55c912b0979a6363c2923ae9fdcb8fec0360f9c36adea1f461ac50ecab7aa334). The current native test executable, sc2k-save-key-fixed.exe, has SHA-256 419c496bcceee8c5605df27d1e9eec315b618fb1dc7b2ac7f8237c91aab5c1d0 and includes unmerged compatibility fixes on base 61d01c61ef774df9286a3849bb6a1bc6094858fc. Exact first-affected revision is not established.

Fresh-launch before/after capture and call-path tracing are in progress. Investigate whether selection redraw invalidates or replaces the retained high-resolution label without recreating it; this is a hypothesis, not a confirmed cause. Preserve guest-visible selection behavior and reference Inside Macintosh's Control Manager/QuickDraw contracts when fixing it.

Related: #1788 tracks the separate investigation into general small-text sharpness. This issue is specifically the persistent quality loss caused by interacting with a control, not general font softness or the spacing issue in #1787.

rlanday:
## Summary

SimCity 2000 shades the New City year and difficulty buttons by reading and rewriting their framebuffer pixels. It changes gray palette indexes and writes the unchanged black letter pixels back too. Systemless treated every store as a bitmap replacement, discarding the retained outline coverage. **Easy looked particularly bad even on the initial dialog**, because the game shades the default selection before the first display. Clicking Medium or a year permanently made those labels rough as well.

Fixes #1791.

Systemless's high-resolution text presentation already maintains two representations of text. The ordinary QuickDraw path handles the guest's DrawString/DrawText calls using the port's font, size, style, pen position, transfer mode, and clipping. It writes the normal guest-resolution raster and advances the pen using the existing text metrics. Where outline artwork is available, a companion presentation path also rasterizes the glyph at higher resolution and retains its coverage for host display. The normal retained surface has four samples per axis—16 samples per guest pixel—and the frontend resolves those samples to the actual window resolution. This can produce smoother edges even in an 800×600 window; it does not turn the guest desktop into a larger desktop.

| Property | What the guest game sees | Extra data used for host display |
| --- | --- | --- |
| Coordinates and layout | Existing guest coordinates, glyph advances, pen positions, clipping, and hit testing | Outline samples positioned using that same layout |
| Pixel memory in this SC2K mode | One palette index per pixel in the 8-bit framebuffer | Coverage, foreground/background indexes, and higher-resolution samples |
| Pixel reads and calculations | Actual guest bytes, including exact palette-index comparisons | No antialiased RGB values are substituted into guest reads |
| Text copied or overwritten | Normal bitmap operations and memory stores | Supported copies carry retained detail; ordinary overwrites invalidate it |
| Final display | The game does not consume this output | Retained text and ordinary game pixels combined at the host window resolution |

```mermaid
flowchart LR
    Q[Guest QuickDraw text call] --> G[Guest raster and text metrics]
    Q --> H[Retained outline coverage]
    G --> R[Game reads and rewrites pixels]
    R --> G
    G --> P[Host presentation]
    H --> P
    R -. Qualifying recolor updates coverage .-> H
```

This separation matters for compatibility. A classic application can measure a string, center it in a button, draw it, save the resulting pixels, and later manipulate those pixels without making another text call. Its layout and pixel algorithms must continue to receive the guest representation they were written for. QuickDraw's text routines cover both drawing and measurement, including caret positioning; [Inside Macintosh: Text, QuickDraw Text](https://dev.os9.ca/techpubs/mac/Text/Text-126.html) describes that relationship. The Font Manager also defaults to integer widths for compatibility with older applications, as explained in [How the Font Manager Calculates Glyph Widths](https://dev.os9.ca/techpubs/mac/Text/Text-197.html). Higher-resolution host artwork therefore uses the existing layout rather than silently replacing glyph advances with a modern font's preferred spacing.

For SC2K's buttons, the observed original guest behavior is a raster operation: after drawing the label, the game shades its rectangle by comparing and changing palette indexes in memory. The game has no knowledge of the retained outline and no reason to redraw the text after shading. That operation is valid for its bitmap representation. The regression was in the host presentation's response to those stores: it discarded supplemental detail, so the label visibly switched to the underlying bitmap and stayed there. Game drawing did not suddenly choose a different font or text size.

Retained detail already follows supported QuickDraw transfers, saved pixel snapshots, and CPU memory-to-memory copies. It also retains palette indexes so later CLUT changes recolor existing antialiased edges. Ordinary stores deliberately remove the detail: keeping an old outline after the game paints over it would leave stale letters or edges. The new fix extends preservation to a restricted class of direct CPU recoloring, which differs from a copy because its source and destination are the same pixels and the indexes may change.

The raw-frame comparisons below establish that this change preserves Systemless's existing guest rendering for the reproduced sequence. They do not establish pixel-perfect agreement with every original Macintosh font. Existing font substitutions, metric/artwork spacing (#1787), and general text sharpness (#1788) remain separate concerns. Bitmap-only artwork and unsupported outline cases continue to use the existing rendering path.

The fix preserves the existing outline coverage through qualifying indexed recoloring. The 68K CPU wrapper marks guest execution, and the presentation layer records old-to-new palette-index mappings and affected text coverage until a trap boundary. An instruction-budget pause keeps the observation pending. Before native drawing overwrites the screen, it finishes the pending observation so later drawing still wins. When the writes establish a consistent, non-identity, one-to-one index map, the presentation layer applies that map to the retained samples and to both foreground and background indexes used for antialiasing. Guest memory is never rewritten by this inference. There are no game names, button labels, palette ranges, or guest instruction addresses in the detection logic.

For example, the game's observed shading loop maps gray index 142 to 150, maps 150 to 142, and leaves black index 255 unchanged. A guest row containing `[142, 255, 142, 150]` therefore becomes `[150, 255, 150, 142]`. Previously the second store removed the smooth letter edge even though the guest byte remained 255. The ordinary framebuffer still receives all four stores with this fix.

A retained antialiased sample is more than the guest byte: it carries coverage and indexed foreground/background information. If a sample contains black index 255 at coverage `a` over background index 142, its displayed color is a blend of `CLUT[255]` and `CLUT[142]`. Recoloring changes the latter to `CLUT[150]` and keeps `a` unchanged. The same transformation applies to fully covered samples and to nested background blends from overlapping glyphs. Transforming only the logical byte, or retaining only the old blended RGB color, would not update these edges correctly.

The observation is separate from the actual store path. Guest stores take effect immediately and invalidate coverage as before; the candidate holds references to affected coverage that existed before those stores. At a qualifying boundary, it restores only transformed presentation metadata. It does not suppress stores, rewrite guest memory, change QuickDraw metrics, rerun DrawString, or redraw the button on the game's behalf. Simply ignoring every same-value store would be incorrect: an application can intentionally paint over an existing outline with its underlying bitmap value, including in an erase operation.

[Inside Macintosh: Imaging With QuickDraw, Color QuickDraw](https://dev.os9.ca/techpubs/mac/QuickDraw/QuickDraw-198.html) defines indexed pixels as CLUT indexes. The guest’s index writes remain authoritative. Preserving antialiasing through an inferred remap is a Systemless presentation enhancement; it is not an additional QuickDraw behavior specified by Inside Macintosh.

Safeguards and limits:

- Restrict inference to forward writes to an 8-bit screen by the 68K CPU. Other formats, offscreen recoloring, and PPC execution keep their existing behavior.
- Reject conflicting mappings, multiple source indexes collapsing into one destination index, and repeated or backward stores. Ordinary fills and same-value rewrites retain normal coverage invalidation.
- Restore a text cell only when every index in its retained foreground/background samples has an observed mapping. Do not guess unseen colors.
- Retain at most 4,096 affected cells per observation; exceeding the limit discards the candidate and keeps the guest bitmap rendering.
- Keep CPU memory-copy preservation separate, and let later native drawing overwrite restored coverage normally.

The mapping checks apply to palette indexes, including indexes that currently resolve to equal RGB colors. Each observed source index must always produce the same destination, and each destination must have only one observed source. A fill that maps both letter and background indexes to the same value consequently fails validation. A cell whose antialiased background contains an index never seen in the pass also fails; an unknown entry is not assumed to mean “unchanged.” A wholly identical rewrite has no non-identity mapping and is rejected.

CPU slicing must not change the result. Ending a batch because its instruction budget expired suspends observation rather than treating the partially executed drawing loop as complete. Single-instruction stepping follows the same rule. Trap and terminal CPU exits finish observation before host trap handling. Native screen writes or coverage restoration between slices finish any pending observation first, then perform their own drawing in the normal order. A memory-to-memory CPU copy already has its own source-coverage transfer; that path invalidates recoloring inference instead of letting destination history override the copied source.

Regression risk is not zero. The map is inferred from observed stores, not from the application's intent. Unrelated writes between traps could happen to satisfy the same conditions, preserving or recoloring old antialiased edges where the application intended a replacement. This would affect host presentation, not guest bytes or text metrics. A long direct-drawing loop can also display temporarily flattened pixels before it reaches the boundary where coverage is restored. Restrictive detection can miss legitimate recoloring, leaving the existing bitmap appearance. Finally, observing direct framebuffer stores and retaining metadata adds work; Windows profiling is still needed to quantify that overhead. Other games and macOS have not yet been tested with this change.

One concrete false-positive case is a guest that changes a color in one screen region and then intentionally redraws identical bitmap text elsewhere, all in increasing address order before the next trap. If the combined writes establish a consistent mapping for all the text's retained colors, they can look like one recoloring pass. The enhancement can then retain old outline edges even though the caller intended bitmap replacement. The current checks reduce this possibility but do not establish semantic provenance for each CPU store. The implementation does not claim to recognize arbitrary guest drawing perfectly.

There are also deliberate false negatives: reverse-order loops, overlapping passes, transformations interrupted by native drawing, unsupported pixel formats, and transformations requiring unseen colors may still lose their outline detail. The 4,096-cell cap limits temporary retained coverage, not total presentation memory in bytes; cells have variable metadata sizes. Large transformations fall back to the guest bitmap instead of growing the candidate without limit. The index and reverse-index tables contain 256 entries each, and screen stores pay the observation checks while a candidate is active. The frame images and unit tests establish rendering correctness for the covered cases; they do not establish that the added work is free or that every application is unaffected.

## Validation

This PR contains only the CPU/presentation changes for #1791. Commit `df1a93991d7d89a687a6b633adf50ec30877885b` was tested independently on upstream master `888f72a261259d618f5331d9bf7663ea203b80e6`.

- Native Windows tests: 5,446 library tests passed, 3 ignored; all 59 desktop tests passed.
- The isolated native Windows desktop executable built successfully with the optimized `ci-test` profile; whitespace checks passed.
- A native Windows replay used extracted game files from the supplied SC2K archive, isolating this change from the separate installer fix. All six raw guest-frame PNGs are byte-for-byte identical to the baseline before the button fix. All six presented PNGs exactly match the earlier validated fixed replay.
- Regression tests execute real 68K instructions under single-step execution and batch budgets of 1, 7, and 1,000 instructions. They check erases, conflicting maps, missing colors, repeated stores, and native drawing between CPU slices.

The CPU regression leaves black pixels unchanged while recoloring the surrounding background and compares the retained indexed samples with the expected transformation. The negative cases verify that fills, inconsistent mappings, missing background evidence, overlapping stores, and a subsequent native overwrite do not resurrect old coverage. These checks do not replace broader game/host testing or a performance comparison under direct drawing workloads.

## Screenshots

These native Windows GUI captures were taken with the earlier combined build, before this fix was split into its own PR. The independent replay described above reproduces all six displayed frames from the earlier fixed replay that accompanied those GUI checks. In the GUI checks, deselected Medium and 1950 returned to exactly their initial smooth pixels; untouched 2050 also remained identical.

**Before fix — initial New City dialog, Easy and 1900 selected.** Easy already looks especially rough because shading the default selection discards its retained text detail.

1-before-selection

**After fix — initial New City dialog, Easy and 1900 selected.** The default shading preserves Easy’s smooth text.

4-fixed-before-selection

**Before fix — after selecting 1950/Medium, then 2000/Hard.** Previously selected Medium and 1950 remain rough after deselection; 2000 and Hard are now selected.

3-2000-hard-selected

**After fix — the same selection sequence, ending on 2000/Hard.** Medium and 1950 retain their smooth text after deselection, and the selected labels keep their detail.

6-fixed-2000-hard-selected


rlanday:
Ready for review; the remaining license-job failure is a tooling-download failure. The build/tests, headless/package, and formatting/clippy jobs passed. The [license job](https://github.com/benletchford/systemless/actions/runs/34805285198/job/103855870636) failed while downloading/extracting cargo-deny (`gzip: invalid magic`), before any license checks ran. No dependencies or license configuration changed in this PR. A maintainer needs to retry that job; the author account cannot rerun upstream jobs (GitHub returned 403).

rlanday:
The focused fixes from the original combined PR are now split as follows:

- #1800: retained high-resolution button text
- #1805: external files in VISE installers
- #1806: writable volume for expanded installer output
- #1807: QuickDraw color restoration
- #1809: cached dialog content bounds
- #1810: MacRoman window titles
- #1812: TextEdit caret clipping
- #1822: host volume-key filtering and function-key translation
- #1813: native Windows cursors, including measured before/after cursor update delay
- #1814: portable headless screenshot paths
- #1818: errors for unsupported device controls

Each PR has isolated native Windows validation and is ready for review. #1812 and #1822 are independently based on upstream master 559d8ab and preserve the remaining ordinary F2/End work in #1753. The uploaded screenshots have been verified against the intended captures and retained with their captions. #1818 removes the garbage icon pixels but leaves the actual icon implementation tracked in #1719.

The remaining #1800 license-job failure is the tooling-download problem described above; it needs a maintainer retry. Newspaper-spin performance is being investigated separately.

benletchford:
The dependency-license failure is an action download failure (`gzip: invalid magic`), not a cargo-deny license finding. Because this is a substantial CPU/presentation-boundary change and related presentation work has landed since the current run, please rebase onto current master and rerun the current CI matrix. The current workflow no longer runs rustfmt.

PR #1821, opened 2026-09-14
fix: honor PowerPC directory queries and application memory requirements

### LINKED ISSUE #1816: Fix EV Nova PowerPC startup directory lookup and memory allocation
EV Nova 1.0.8 reaches “Error initializaing Registration Tool!” in the browser with Systemless 0.40.1 after its original unregistered StuffIt archive is delivered successfully. A native probe using the published 0.40.1 crate reproduces the same dialog. The loader correctly selects `EV Nova 1.0.8/EV Nova` as PowerPC; executable selection is not the cause.

Two runtime problems affect startup:

- `PBGetCatInfo` with a negative directory index must return the directory name through `ioNamePtr`, without reading its prior contents. Leaving the old volume name in that buffer makes Nova look for the wrong Preferences directory and breaks registration-support initialization.
- After correcting the directory lookup, browser startup reaches preferences but reports “Nova needs more memory to run.” Its `SIZE` resource requests a 92,160,000-byte preferred partition, while the native allocator still has a fixed roughly 32 MiB heap. Native launch should honor the preferred partition and explicit memory override without allocating over fixed stack, display, or system mappings.

The expected result is that the unchanged, unregistered shareware game reaches its trial menu and gameplay. Preserve its registration restrictions, and verify actual PowerPC gameplay in the browser before enabling the catalogue entry.

Website tracking: https://github.com/benletchford/systemless.org/issues/95.

benletchford:
EV Nova 1.0.8's unchanged PowerPC shareware build fails during registration-support initialization, then reports insufficient memory once directory lookup is corrected. Return the directory name for negative-index `PBGetCatInfo` queries without reading the output buffer's old contents, and honor the launched application's preferred `SIZE` partition or explicit memory override.

The native heap grows through the shared sparse address space while excluding fixed stack, display, and system reservations. This preserves the existing fixed addresses and the game's shareware restrictions; no game-specific executable patch or registration bypass is involved.

Validation: six focused directory-query tests, a heap-growth/reservation test, a launch/shared-memory aliasing test, the wasm32 compile check, and a release browser build passed. The browser loaded the original archive from R2, created a pilot and ship, completed the introduction, and reached actual PowerPC flight. Initial loading takes several minutes. Separate browser input handling is tracked in benletchford/systemless.org#96.

Closes #1816. Website readiness remains tracked in benletchford/systemless.org#95; the site should adopt the published runtime release before enabling Nova.


benletchford:
This now conflicts with current master after the PPC dispatcher/test modularization and process-context changes. Please rebase onto current master (rather than merging master into the branch), resolve the three touched files, and rerun the current CI matrix; the previous green run does not validate the conflict resolution.

PR #1828, opened 2026-09-14
perf: avoid copying file forks during native metadata publication

### LINKED ISSUE #1827: Avoid copying native file forks during metadata publication
PowerPC runs with large archives spend most of their time copying file forks after each CPU slice. A one-second native profile of EV Nova 1.0.8 attributed about 83% of samples to `publish_native_vfs_catalogue`, predominantly `ProcessForkBytes::clone` and memory copying.

`publish_native_vfs_catalogue` clones complete native file/resource-file records into temporary vectors, but its publication loops consume only the path, type, creator, Finder flags and dirty bit. Fork clones are deliberately detached deep copies, so this repeatedly copies the archive's contents without using them.

Collect only the metadata needed by the compatibility indexes. Preserve detached fork-clone semantics and existing shared-fork ownership. Validate metadata/deletion behavior and compare runtime profiling after the change.

benletchford:
Publishing the native file catalogue after every PowerPC CPU slice cloned complete file records, including detached copies of every data and resource fork, even though publication only reads metadata. A native EV Nova profile attributed about 83% of samples to this publication path, predominantly fork cloning and memory copying.

Collect only each file's path, type, creator, Finder flags and dirty bit, then publish those values to the classic metadata indexes. Ordinary fork clones remain detached, shared process fork ownership is unchanged, and deletion handling follows the same path.

Validation: four catalogue-sharing/mutation tests and the native resource-file metadata mutation test pass. Browser and native profiling follow-up is being checked with the Nova startup fixes separately proposed in #1821.

Closes #1827.


benletchford:
This now conflicts with current master after the process-state soundness changes, and the PR currently has no CI result beyond CLA. Please rebase onto current master, resolve `src/process_context.rs` against the scoped mutation API, and rerun the current CI matrix. I will reassess once the rebased result is green.

Issue #1724, opened 2026-09-13
Add Windows host cursors and profile native GUI presentation performance


rlanday:
Track Windows validation of cursor responsiveness and graphics/CPU performance. The macOS host cursor path is already implemented, but at `61d01c61ef774df9286a3849bb6a1bc6094858fc` Windows still composites the cursor into emulated frames. Its visible movement therefore shares the application's rendering/presentation schedule.

Test environment: native Windows 11 x64 binary cross-built in WSL, Ryzen 7 5700U with Radeon integrated graphics, 1920×1080 at 60 Hz. Test inputs are SimCity 2000 1.2 and EV Override 1.0.1. SimCity currently reaches the installer only because of a separate split-archive blocker; installer timings must not be reported as city-gameplay performance.

Work to validate:

- Provide a Windows host cursor path with guest bitmap/mask, hotspot, visibility, shape changes, DPI scaling, and a software fallback for destination-dependent inversion. Exercise transitions and retain an explicit software-cursor comparison option.
- Run the actual native Windows GUI through both games. Capture CPU and graphics/presentation profiles under controlled workloads; account for scaling, compositing, presentation, and input processing. Avoid overlapping builds/profiling with benchmark samples.
- Compare the retained-pixel hashing change in draft PR #1598 with the same guest time, inputs, saves, and useful gameplay work. The PR's prior macOS EV Override regression remains unresolved by an installer smoke test. Link measured results back to that PR.
- Quantify frame delivery and input-to-presentation timing where possible, clearly distinguishing those measurements from physical input-to-photon latency. No absolute Windows latency or speedup has been measured yet.

Related #1532 tracks intermittent macOS outline-presentation stalls; #1050 contains macOS fullscreen latency evidence. Neither establishes Windows behavior. This is a Windows implementation/profiling task, not a claim that the same macOS bottleneck has been reproduced on Windows.

rlanday:
Windows cursor measurements and implementation are now in #1813: same executable and registration scene, 400 measured moves per mode. Capture-observed update delay was 51.54 ms median / 76.66 ms p95 for software cursors and 0.61 ms / 1.07 ms for native cursors. The PR documents the separate surface/pointer metadata endpoints, variability across trials, and the exclusion of physical input and display scanout. It is ready for review.

[...]

Issue #442, opened 2026-08-08
fix: keep simulation update cadence aligned with guest ticks


benletchford:
## Observed

SimCity 2000's simulation clock advances substantially faster under Systemless than under BasiliskII when both start from the same visible state and receive the same guest-tick input schedule.

After dismissing the welcome notice and Annual Budget at **April 1979**, both runtimes show the same live Demo City. After selecting the bulldozer and advancing the same scripted intervals:

- BasiliskII reaches **August 1979** after the build interaction and **September 1979** after the next 60 guest ticks.
- Systemless reaches **November 1979** and then **January 1980**.
- The premature year boundary opens another Annual Budget dialog in Systemless, so the subsequent Speed-menu click is intercepted.

This is not a font difference: the semantic game date and modal state diverge.

### Paired evidence

Same aligned starting state (April 1979):

| BasiliskII | Systemless |
|---|---|
| ![BasiliskII April 1979](https://raw.githubusercontent.com/benletchford/systemless/issue-evidence/issues/2026-08-09-simcity-simulation-pacing/basiliskii-april.png) | ![Systemless April 1979](https://raw.githubusercontent.com/benletchford/systemless/issue-evidence/issues/2026-08-09-simcity-simulation-pacing/systemless-april.png) |

After the same tool-selection/build schedule:

| BasiliskII (August 1979) | Systemless (November 1979) |
|---|---|
| ![BasiliskII after build](https://raw.githubusercontent.com/benletchford/systemless/issue-evidence/issues/2026-08-09-simcity-simulation-pacing/basiliskii-after-build.png) | ![Systemless after build](https://raw.githubusercontent.com/benletchford/systemless/issue-evidence/issues/2026-08-09-simcity-simulation-pacing/systemless-after-build.png) |

At the next menu interaction:

| BasiliskII (Speed menu, September 1979) | Systemless (Annual Budget, January 1980) |
|---|---|
| ![BasiliskII Speed menu](https://raw.githubusercontent.com/benletchford/systemless/issue-evidence/issues/2026-08-09-simcity-simulation-pacing/basiliskii-speed.png) | ![Systemless premature budget](https://raw.githubusercontent.com/benletchford/systemless/issue-evidence/issues/2026-08-09-simcity-simulation-pacing/systemless-speed.png) |

## Reproduction

- Systemless revision: `55b8c40dfdf9c4a78bdd3ed074385a408ba5df65`
[...]

PR #1943, opened 2026-09-15
perf: bound offscreen text metadata lookups

rlanday:
SC2K's Fire animation spends time searching retained offscreen text metadata while copying ordinary sprite bytes. Keep conservative address bounds around that metadata so reads and writes outside those bounds can skip the tree lookup. This applies to master independently of the DirectWrite prototype.

The game stores its original indexed pixels in guest RAM. Systemless separately retains outline coverage so copying a text bitmap does not flatten its high-resolution text. CPU memory copies must check that coverage, even when most copied data consists of game graphics. Sampling identified the repeated offscreen-tree searches in the Fire sprite loop; its decoding operation is approximately:

```c
for (;;) {
    count = *src++;
    op = *src++;
    switch (op) {
    case 0: break;
    case 1: copy_bytes(dst, src, count);
            dst += count; src += count + (count & 1); break;
    case 2: dst += count; break;
    case 3: row += row_bytes; dst = row; break;
    case 4: return;
    default: debugger_trap();
    }
}
```

This loop performs real sprite decoding. Skipping it as an idle wait would lose pixels. Enabling raw fast memory globally would bypass the interception required to preserve and invalidate copied text. The change instead reduces the cost of the existing memory observation path.

Bounds expand on every insertion path and are rebuilt when a framebuffer replacement retains the offscreen map. Deleting an entry may leave bounds larger than necessary; the map remains authoritative within them, so that only costs extra lookups. The optimization does not alter the guest instruction stream, memory access rules, palette mapping, or text coverage.

Native Windows 11 measurements on a Ryzen 7 5700U, AC power, battery saver off, Balanced plan. Production `--release` builds (opt-level 3, fat LTO, one codegen unit), same SC2K 1.2 input archive and scripted Fire scene. Medians over 250 simulated frontend ticks after the Fire notification is dismissed:

| Build | CPU + audio work per tick | Composition |
|---|---:|---:|
| Master `172027e` | 28.84 ms | 8.74 ms |
| Bounds, candidate run 1 | 26.10 ms | 8.76 ms |
| Bounds, candidate run 2 | 26.46 ms | 8.66 ms |

This is an 8.2–9.5% reduction in CPU work. The earlier `ci-test` comparison improved from 27.09 to 22.80 ms (15.8%); it is a different build profile and is not the production claim. An earlier, cooler release baseline measured 24.86 ms; the table uses the fresh interleaved candidate/master/candidate series to limit machine-state drift. These measurements exclude Windows presentation and do not claim equivalent input latency or displayed FPS. No compiler, guest instruction tracer, or native sampler ran during the measurements.

All 1,550 instruction-count and guest-tick samples, both 4 MiB guest RAM snapshots, and all 11 captured PNGs match master exactly in both production candidate runs. A separate native GUI run exercised registration, year/difficulty buttons, newspaper dismissal, Disasters → Fire, and the National Guard notification.

The native Windows library suite passed 5,468 tests with three ignored, including the new bounds regression and existing CPU text-copy/recoloring tests. The full suite also passed after rebasing onto `4639fd6`, at PR head `72d4b76`; the three intervening upstream commits only refactor PowerPC dispatch.

The reported multi-second menu-input delay and #442's guest-tick cadence divergence remain separate investigations. This change preserves guest timing and does not claim to resolve either.

Risk: missing an insertion path could incorrectly bypass metadata handling. The regression test covers glyph creation, copied detail, both range boundaries, sparse holes, erasure, maximum addresses, and framebuffer recreation with retained offscreen text. Conservative bounds can lose effectiveness as the occupied address span widens.

Related to #1724.


benletchford:
Thanks for your continued perf work @rlanday :)

PR #1948, opened 2026-09-15
perf: reuse unchanged menu bar pixels and text coverage

### LINKED ISSUE #442: fix: keep simulation update cadence aligned with guest ticks

rlanday:
Systemless redraws the complete menu bar during composition even when the menu titles and their appearance have not changed. Reuse a saved 8-bit menu bar, including its retained outline text coverage, to reduce composition work during scenes such as SC2K Fire.

Classic applications still see their original indexed framebuffer bytes and text advances. Systemless keeps separate higher-resolution coverage so menu text can remain sharp when presented larger than the guest's logical pixels. The cache stores both representations using SavedPixels. Every hit replays the snapshot and repairs guest overwrites; unchanged pixels and coverage can be skipped by the existing restore path.

The menu still follows the application’s event handling: it receives the mouse-down event, adjusts enabled items and marks, then calls MenuSelect, as described in [Inside Macintosh: Handling User Choice of a Menu Command](https://dev.os9.ca/techpubs/mac/Toolbox/Toolbox-103.html). The cache is only a replacement for repeated drawing after those inputs have been established.

The key includes guest screen geometry and depth, menu height, theme, current palette contents, live MenuCInfo bytes, title text/IDs/positions/enabled state, highlighting, and retained outline scale. Changes take the normal repaint path and produce a fresh snapshot. Only the existing 8-bit snapshot path is cached.

A cache hit also completes pending CPU recoloring at the same native drawing boundary as a normal repaint. That matters for CPU-drawn highlights: skipping every menu store must not leave the body's text coverage uncommitted. This is covered by a dedicated regression test.

Native Windows 11 measurements on a Ryzen 7 5700U, AC power, battery saver off, Balanced plan. Both builds use production `--release` settings (opt-level 3, fat LTO, one codegen unit) and the same scripted SC2K 1.2 Fire scene. Medians over 250 simulated frontend ticks after the notification is dismissed, from an interleaved candidate/baseline/candidate run:

| Build | Composition per tick | CPU + audio per tick |
|---|---:|---:|
| Baseline with merged Fire CPU fix (#1943) | 7.22 ms | 23.82 ms |
| Menu cache, run 1 | 3.21 ms | 24.20 ms |
| Menu cache, run 2 | 3.38 ms | 24.40 ms |

Composition falls by **53–56%**. CPU execution is essentially unchanged; the measurements do not claim a CPU speedup. No compiler, instruction tracer, or native sampler ran during these measurements. These are host work durations per simulated tick, excluding desktop acquisition/submission and physical scanout; they are not displayed FPS or click latency.

Both candidate runs exactly match all 1,550 guest instruction/tick samples, both 4 MiB RAM snapshots, and all 11 captured PNGs. The candidate is based on upstream `c9694d4`, including #1943. The frozen comparison executable is `172027e` plus the same #1943 bounds change; the intervening upstream changes are PowerPC dispatcher refactors. A native GUI run also exercised registration, the New City dialog, newspaper dismissal, Disasters → Fire, and the National Guard notification.

A separate native GUI diagnostic, using the same production library and buffering its records in memory, measured three Fire menu openings at 70–314 ms. These are individual PostMessage-to-capture observations with different amounts of guest work before the next event poll, not a controlled before/after latency benchmark. Per-frame stderr tracing was disabled: writing that trace to a WSL file costs about 9.25 ms per frame on this setup and creates multi-second delays. The performance table above was already measured without that tracing. This PR reduces composition cost and preserves guest event/tick semantics; it does not claim to resolve #442 or establish the cause of the original user-reported delay.

The final native Windows library suite passed **5,471 tests, zero failures, three ignored** on upstream `c9694d4` plus this change, including both new menu-cache regression tests. The previous-base run also passed (5,469 tests).

The primary risk is a missing cache input causing stale menu appearance. The regression checks compare cached output with a forced repaint after title changes, enabled-state changes, highlighting, palette edits without a seed bump, menu height, theme, outline scale, SetMCEntries, and direct guest writes to MenuCInfo. They also overwrite the bar and require its pixels and retained coverage to be restored exactly. A separate test covers the pending recoloring boundary. The cumulative glyph-draw counter can decrease because a cache hit no longer redraws glyphs; it is not guest-visible state.

Related to #1724 and the SC2K Fire menu-latency investigation. This does not change guest-tick accounting and does not claim to fix #442.


benletchford:
@cla-assistant recheck

PR #1949, opened 2026-09-15
perf: precompute image coverage and normalization

rlanday:
Fractional window sizes spend substantial time resampling the presentation image. SC2K at 1104×828 presents a 1600×1200 retained-text surface through the software resize path; a native Windows Fire trace with menu-bar caching measured about 45 ms rendering per frame, compared with 6.3 ms at 800×600. This change reduces the resampling work while preserving every output pixel.

The software filter averages exact source-cell coverage on shrinking axes and uses nearest samples on enlarging axes. Previously it repeatedly calculated interval intersections for each output pixel, then divided each of four channel sums by the same total coverage. Precompute the source cells and integer weights once per axis, plus an integer reciprocal for the invariant coverage total. Keep the existing accumulation and round only once, after both axes.

For a rounded channel numerator `n` and positive total `d`, normalization is:

```c
r = UINT64_MAX / d;                  // once for the resize
q = high_64_bits((uint128_t)n * r);
q += (n - q * d >= d);
```

The multiply estimate never exceeds `n / d` and is at most one below it: `r` underestimates `2^64 / d` by at most one, and `n < 2^64`. The remainder comparison corrects that final unit. This is exact integer division, including its rounding behavior; it does not introduce floating-point filter approximations. Guest pixels, font advances, retained text coverage, and guest instruction/tick accounting are unchanged.

Isolated native Windows 11 resampling measurements on a Ryzen 7 5700U, AC power, battery saver off, Balanced plan; Rust opt-level 3, fat LTO, one codegen unit. Each pair alternates its execution order, with medians over 20 measured repetitions after two warmups:

| Source → destination | Before | After |
|---|---:|---:|
| 1600×1200 → 1104×828 | 34.24 ms | 14.79 ms |
| 1600×1200 → 1200×900 | 34.76 ms | 13.43 ms |
| 800×600 → 600×450 | 8.77 ms | 3.62 ms |
| 800×600 → 1600×1200 | 53.81 ms | 13.75 ms |
| 800×600 → 800×600 | 0.21 ms | 0.21 ms |

The affected fractional-size case improves by **57%**. These numbers isolate resizing; they are not a promise of the same improvement in total frame time or menu latency. No compiler or profiler ran during the measurements. The earlier multi-second GUI probes with per-frame stderr tracing are excluded; that file I/O itself costs about 9 ms per frame when directed to WSL.

Validation: native Windows execution matched the old filter exactly on 1,500 generated source/destination cases and all larger benchmark images. The new mixed-axis test checks all four channels and rounding; reciprocal checks cover small divisors, maximum integers, and 100,000 generated numerator/divisor pairs. The native Windows production release build passed. The branch is now based on refreshed master `2382758`, including merged #1948, and remains independent of the text WIP. The tested GUI stack has identical 68k, display, and menu-cache sources; the rebase adds only upstream PowerPC dispatch refactors, covered by the new CI run. The earlier CI packaging failure was an external MPW download connection reset (`curl: (56)`), before the Rust/package checks.

A native Windows GUI run exercised registration, New City, newspaper dismissal, Disasters → Fire, the National Guard notification, and resizing between 800×600 and 1104×828. During the first resized Fire menu probe, median rendering fell from 44.67 ms in the earlier cached-menu baseline to 23.06 ms with this change. At 800×600, rendering remained about 6.3–6.4 ms. Records were buffered in memory until close; per-frame stderr tracing was disabled.

All three menu probes opened successfully. The two resized-window captures observed the menu by 536 ms and 265 ms; the normal-size capture observed it by 272 ms. Earlier resized baseline observations were 1.92 and 2.06 seconds. **These are not a controlled before/after latency ratio:** the clicks encountered different amounts of guest work before the next event poll (baseline about 5.2 million instructions; candidate 1.88 million and 0.22 million). PostMessage-to-PrintWindow measurements include capture overhead and exclude physical scanout. The rendering-time reduction is the stronger result; this change does not promise instantaneous menus or fix guest simulation cadence.

The main risks are an incorrect weight table or quotient correction changing a pixel. Exact old/new comparisons and the arithmetic boundary tests cover those paths. Axis tables add short-lived storage proportional to the axis dimensions; the identity-size copy still returns before allocating them. The integer operations are portable Rust, but these performance measurements are for Windows, not macOS or browser engines.

Related to #1724. This does not alter the simulation cadence discussed in #442.


benletchford:
Reviewed the draft. The implementation and focused arithmetic/resize coverage look coherent, and the main build/test job is green. The current `headless & package` failure is an external MPW image download reset (`curl: (56)`), not a code failure. Please rebase onto current `master` now that #1948 has landed, rerun the failed gate, and complete the native release plus resized-window GUI verification already called out in the description before marking this ready for review.

rlanday:
Addressed all three requests:

- Rebased to current master `2382758`, including merged #1948; PR head is `9299465`.
- The fresh CI run is entirely green, including `headless & package`: https://github.com/benletchford/systemless/actions/runs/34931289660. The previous failure was the MPW download connection reset you identified.
- Completed the native Windows release build and the combined-cache GUI run through registration, New City, newspaper dismissal, Fire, the National Guard notification, and resizing between 800×600 and 1104×828. Median rendering in the first resized Fire menu probe fell from 44.67 to 23.06 ms; normal-size rendering stayed about 6.3–6.4 ms. All three menu probes opened successfully. The tested 68k/display/cache sources are identical to the rebased branch; the intervening upstream changes are PowerPC refactors, covered by the green CI run.

I updated the description with the native results and the measurement limits: menu captures were visible by 265–536 ms, but the clicks encountered different amounts of guest work, so I am not claiming a controlled latency percentage. The isolated filter comparisons remain pixel-exact. Marking ready for review.

benletchford (approved):
The requested rebase, complete green CI rerun, native release build, and resized-window GUI verification are all in place. The precomputed coverage tables preserve the existing filter semantics, and the reciprocal quotient correction is exercised across boundaries and generated inputs. Approved.

PR #2013, opened 2026-09-18
perf(cpu): enable native memory traces with observed writes

rlanday:
SC2K's sprite and pixel-copy loops currently lose native memory traces while high-resolution text presentation is active. Raw JIT stores would bypass the metadata that keeps copied text sharp and erases stale text on later guest writes. Opt into the new tracked-memory capability so native traces read stable RAM directly and route stores through the existing bus hooks.

**Draft dependency:** depends on [m68k-rs #181](https://github.com/benletchford/m68k-rs/pull/181). The Cargo patch pins the public fork commit `02cb0a169e9a9bd5008617a9434db7c21f19ebf8` so reviewers can build this checkout. Before merging, replace it with a released m68k version and remove the patch. Cargo packaging cannot publish the unreleased API through this temporary patch.

Related to #1724. This applies to master independently of GPU PR #2008 and the font work.

Rebased onto master `35f7a15` (0.41.10), including upstream menu-hover fix #2010. The full 5,494-test native-JIT library suite passes again at `82309bc`. The timing comparisons below were captured before #2010 on the 0.41.9 base; they isolate this CPU capability. A fresh combined-stack Windows release build including #2010 also passes; the additional production GUI check below records remaining latency and whole-process CPU cost. The earlier GUI latency observations below predate #2010. The menu-hover fix handles work after tracking begins; the queued redraw delay described below precedes tracking.

### How it works

The CPU API adds a read-direct RAM window with notifying writes. Native code handles the guest loads, arithmetic, address updates, and control flow; each compiled store calls the existing native bus handler. Memory-to-memory MOVE also invokes the begin/end-copy hooks so source text coverage is captured before overlapping writes. The m68k PR contains the code generation and interpreter-equivalence tests; this PR only exposes an eligible window and tests the real Systemless metadata paths.

Systemless offers the capability for stable owned/external RAM with 32-bit addressing. Shared/foreign mappings, 24-bit addressing, watchpoints, and per-access diagnostics retain their ordinary paths. The CPU additionally declines MMU translation and pre-68020 models. Presentation, write-probe journaling, and protected stores remain active. Existing raw `FastMem` stays preferred when it is already eligible.

Each compiled trace records whether stores are direct or notifying. The CPU rejects and re-records traces if that capability changes between batches. There is no per-store mode branch on the original raw path. `SYSTEMLESS_DISABLE_TRACKED_JIT=1` provides a diagnostic before/after comparison; omit it for normal use.

This enables a general JIT capability. SC2K's RLE sprite loop, approximately, is:

```c
for (;;) {
    count = *src++;
    op = *src++;
    switch (op) {
    case 0: break;
    case 1: copy_bytes(dst, src, count);
            dst += count; src += count + (count & 1); break;
    case 2: dst += count; break;
    case 3: row += row_bytes; dst = row; break;
    case 4: return;
    default: debugger_trap();
    }
}
```

We could not safely turn on raw writes for this loop: equal-value writes can erase retained outline detail, and copies must carry that detail. A framebuffer diff after execution cannot recover these operations. The new API preserves the operations while removing guest instruction dispatch. No game-specific code substitution or change to the guest clock/instruction budget is involved.

### Windows performance and correctness

Optimized native Windows build (`x86_64-pc-windows-gnu`, Rust 1.98.0, fat LTO, one codegen unit). Same executable with the capability toggled, isolated copies of the same SC2K 1.2 archive, fixed guest start time/instruction rate and scripted input/audio. Each run has 2,500 virtual frontend frames; the table covers the same active 1,200 frames, excluding initialization and Fire selection. Fire was run after→before; ordinary city before→after. These are headless CPU measurements, not displayed FPS or input latency.

| CPU measure | Before | After |
| --- | ---: | ---: |
| Ordinary city total | 3.968 s | 0.986 s (4.02× faster) |
| Ordinary city p95/frame | 21.985 ms | 3.552 ms |
| Fire total | 27.069 s | 5.123 s (5.28× faster) |
| Fire median/frame | 21.852 ms | 3.581 ms |
| Fire p95/frame | 27.229 ms | 5.823 ms |

In both pairs, all 2,500 records agree in retired/foreground/cumulative instructions, guest tick, PC, GetNextEvent count, and UI-tracking state. All four sampled presented PNGs are byte-identical. This checks the sampled state/output, not every byte of guest RAM throughout the run. Earlier paired runs gave similar results.

Separate native GUI trials on the combined CPU/GPU stack used the same executable with the capability toggled: 20 ordinary and 20 Fire menu clicks per setting. All 80 clicks were found and matched the guest event trace, with no pointer interference. Ordinary median/max response was 43.7/912.0→40.5/320.9 ms; Fire was 396.5/798.9→233.9/318.2 ms. The generated cities had different terrain, so these are observational comparisons, not a controlled same-city latency claim. They measure click-to-observed-desktop-menu, not physical input-to-photon latency.

A further production GUI check on 0.41.10 + this PR + GPU PR #2008 verified Fire visually before testing. All 12 menu clicks succeeded without pointer interference: median **146.4 ms**, maximum **202.0 ms**. This is a fresh-city observation, not a controlled comparison with the older terrain above. Separate 15-second intervals, with no profiler or menu probe running, measured total process CPU across all threads at **82.0% of one core at 800×600** and **81.2% at 1104×828**; Fire remained visible before and after both intervals. These totals include the host rendering/audio/presentation work excluded from the headless CPU table. Significant latency and CPU cost remain and are being investigated separately. No experimental input-speed boost was used in this production check.

The trace explains the remaining delay: SC2K executes millions of sprite instructions before its next GetNextEvent. The baseline repeatedly missed its allotted guest instruction work within the host frame deadline; the new path generally completes it. Remaining waits around 300 ms still contain redraw work at the configured guest CPU rate. This PR improves menu responsiveness but does not claim to eliminate all menu delays or address #442's separate guest-clock policy.

### Validation and risks

New integration tests cover hot glyph copies out of and back into the framebuffer, equal-value stores erasing outline detail, mutation of a compiled instruction, write-probe journals, and protected stores. The CPU PR compares registers, flags, RAM, instruction counts, and exact copy/write notifications against single stepping; all ten focused tests also pass on native Windows. Existing raw-memory microbenchmarks show no material regression.

The full library suite with native JIT passes against the pinned Git dependency: **5,494 passed, 0 failed, 3 existing ignores** (`cargo test --lib --no-default-features --features jit --locked`). Linux all-feature GUI compilation and the no-default-features portable build check also pass. Native Windows release and the combined GPU/CPU GUI were validated with the same CPU/runtime implementation.

`cargo package --locked --allow-dirty` was checked and currently fails as expected: Cargo removes the review-only patch when verifying the packaged crate, so released m68k 0.13.0 lacks `TrackedMem`/`tracked_mem`. The **headless & package CI job is consequently expected to fail on this draft** until m68k-rs #181 is released and this dependency is updated. Other checks should pass; this is not ready to merge.

The main risk is offering direct reads for a memory mapping with hidden read side effects or unstable backing storage. The window conservatively excludes those existing cases, and future mapping features must preserve that contract. Writes retain bus bookkeeping rather than bypassing it. Frequent capability changes can cause trace recompilation. The temporary dependency patch is a merge/release blocker, not intended production packaging.


benletchford:
I merged m68k-rs #181 after validating it end to end. This draft is still correctly blocked on the published crate: its temporary Git patch makes package verification fail, and public manifests here must use crates.io versions. Once release-please publishes the m68k release, this branch can be updated to the released dependency and lockfile, the patch can be removed, and package plus full tests can be rerun against current master. I’m leaving it open rather than merging the review pin.

benletchford:
The release blocker is resolved. I rebased this branch onto current master, replaced the review-only Git patch with published m68k 0.14.0, regenerated Cargo.lock, and marked the PR ready. Local validation against the published crate passed: cargo package --locked --allow-dirty and the native-JIT library suite (5,509 passed, 3 ignored). Fresh CI is now running on the updated branch.

benletchford (approved):
Reviewed the tracked-memory integration and the rebased dependency update. The implementation preserves protected writes, write probes, presentation bookkeeping, and capability invalidation; local package verification and the full native-JIT library suite pass against published m68k 0.14.0.

PR #2225, opened 2026-09-20
perf(ppc): bulk guest memory and detail walks instead of per-byte

wraitii:
## Summary

3 commits for a rather significant performance improvement on PPC. Tested on escape velocity Nova, about 3.5x faster to get in game. Still far from 'playable' but it's not horribly laggy any more.

## Validation

Instructions, audio_hash and captured_mono_samples bit-identical across 20 game/mode combinations; 68K wall times unchanged; full suite green including toolbox_showcase; no new clippy warnings.


benletchford:
Wow thanks!

PR #2282, opened 2026-09-20
perf: skip unchanged browser frame transfers

### LINKED ISSUE #2277: perf: avoid transferring unchanged browser frames
The browser runtime currently serializes and transfers a full RGBA framebuffer after every PPC visual CPU slice, even when the retained presentation cache reports that pixels and host overlays are unchanged. EV Nova flight profiling shows worker run time around 20 ms while only about 13 changed paints/s are produced.

Acceptance criteria:
- unchanged cached frames produce no RGBA transfer or canvas upload;
- first paint, palette/cursor/debug changes still present correctly;
- deterministic browser fixture output remains unchanged;
- measure worker frame time and changed-paint cadence on an EV Nova flight.

benletchford:
EV Nova's browser runtime ran the retained-frame renderer and then copied/transferred a full RGBA framebuffer after every PPC slice, even when the presentation cache reported unchanged pixels. That wasted worker time and caused needless canvas uploads.

This change returns the cache-hit status with the rendered slice. Worker frames are only transferred when the presentation changed, and the local canvas path skips identical uploads while preserving first paint, cursor, palette, debug, scale, and QD3D behavior.

Closes #2277.

Validation:
- `cargo test -p systemless-org --no-default-features` (71 tests)
- `cargo check -p systemless-org --target wasm32-unknown-unknown --no-default-features`
- changed-file `rustfmt --check`
- deterministic output path remains unchanged; browser flight measurement is being run against the candidate artifact.


benletchford:
Closing this candidate after controlled EV Nova flight measurements. The unchanged-frame path did reduce changed transfers (about 9.3/s versus the 13.3/s baseline), but it also regressed guest throughput to 37.1 ticks/s versus the available same-scheduling 52.5 ticks/s control, with worker p50/p95 rising to 23.7/36.1 ms from 20.8/29.4 ms. The repeat baseline could not complete, so this is not a mergeable performance result; the source needs rework with a matched artifact/control pair.

benletchford:
Closing after the controlled flight regression described above; retain the issue for a future matched rework.

PR #2363, opened 2026-09-21
perf(ppc): cache decoded blocks for writable guest code

wraitii:
## Summary

- `PpcSectionMem` issues an instruction-cache token only for a read-only region, treating a writable one as possibly self-modifying.
- Classic Mac OS has no memory protection and CFM publishes fragments into the writable application heap, so that gate refused a token for **every instruction a PowerPC app executes**.
- `run_cached_basic_block` therefore bailed on its first line, and each instruction paid a full single-step fetch — a region lookup plus two lazy-allocation checks, per block start and per word while building one.
- Two commits: issue our own token for writable code, stamped with page and generation; then answer repeat token requests from a one-entry page cache.
- Invalidation follows the contract PowerPC hardware already requires of the guest: a write to a page we have executed from rotates that page's generation, and `MakeDataExecutable` / `FlushCodeCache` rotate the whole process. The write side is gated by the existing shared-mapping page filter, so an ordinary store pays one word test.

### Performance

Measured on EV Nova at 1024×768, same workload both runs (loading → menu → in-game), `samply` main thread. Both builds already carry the page-index work from #2252, so this isolates the block cache.

- Instruction-fetch cluster: **22.47% → 2.32%** of main thread (6.97s → 0.54s)
  - `locate_cached` 11.58% → 1.48%
  - `read_instruction_u32_be` 4.33% → 0.06%
  - `read_u32_be` 4.50% → 0.62%
  - `instruction_cache_token` 2.06% → 0.00%
- The new `resolve_sparse_instruction_token` costs **0.15%** (0.03s) — the token cache pays for itself many times over.
- Interpreter subtree `run_with_imports_and_cycle_handler`: **80.4% → 60.7%** inclusive (24.94s → 14.10s).
- Net effect: the emulator is no longer instruction-fetch-bound. `render_frame` is now the largest remaining subtree at 23.6%, up from 11.6%, because the guest drives more frames once execution stops being the bottleneck.

## Validation

Done:
- Full library suite passes on this branch rebased onto current master: 5598 passing, 0 failing. `cargo build --all-targets` clean, no new clippy errors.
- Five new tests covering the invalidation contract:
  - `writable_heap_code_is_decoded_once_across_repeated_runs`
  - `writes_clear_of_executed_code_leave_the_decode_cache_alone`
  - `rewriting_one_code_page_leaves_the_other_pages_cached`
  - `guest_instruction_cache_flush_retires_writable_code_tokens`
  - `token_reuse_stops_at_a_region_that_ends_inside_its_page`
- The page cache is populated only when one writable region covers the whole page, so the fast path cannot hand out a token past a region's end; every rotation clears it alongside the mappings that could shadow the page.
- A/B confirmed by build fingerprint, not by file naming — the "before" profile has no `resolve_sparse_instruction_token` symbol.

Potentially to do:
- **No end-to-end speedup figure.** The two captures are different wall lengths (34.6s vs 24.9s) and carry no frame markers, so the share numbers above are solid but a wall-clock ratio would not be. A fixed-tick run would give a real one.
- **Self-modifying code is covered by contract, not by a hostile test.** Invalidation rests on the guest honouring `MakeDataExecutable` / `FlushCodeCache`. A guest that rewrites code without either call would execute stale blocks. Worth deciding whether that deserves a deliberate test or a documented limitation.
- Interaction with `ade2f646` and `3a7f239d` (landed after these commits were written) has not been examined beyond confirming the code they touch is disjoint.
- Only exercised on EV Nova; no coverage of other PPC titles or of a CFM fragment mapped read-only.

(PR written by claude opus 5 and tested manually)


benletchford:
This looks like a useful next performance candidate. It targets PPC instruction-fetch overhead and complements the visible-frame caching improvement in #2362. The native profile is encouraging, but the captures do not yet establish an end-to-end browser speedup. I would require matched queued EV Nova browser runs against current master, plus a fixed-input/fixed-progress framebuffer comparison, before merging.

Two correctness concerns from reading the diff:

1. **Writable tokens lack an address-space identity.** `writable_code_token` combines the page number with counters initialized independently in each address space. Two spaces containing different instructions at the same address, with the same mapping/write counts, can therefore return equal tokens. The `PpcMemory::instruction_cache_token` contract explicitly permits CPU cache reuse across different memory values when tokens match. Please add a test that runs one `PpcCpu` against two such spaces and verifies the second program executes, and give the tokens an identity that cannot alias between those spaces. Generation wrapping needs the same consideration.
2. **Cached-span writes bypass invalidation.** `GuestAddressSpace::write_u16_be_in_span` still calls `regions.write_u16_be_in_span` directly, without `note_write` or rotating the code-page generation. Please test decoding a writable block, changing an instruction through this API, and rerunning it on the same CPU. The span path should invalidate correctly, including when a span was obtained before the page was first executed.

These are code-review findings, not locally reproduced failures yet. The implementation already invalidates ordinary observed stores, so I would describe the self-modifying-code limitation more precisely than “all rewrites require an explicit guest flush”: the concern is ensuring every mutation path and token lifetime obeys the cache contract.

For browser validation, #2362 reached 59.91 guest ticks/s in two 20-second runs and 60.02 ticks/s over a 60-second run. Additional CPU savings could still provide valuable headroom for busier scenes and slower machines. This is also materially different from the previously rejected immutable-token reuse experiment; it deserves its own matched A/B.

Issue #2367, opened 2026-09-21
fix: preserve writable PPC instruction cache coherence


benletchford:
The writable PPC block cache introduced in #2363 needs stronger invalidation guarantees before release.

Code review identifies two ways cached instruction bytes can outlive their backing contents:

- Independent address spaces can mint equal page/generation tokens for different code. Reusing one `PpcCpu` across them can select the first space’s cached instructions.
- `GuestAddressSpace::write_u16_be_in_span` bypasses the new write notification, so modifying cached code through a previously acquired writable span does not retire its token.

Add executable regression tests for these cases, detached clones, generation exhaustion/wrapping, and a write to a later instruction within a currently executing block. Ensure every successful mutation retires the affected instruction view, and that tokens cannot alias unrelated memory instances. Preserve caching for unaffected pages and ordinary data stores.

Validate the corrected implementation with full relevant tests and matched queued EV Nova browser A/B against v0.47.3, including a fixed-input/fixed-progress framebuffer comparison. This follows up #2363 and contributes to #2242.

benletchford:
#2368 landed the address-space identity and cached-span fixes. I also reproduced two remaining coherence cases in focused tests:

- a store rewriting the next instruction inside the currently executing cached block;
- a shared mapping shadowing a later instruction in a cached block, including across a page boundary.

I tested a follow-up that used process-wide non-wrapping token allocation, page-scoped adapter tokens, and a cheap mutation epoch checked after store instructions. The deterministic 600-frame Nova run remained identical (`58d8a3b54744981ef9b8ce0a5374660028c354eb02b2e781182ec31041a9edc7`), and both builds sustained about 60 guest ticks/s.

The matched browser A/B nevertheless regressed the hot path:

| Metric | Control mean | Candidate mean |
| --- | ---: | ---: |
| Worker run p50 | 8.00 ms | 8.45 ms |
| Worker run p95 | 21.85 ms | 22.20 ms |
| Paint throughput | 41.23 fps | 39.94 fps |

I rejected the candidate and did not publish it. The remaining cases need a cheaper invalidation mechanism that does not add work to ordinary cached execution.

benletchford:
[...]

PR #2368, opened 2026-09-21
fix(ppc): keep writable-code tokens unique to their address space

wraitii:
## Summary

Follow-up to #2363, fixing the two correctness findings from its review.

**1. Writable-code tokens could alias across address spaces.** The token combined a page number with counters that start at zero in every fresh space, so two spaces holding different code at the same address returned identical tokens — and `PpcMemory`'s contract lets the CPU reuse a decoded block on a token match. `GuestAddressSpace::clone` reached that state directly: it copied the counters while the regions became independent allocations.

Tokens now come from one process-wide counter. A page keeps its token until a write retires it; a retired page mints a fresh one. Every `(space, page, revision)` is distinct by construction, and there is no counter left to wrap. Clones start empty.

**2. Span writes bypassed invalidation.** `write_u16_be_in_span` went straight to the region, skipping the `note_write` that retires an executed page. This was live — `loader/ppc` imports `GuestAddressSpace as PpcSectionMem`, so the Q3 software front-buffer path used it. Spans are opaque, so `writable_span` now returns a `GuestWritableSpan` carrying the guest base address, and the write reports like any routed one.

## Validation

- 5601 passing, 0 failing. `--all-targets` clean, no clippy errors.
- Three new tests, **each confirmed to fail on the pre-fix commit**:
  - `one_cpu_keeps_two_address_spaces_at_the_same_address_apart`
  - `a_detached_clone_does_not_inherit_its_parents_code_tokens`
  - `a_write_through_a_cached_span_retires_the_code_it_overwrites`
- A span taken before its page is ever executed from still invalidates, because the executed-page test happens at write time. The third test takes the span first.
- **Behavioural change to review:** `note_write` also mirrors span writes to presentation, which that path previously skipped. Should be a no-op — `writable_span` refuses when `presentation.observes(addr, len)` — but it is on a per-pixel path.

(PR written by claude opus 5)


wraitii:
This fixes #2367  I think ?

